Issue #139, opened 2014-10-15
MIME multipart support


danroth27:
We should provide parsing support for dealing with MIME multipart messages. This would be useful for things like file uploads and batching.

rosieks:
Does that mean that asp.net 5 rc1 will contain batch support?

Tratcher:
No. Open a new issue here https://github.com/aspnet/home/issues to discuss support for batch.

PR #146, opened 2014-11-05
Mime multipart request parsing.

Tratcher:
#139
This implementation includes several layers of support for reading multipart request bodies.
1. The lowest level is the MultipartReader that can parse any stream minimal buffering. It gives you the headers and body of each section one at a time and then you do what you want with the body of that section (buffer, discard, write to disk, etc.).
2. The next level up supports buffering the individual bodies to memory or disk and returning you an  IEnumerable of all the sections.
3. The next level up exposes extension methods on HttpRequest to determine if this is a multipart request and to read it using the methods from level 2 above. This will also cache the results in a feature interface so they can be accessed again by later calls.
4. There is also extension method support for reading nested multipart entries.

Here is a functional sample:
https://github.com/aspnet/Entropy/commit/4e083df512b90666f65cf406eae6c7128649c7fd

None of this work is prescriptive about how you want to interpret the content. If we want to interpret it as files then we would add one additional layer that would read the content-disposition header to get the file details. However, when we get strongly typed headers that information should be more readily available anyways.


src/Microsoft.AspNet.WebUtilities/StreamHelperExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+    public static class StreamHelperExtensions
+    {
+        public static async Task DrainAsync(this Stream stream, CancellationToken cancellationToken)
+        {
+            byte[] buffer = new byte[1024];
+            cancellationToken.ThrowIfCancellationRequested();
+            while (await stream.ReadAsync(buffer, 0, buffer.Length, cancellationToken) > 0)
+            {
+                // Not all streams support cancellation directly.
+                cancellationToken.ThrowIfCancellationRequested();
```
davidfowl: Ah yes true.

src/Microsoft.AspNet.PipelineCore/FormFeature.cs
```diff
@@ -1,71 +1,200 @@
...
+
+                return HasApplicationFormContentType() || HasMultipartFormContentType();
+            }
+        }
+
+        public IFormCollection Form { get; set; }
+
+        public IFormCollection ReadForm()
+        {
+            if (Form != null)
```
davidfowl: Does this need to be thread safe or anything? What happens if more than one thread accesses Form at the same time? Does it get read twice and fail?

Tratcher: Nothing else in the platform is thread safe, why start here?

davidfowl: I'm mostly asking about being about to corrupt to form reading/body reading because it might be expensive. I'm not sure if there are other things like this currently on the request/response.

Tratcher: There's nothing nearly as expensive elsewhere on the request or response yet. However, there are plenty of things on both request and response that are not thread safe. Reading the request body, writing the response body, setting response headers, etc.. Even reading the request headers is not guaranteed to be thread safe.

ajaybhargavb:
The `ReadFormAsync` method doesn't support multiple files with same name. It always returns the last file.
Eg:
> `<input type="file" name="files" id="file1" />`
`<input type="file" name="files" id="file2" />`

Here `ReadFormAsync` returns only the second file.

Tratcher:
@ajaybhargavb Fixed.

rynowak:
:shipit: from me based on the public surface area

src/Microsoft.AspNet.PipelineCore/FormFeature.cs
```diff
@@ -1,71 +1,195 @@
...
 
-            var body = _request.Fetch(_features).Body;
+            if (!HasFormContentType)
+            {
+                throw new InvalidOperationException("Incorrect Content-Type: " + _request.ContentType);
+            }
 
-            if (_bodyStream == null || _bodyStream != body)
+            // Read the form async and block, but don't block indefinitely, this could be a 2gb file upload.
+            using (var cts = new CancellationTokenSource(TimeSpan.FromSeconds(30)))
```
lodejard: This would be a really difficult timeout to predict, reproduce, etc in production. How about adding an optional cancellation token argument to ReadForm() but have the default behavior assume a timeout for synchronous-request-body-reading is provided in a different way.

src/Microsoft.AspNet.PipelineCore/FormFeature.cs
```diff
@@ -1,71 +1,195 @@
...
+                {
+                    return true;
+                }
+
+                return HasApplicationFormContentType() || HasMultipartFormContentType();
+            }
         }
 
-        public async Task<IReadableStringCollection> GetFormAsync(CancellationToken cancellationToken)
+        public IFormCollection Form { get; set; }
```
lodejard: Form property should be backed by a field. The property get should return ReadForm implicitly.

lodejard: Ah, I see, the outer HttpRequest.Form property get calls ReadForm internally, so that's already how it works at HttpRequest level. That said - wouldn't it be better if IFormFeature and HttpRequest Form properties acted the same?

src/Microsoft.AspNet.PipelineCore/FormFile.cs
```diff
@@ -0,0 +1,46 @@
...
+            _length = length;
+        }
+
+        public string ContentDisposition
+        {
+            get { return Headers["Content-Disposition"]; }
+            set { Headers["Content-Disposition"] = value; }
+        }
+
+        public string ContentType
```
lodejard: I still don't trust this property, :), but no harm exposing it I suppose.

src/Microsoft.AspNet.PipelineCore/FormFeature.cs
Tratcher: `Form` is the storage "field", we just don't put fields on interfaces. We want the raw value to be directly accessible so you can do things like check if the form has been read yet without actually triggering a read.

The only reason I put a Form property on HttpRequest instead of a ReadForm() method is to make it easier to assign the form in a unit test.  ReadForm() would better convey the potential costs of accessing the form.

PR #161, opened 2015-01-09
Added extension methods for FormFile

ajaybhargavb:
@Tratcher Please review.

### DIFF
--- a/src/Microsoft.AspNet.Http.Extensions/FormFileExtensions.cs
+++ b/src/Microsoft.AspNet.Http.Extensions/FormFileExtensions.cs
@@ -1,0 +1,47 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Http
+{
+    /// <summary>
+    /// Extension methods for <see cref="IFormFile"/>.
+    /// </summary>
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Http.Extensions/FormFileExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+                inputStream.CopyTo(fileStream);
+            }
+        }
+
+        /// <summary>
+        /// Asynchronously saves the contents of an uploaded file.
+        /// </summary>
+        /// <param name="formFile">The <see cref="IFormFile"/>.</param>
+        /// <param name="filename">The name of the file to create.</param>
+        public async static Task SaveAsAsync([NotNull] this IFormFile formFile, string filename)
```
Tratcher: This should take an optional CancellationToken.

Tratcher:
:shipit: after adding the CancellationToken

src/Microsoft.AspNet.Http.Extensions/FormFileExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Http
+{
+    /// <summary>
+    /// Extension methods for <see cref="IFormFile"/>.
+    /// </summary>
+    public static class FormFileExtensions
+    {
+        private static int DefaultBufferSize = 81920;
```
Eilon: Usually we write these out as multiples of 1024 so it's easier to see the "native" number. And, if possible, a comment indicating why this number was chosen (what's so special about 80KB?).

Issue #163, opened 2015-01-10
Remove WebUtilities dependency on Http


Tratcher:
Microsoft.AspNet.WebUtilities currently depends on Microsoft.AspNet.Http for the following classes:
QueryBuilder - Has a ToQueryString() method returning QueryString
UriHelper - Uses HostString, PathString, QueryString, FragmentString, HttpRequest, etc..

This library is only supposed to contain stand-alone parsers and utility methods. Consider moving the dependent classes to Microsoft.AspNet.Http.Extensions.

davidfowl:
Feels like we should do this sooner rather than later. Would we keep those types in the Microsoft.AspNet.Http namespace? Feels weird to have those types on the default abstractions in another namespace.

Tratcher:
I fixed this by moving QueryBuilder and UriHelper as part of my typed headers work. PR incoming.

Issue #118, opened 2014-08-25
Remove temp HttpAbstractions for conneg from MVC.


harshgMSFT:
https://github.com/aspnet/Mvc/issues/838

Issue #68, opened 2014-06-03
Provide parsed Accept and ContentType headers


yishaigalatzer:
Something ala - http://msdn.microsoft.com/en-us/library/system.net.http.headers.mediatypewithqualityheadervalue(v=vs.118).aspx

Issue #19, opened 2014-03-14
Add support for parsing specific header values


pranavkm:
This is an extension of #15. MVC uses the media type segment of the content type for determining what value provider \ input formatter to use. It would be convenient for framework developers to have this as part of HttpAbstractions \ some common code that sits on top of HttpAbstractions.

PR #164, opened 2015-01-13
Strongly typed headers

Tratcher:
Includes:
- Adding strongly typed parsers for all common header types. Most of these came from System.Net.Http.Headers, plus cookies as inspired by WebApi. The code infrastructure has been cleaned up and I've added list parsing for the appropriate header types.
- Removing some headers from HttpRequest (Accept, etc.), but leaving ContentLength and ContentType
- Adding HttpRequest/Response.GetTypedHeaders() extensions to get direct accessors for common typed headers. See RequestHeaders and ResponseHeaders.
- Using the new parsers in the FormFeature code.
- Related PRs for StaticFiles, MVC, and Routing.

What this does not include:
- Full integration of the new cookie types with the old cookie infrastructure. I'll file a new work item for that.
- More relaxed parsing rules. We'll have to triage such issues on a case-by-case basis.

Related work items:
#163 - Remove WebUtility's dependency on Http
#19, #68 - Header parsers
#118 - Remove temp conneg types from MVC

@muratg @davidfowl @loudej @rynowak @GrabYourPitchforks

### DIFF
--- a/src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryTypeExtensions.cs
+++ b/src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryTypeExtensions.cs
@@ -1,0 +1,202 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Microsoft.AspNet.Http.Headers;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Http
+{
+    public static class HeaderDictionaryTypeExtensions
+    {
+        public static RequestHeaders GetTypedHeaders(this HttpRequest request)
+        {
+            return new RequestHeaders(request.Headers);
+        }
+
+        public static ResponseHeaders GetTypedHeaders(this HttpResponse response)
+        {
+            return new ResponseHeaders(response.Headers);
+        }
+
+        public static DateTimeOffset? GetDate([NotNull] this IHeaderDictionary headers, [NotNull] string name)
+        {
+            return headers.Get<DateTimeOffset?>(name);
+        }
+
+        public static void Set([NotNull] this IHeaderDictionary headers, [NotNull] string name, object value)
+        {
+            if (value == null)
+            {
+                headers.Remove(name);
+            }
+            else
+            {
+                headers[name] = value.ToString();
+            }
+        }
+
+        public static void SetList<T>([NotNull] this IHeaderDictionary headers, [NotNull] string name, IList<T> values)
+        {
+            if (values == null || values.Count == 0)
+            {
+                headers.Remove(name);
+            }
+            else
+            {
+                headers.SetValues(name, values.Select(value => value.ToString()).ToArray());
+            }
+        }
+
+        public static void SetDate([NotNull] this IHeaderDictionary headers, [NotNull] string name, DateTimeOffset? value)
+        {
+            if (value.HasValue)
+            {
+                headers[name] = HeaderUtilities.FormatDate(value.Value);
+            }
+            else
+            {
+                headers.Remove(name);
+            }
+        }
+
+        public static void Append([NotNull] this IHeaderDictionary headers, [NotNull] string name, [NotNull] object value)
+        {
+            headers.Append(name, value.ToString());
+        }
+
+        public static void AppendList<T>([NotNull] this IHeaderDictionary headers, [NotNull] string name, [NotNull] IList<T> values)
+        {
+            headers.AppendValues(name, values.Select(value => value.ToString()).ToArray());
+        }
+
+        private static IDictionary<Type, object> KnownParsers = new Dictionary<Type, object>()
+        {
+            { typeof(CacheControlHeaderValue), new Func<string, CacheControlHeaderValue>(value => { CacheControlHeaderValue result; return CacheControlHeaderValue.TryParse(value, out result) ? result : null; }) },
+            { typeof(ContentDispositionHeaderValue), new Func<string, ContentDispositionHeaderValue>(value => { ContentDispositionHeaderValue result; return ContentDispositionHeaderValue.TryParse(value, out result) ? result : null; }) },
+            { typeof(ContentRangeHeaderValue), new Func<string, ContentRangeHeaderValue>(value => { ContentRangeHeaderValue result; return ContentRangeHeaderValue.TryParse(value, out result) ? result : null; }) },
+            { typeof(MediaTypeHeaderValue), new Func<string, MediaTypeHeaderValue>(value => { MediaTypeHeaderValue result; return MediaTypeHeaderValue.TryParse(value, out result) ? result : null; }) },
+            { typeof(RangeConditionHeaderValue), new Func<string, RangeConditionHeaderValue>(value => { RangeConditionHeaderValue result; return RangeConditionHeaderValue.TryParse(value, out result) ? result : null; }) },
+            { typeof(RangeHeaderValue), new Func<string, RangeHeaderValue>(value => { RangeHeaderValue result; return RangeHeaderValue.TryParse(value, out result) ? result : null; }) },
+            { typeof(EntityTagHeaderValue), new Func<string, EntityTagHeaderValue>(value => { EntityTagHeaderValue result; return EntityTagHeaderValue.TryParse(value, out result) ? result : null; }) },
+            { typeof(DateTimeOffset?), new Func<string, DateTimeOffset?>(value => { DateTimeOffset result; return HeaderUtilities.TryParseDate(value, out result) ? result : (DateTimeOffset?)null; }) },
+            { typeof(long?), new Func<string, long?>(value => { long result; return HeaderUtilities.TryParseInt64(value, out result) ? result : (long?)null; }) },
+        };
+
+        private static IDictionary<Type, object> KnownListParsers = new Dictionary<Type, object>()
+        {
+            { typeof(MediaTypeHeaderValue), new Func<IList<string>, IList<MediaTypeHeaderValue>>(value => { IList<MediaTypeHeaderValue> result; return MediaTypeHeaderValue.TryParseList(value, out result) ? result : null; })  },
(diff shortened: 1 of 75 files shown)


Tratcher:
https://github.com/aspnet/StaticFiles/pull/24
https://github.com/aspnet/Routing/pull/139
https://github.com/aspnet/Mvc/pull/1801

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -0,0 +1,247 @@
...
+                throw new ArgumentException("An empty string is not allowed.", parameterName);
+            }
+
+            if (HttpRuleParser.GetTokenLength(value, 0) != value.Length)
+            {
+                throw new FormatException(string.Format(CultureInfo.InvariantCulture, "Invalid token '{0}.", value));
+            }
+        }
+
+        internal static void CheckValidComment(string value, string parameterName)
```
GrabYourPitchforks: Is this still used anywhere?

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -0,0 +1,247 @@
...
+            {
+                return false;
+            }
+
+            if (x.Count == 0)
+            {
+                return true;
+            }
+
+            // We have two unordered lists. So comparison is an O(n*m) operation which is expensive. Usually
```
GrabYourPitchforks: "Usually" - can malicious user input force n or m to be large?

src/Microsoft.Net.Http.Headers/HttpHeaderParser.cs
```diff
@@ -0,0 +1,137 @@
...
+            T result = default(T);
+            if (!TryParseValue(value, ref index, out result))
+            {
+                throw new FormatException(string.Format(CultureInfo.InvariantCulture, "Invalid value '{0}'.",
+                    value == null ? "<null>" : value.Substring(index)));
+            }
+            return result;
+        }
+
+        public virtual bool TryParseValues(IList<string> values, out IList<T> parsedValues)
```
GrabYourPitchforks: This doesn't require an out parameter since you're returning a reference type (IList) instead of a value type.  You could just return the parsed values directly.

src/Microsoft.Net.Http.Headers/HttpRuleParser.cs
```diff
@@ -0,0 +1,394 @@
...
+                if ((c == ' ') || (c == '\t'))
+                {
+                    current++;
+                    continue;
+                }
+
+                if (c == '\r')
+                {
+                    // If we have a #13 char, it must be followed by #10 and then at least one SP or HT.
+                    if ((current + 2 < input.Length) && (input[current + 1] == '\n'))
```
GrabYourPitchforks: Nit: You have defined consts for \r and \n but aren't using them here.

src/Microsoft.Net.Http.Headers/HttpRuleParser.cs
```diff
@@ -0,0 +1,394 @@
...
+            // All characters between startIndex and the end of the string are LWS characters.
+            return input.Length - startIndex;
+        }
+
+        internal static bool ContainsInvalidNewLine(string value)
+        {
+            return ContainsInvalidNewLine(value, 0);
+        }
+
+        internal static bool ContainsInvalidNewLine(string value, int startIndex)
```
GrabYourPitchforks: Do we need to check for \n without \r?  I realize it's allowable per spec, but is there any legitimate scenario for accepting such a thing or is it really an attempt to confuse the server into doing something strange?  I'm mainly concerned about UNIX scenarios where the environment newline character is \n instead of the \r\n combo.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -0,0 +1,408 @@
...
+                    }
+                    else
+                    {
+                        Parameters.Add(new NameValueHeaderValue(CharsetString, value));
+                    }
+                }
+            }
+        }
+
+        public Encoding Encoding
```
GrabYourPitchforks: Note to callers: Don't trust the Encoding coming from the client.  Enforce UTF8 or similar.  Do we need enforcement at this API level or in the caller (MVC)?

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -0,0 +1,408 @@
...
+            get
+            {
+                return _mediaType.Substring(_mediaType.IndexOf('/') + 1);
+            }
+        }
+
+        /// <summary>
+        /// MediaType = "*/*"
+        /// </summary>
+        public bool AllTypes
```
GrabYourPitchforks: Recommendation: should be "IsAllTypes" or similar since you're asking "is this X?"  Or "Matches", "Supports", etc.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -0,0 +1,408 @@
...
+            }
+
+            return (string.Compare(_mediaType, other._mediaType, StringComparison.OrdinalIgnoreCase) == 0) &&
+                HeaderUtilities.AreEqualCollections(_parameters, other._parameters);
+        }
+
+        public override int GetHashCode()
+        {
+            // The media-type string is case-insensitive.
+            return _mediaType.ToLowerInvariant().GetHashCode() ^ NameValueHeaderValue.GetHashCode(_parameters);
```
GrabYourPitchforks: Use StringComparer.OrdinalIgnoreCase.GetHashCode(string) instead.  It's closer to what you actually want and saves an allocation.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -0,0 +1,408 @@
...
+        }
+
+        public static MediaTypeHeaderValue Parse(string input)
+        {
+            var index = 0;
+            return SingleValueParser.ParseValue(input, ref index);
+        }
+
+        public static bool TryParse(string input, out MediaTypeHeaderValue parsedValue)
```
GrabYourPitchforks: Don't need an out parameter when the output type is a reference type.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -0,0 +1,347 @@
...
+                CheckValueFormat(value);
+                _value = value;
+            }
+        }
+
+        public override int GetHashCode()
+        {
+            Contract.Assert(_name != null);
+
+            var nameHashCode = _name.ToLowerInvariant().GetHashCode();
```
GrabYourPitchforks: StringComparer.OrdinalIgnoreCase.GetHashCode(string) instead of calling ToLowerInvariant.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -0,0 +1,347 @@
...
+        {
+            if ((values == null) || (values.Count == 0))
+            {
+                return 0;
+            }
+
+            var result = 0;
+            foreach (var value in values)
+            {
+                result = result ^ value.GetHashCode();
```
GrabYourPitchforks: Not a good hash function - consider doing what Tuple does instead when combining since order matters.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
Tratcher: This is called from SomeHeaderType.Equals. I could imagine one of the parameters coming from the client where they would be limited by header lengths, but I would expect the other to come from the application.

src/Microsoft.Net.Http.Headers/HttpHeaderParser.cs
Tratcher: Like this? `IList<T> TryParseValues(IList<string> values)` where null or empty means failure? I'd rather stick with the bool result to indicate failures.

src/Microsoft.Net.Http.Headers/HttpRuleParser.cs
Tratcher: Removed, this is no longer used.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
Tratcher: Won't fix for now.

src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryTypeExtensions.cs
```diff
@@ -0,0 +1,202 @@
...
+                .Where(methodInfo =>
+                {
+                    if (string.Equals("TryParse", methodInfo.Name, StringComparison.Ordinal)
+                        && methodInfo.ReturnParameter.ParameterType.Equals(typeof(bool)))
+                    {
+                        var methodParams = methodInfo.GetParameters();
+                        return methodParams.Length == 2
+                            && methodParams[0].ParameterType.Equals(typeof(string))
+                            && methodParams[1].IsOut
+                            && methodParams[1].ParameterType.Equals(type.MakeByRefType());
```
Eilon: Can create the `ref` type just once outside of the LINQ query. Could make the code a smidgen clearer.

src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
```diff
@@ -0,0 +1,702 @@
...
+                result = result.Replace(@"\", @"\\");
+                result = result.Replace(@"""", @"\""");
+                // Re-add quotes "value"
+                result = string.Format(CultureInfo.InvariantCulture, "\"{0}\"", result);
+            }
+            return result;
+        }
+
+        // Replaces characters not suitable for HTTP headers with '_' rather than MIME encoding them.
+        private string Sanatize(string input)
```
Eilon: Tpyo in method name.

Eilon:
Talked to @Tratcher about whether the header value collection returned by the APIs should be read/write but ignore changes (current behavior), throw on changes, or be "live" so he logged a bug on it. Conclusion is that they should probably throw on changes.

PR #166, opened 2015-01-15
Optimized performance of TryStringToDate via short-circuiting

ironyx:
Note that I did this without an IDE, but I'm 99% sure it compiles. ;)

Basically this will allow less work to be done if the input *clearly* does not conform to any known date length.

The max length has to be expanded by `7`. +1 per `H`, `m`, `s`, and `d` char and +3 for `zzz`.

`H` can be `0` to `23`
`m` can be `0` to `59`
`s` can be `0` to `59`
`d` can be `1` to `31`
`zzz` can be a lot of things, but at largest `+05:00` or `-05:00` type values

### DIFF
--- a/src/Microsoft.Net.Http.Headers/HttpRuleParser.cs
+++ b/src/Microsoft.Net.Http.Headers/HttpRuleParser.cs
@@ -89,6 +89,10 @@ namespace Microsoft.Net.Http.Headers
                     DateFormatsMaxLength = dateFormat.Length;
                 }
             }
+            
+            // DateTime formats: H, m, s, and d could all be 2 digits, so add one for each. 
+            // zzz can be 6 digits, so add 3
+            DateFormatsMaxLength += 7;
         }
 
         internal static bool IsTokenChar(char character)
--- a/src/Microsoft.Net.Http.Headers/HttpRuleParser.cs
+++ b/src/Microsoft.Net.Http.Headers/HttpRuleParser.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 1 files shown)


Tratcher:
Out of curiosity do you see many invalid dates in headers? In my experience they are mostly valid, but he formats vary.

ironyx:
I see a lot of attempted malformed inputs.

In this case it's sure to be more of an edge case, but in my opinion every cycle counts and this can save a few.

Eilon:
How much extra work is the optimization doing, though? Does it actually perform better? Checking for whitespace (which is kind of a totally bogus check because of all the crazy Unicode stuff it checks for) and computing string lengths aren't free either.

ironyx:
@Eilon you're right, but the cost of enumerating/iterating the 20 some-odd sets of strings and trying to parse each of them is surely more than a whitespace check (even including Unicode)

Eilon:
I think I'd be more comfortable with this if we had some data showing that there was a significant improvement to be had here in both average/common case scenarios (i.e. normal data), and in worst case scenarios (e.g. something potentially malicious). I think getting such data in terms of both a micro benchmark (how much faster/slower is this one method), as well as in terms of an overall request.

ironyx:
Well to be truthful, I don't have time right now to whip up benchmarks for it.

I can guarantee you that a single `IsNullOrWhitespace` check against a single input is much cheaper than the attempted parse of an input against 20-some odd format strings. This of course depends on how frequently you receive malformed input. Think about it from the perspective of any website choosing to adopt vNext. They may be attacked with malformed headers (we are constantly).

If efficiency isn't your top concern, then that's acceptable (it isn't my code to change). However, given that you've accepted many PRs from us in the past in the name of efficiency, I would say that you do value it. If you'd like me to change it in some way to make you more comfortable, let me know. Otherwise I'll leave it as is; ignore it as you please.

Eilon:
I think my "spidey sense" for this change just isn't going off in terms of being obviously more performant. I think in cases where the algorithms are identical but implemented more efficiently it's easier to see the improvement (e.g. avoiding boxing, using smarter string compares, etc.). But in this case the algorithm is changed, and while I don't doubt it's correctness, it's difficult for me to see the definite speed-up in average case and also somewhat in worst case.

Performance and efficiency are certainly very important, but so is code maintainability, and there's of course always a trade-off between these two. That's why we generally measure first, propose a change, and measure again. We want to be as certain as possible that perf improvements actually make the app perform better, while hopefully not making the code any more (or not significantly more) difficult to maintain.

ironyx:
That's a good reply, thank you. I can see where you're coming from fully now. I'd say hold off on it until it's an issue performance wise in this case. I'll close it.

Eilon:
Understood, and thank you for the discussion as well! We certainly look forward to seeing future contributions as well :smile:

Issue #162, opened 2015-01-10
Rename Microsoft.AspNet.PipelineCore


Tratcher:

davidfowl:
Yes, please. Can we do this before in beta3? /cc @Eilon @Tratcher @muratg ?

Eilon:
Fine with me. What's the new name?

Eilon:
We should get as many renames as possible done for beta3.

Tratcher:
Ms.AspNet....
Http.Impl
Http.Core

Eilon:
`Microsoft.AspNet.Http.Core` doesn't seem terrible to me. That's kind of what's in there... the core types. Or `...Intrinsics`, but I hate that word.

Tratcher:
That's the funny thing, Http contains the abstract intrinsics, PipelineCore contains the actual implementation of said intrinsics.

While we're at it are there alternatives to calling all of the implementation types Default*?

Eilon:
Instead of `Default*` just call them `*`..?

Tratcher:
E.g. DefaultHttpContext implements the abstract HttpContext.

Eilon:
O I C. I guess in this case the `Default*` is kind of fine because no one aside from us ever sees those types. Everyone programs against the friendly-named types, which is most of what we care about...

Tratcher:
DefaultHttpContext is often used in unit tests, but again, mostly in our own tests and less in other people's.

davidfowl:
We shouldn't call it something obscure though.

Eilon:
Default isn't very obscure so I think it's fine...

Now, back to the assembly/project name... Http.Core?

PR #170, opened 2015-01-15
Change PipelineCore namespace to Http.Core. Part-1.

Tratcher:
#162 
This is part 1 to break the namespace in the existing package and flush out who is depending on it. Part 2 will rename the package itself.

@muratg @davidfowl @Eilon


Tratcher:
@davidfowl said :shipit: offline.

Issue #175, opened 2015-01-19
File upload with non-English chars doesn't work


yishaigalatzer:
Tried on both IE and Chrome and they encode the Disposition header (in the multi part body) in UTF-8, out multi-part parser assumes ASCII and hence these files names are mangled.

Original issue - aspnet/Mvc#1840

Tratcher:
The utf-8 decoding needs to happen here:
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.WebUtilities/BufferedReadStream.cs#L357

Tratcher:
https://code.google.com/p/dart/issues/detail?id=14303
Chrome Version 41.0.2272.3 dev-m (64-bit)
With `<meta charset="UTF-8" />`:
 - Chrome: as utf-8
 - IE: as utf-8

Without `<meta charset="UTF-8" />`:
 - Chrome: multi-bytes replaced with `?`
 - IE: as utf-8

PR #178, opened 2015-01-21
Decode multipart headers as UTF-8.

Tratcher:
#175 
@yishaigalatzer @muratg @GrabYourPitchforks @Eilon 

Please consider for Beta3.


blowdart:
Please add tests for unicode wide characters, especially invalid surrogate pairs.

GrabYourPitchforks:
:shipit:

src/Microsoft.AspNet.WebUtilities/BufferedReadStream.cs
```diff
@@ -375,6 +367,13 @@ private void ProcessLineChar(StringBuilder builder, ref bool foundCR, ref bool f
             }
         }
 
+        private string DecodeLine(MemoryStream builder, bool foundCRLF)
+        {
+            // Drop the final CRLF, if any
+            var length = foundCRLF ? builder.Length - 2 : builder.Length;
```
yishaigalatzer: could there be just CR at the end unix style?

src/Microsoft.AspNet.WebUtilities/BufferedReadStream.cs
```diff
@@ -321,19 +322,15 @@ public string ReadLine(int lengthLimit)
...
-            // Stream ended with no CRLF.
-            return builder.ToString();
+            return DecodeLine(builder, foundCRLF);
         }
 
         public async Task<string> ReadLineAsync(int lengthLimit, CancellationToken cancellationToken)
         {
             CheckDisposed();
-            StringBuilder builder = new StringBuilder();
+            var builder = new MemoryStream(200);
```
yishaigalatzer: is the limitation of 200 enforced elsewhere? Any reason not to just construct with lengthlimit + 1?

Tratcher: 200 isn't a limit, it's the initial buffer size. I expect most normal header lines to fit within this.

src/Microsoft.AspNet.WebUtilities/BufferedReadStream.cs
Tratcher: No, HTTP and MIME are very specific about requiring CRLF.

Tratcher:
Merged.

Issue #174, opened 2015-01-19
Define common status code constants


Tratcher:
Enum was always a poor type for Http status codes (e.g. System.Net.HttpStatusCode). Static/const ints may be more appropriate. We recently added the following for common header names: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.Net.Http.Headers/HeaderNames.cs. Something similar for status codes would be reasonable.

shawnwildermuth:
Doesn't matter to me whether it is Enum or static/const, just don't want dev's having to use specific numbers (especial sub-numbers like 403.1)

PR #180, opened 2015-01-27
Constants for status codes, lookup for reason phrases.

Tratcher:
#174

### DIFF
--- a/src/Microsoft.AspNet.WebUtilities/StatusCodes.cs
+++ b/src/Microsoft.AspNet.WebUtilities/StatusCodes.cs
@@ -1,0 +1,54 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.WebUtilities
+{
+    public static class StatusCodes
+    {
+        public const int Status200OK = 200;
+        public const int Status201Created = 201;
+        public const int Status202Accepted = 202;
+        public const int Status203NonAuthoritative = 203;
+        public const int Status204NoContent = 204;
+        public const int Status205ResetContent = 205;
+        public const int Status206PartialContent = 206;
+
+        public const int Status300MultipleChoices = 300;
+        public const int Status301MovedPermanently = 301;
+        public const int Status302Found = 302;
+        public const int Status303SeeOther = 303;
+        public const int Status304NotModified = 304;
+        public const int Status305UseProxy = 305;
+        public const int Status306SwitchProxy = 306;
+        public const int Status307TemporaryRedirect = 307;
+
+        public const int Status400BadRequest = 400;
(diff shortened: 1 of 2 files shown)


Eilon:
Looks lovely :shipit:

src/Microsoft.AspNet.WebUtilities/StatusCodes.cs
```diff
@@ -0,0 +1,54 @@
...
+        public const int Status405MethodNotAllowed = 405;
+        public const int Status406NotAcceptable = 406;
+        public const int Status407ProxyAuthenticationRequired = 407;
+        public const int Status408RequestTimeout = 408;
+        public const int Status409Conflict = 409;
+        public const int Status410Gone = 410;
+        public const int Status411LengthRequired = 411;
+        public const int Status412PreconditionFailed = 412;
+        public const int Status413RequestEntityTooLarge = 413;
+        public const int Status414RequestUurTooLong = 414;
```
PinpointTownes: `Uur`?

src/Microsoft.AspNet.WebUtilities/StatusCodes.cs
```diff
@@ -0,0 +1,54 @@
...
+        public const int Status415UnsupportedMediaType = 415;
+        public const int Status416RequestedRangeNotSatisfiable = 416;
+        public const int Status417ExpectationFailed = 417;
+        public const int Status418ImATeapot = 418;
```
PinpointTownes: I love that one, but is it really useful? :smile:

davidfowl:
:shipit:

PinpointTownes:
My 2 cents: the `[Status] + code + name` pattern is a bit hard to read and I'm not sure it's really useful: if you know the status code, you don't need to rely on a constant to use it. And if you wanna use a constant, that's often because you don't want to play with the status code directly.

davidfowl:
@PinpointTownes has a point, should me model it after the crazy System.Net enum?

Tratcher:
The enum naming assumed you (and everyone reading your code) were deeply familiar with the reason phrases and what the resulting status codes would be. If you just use reason phrases people have to guess the matching status code, where if you just use integers they have to guess the matching reason phrase.

I picked the `Status404NotFound` pattern because it lets you see both parts together for validation. It also lets you use intellisense like a lookup table where you can type in either the status code or the reason phrase and it immediately shows you the other half. It has `Status` on there because you can't start with digits.

Would you prefer `NotFound404`? I find that a little odd because it's the opposite of how it shows up in the response and spec. How about `S404NotFound`?

Eilon:
I like it just fine the way it is. I view the point of these as being to eliminate what *appear* to be magic numbers from the code. And I think this does that beautifully.

Eilon:
And for the record, I *strongly* dislike the System.Net StatusCode enum because it adds no value at all. It just takes one hard to understand piece of information and hides it behind another hard to understand piece of information.

PR #187, opened 2015-02-08
Add HtmlEncode, UrlEncode, and JavaScriptStringEncode

GrabYourPitchforks:
Add HtmlEncode, UrlEncode, and JavaScriptStringEncode.
This is based loosely on the Anti-XSS encoder from WPL.

[Edited from original description.]

/cc @Eilon @Tratcher


GrabYourPitchforks:
Quick stats:
- This gets around 2x the throughput of WebUtility.HtmlEncode for inputs which don't require any chars to be encoded.
- This gets around 3x the throughput of WebUtility.HtmlEncode for inputs in which at least one char must be encoded.
- The memory utilization is 8KB static data (per AD) + 8KB per HtmlEncoder instance. Compare to Anti-XSS, which utilizes approx 2-3 MB static.

It is intended that a single HtmlEncoder instance be created and reused, but applications can create as many instances as they wish in practice. (The ctor is a little slow, so I wouldn't encourage this, but it's not the end of the world if this happens.) I can also consider adding a HtmlEncoder.Instance static convenience property if developers would find it useful.

dougbu:
like the `Instance` approach.  we create lots of `DefaultHtmlGenerator` instances that currently depend on `WebUtility.HtmlEncode()`.  would simplify the switch-over if they could just call `HtmlEncoder.Instance.HtmlEncode()`.

GrabYourPitchforks:
@dougbu Ideally framework developers wouldn't use these APIs directly but would instead accept an instance of an IHtmlEncoder service. That allows application developers to change the default encoder system-wide.

dougbu:
ah, hadn't noticed the service.  much more sensible.

src/Microsoft.AspNet.WebUtilities/Encoders/HtmlEncoder.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        /// <summary>
+        /// A default instance of the HtmlEncoder, equivalent to allowing only
+        /// the 'Basic Latin' character range.
+        /// </summary>
+        public static HtmlEncoder Default
+        {
+            get
+            {
+                HtmlEncoder defaultEncoder = Volatile.Read(ref _defaultEncoder);
```
davidfowl: Any reason you didn't use LazyInitializer?

src/Microsoft.AspNet.WebUtilities/Encoders/HtmlEncoder.cs
```diff
@@ -0,0 +1,229 @@
...
+            // The worst case encoding is 8 output chars per input char: [input] U+FFFF -> [output] "&#xFFFF;"
+            // We don't need to worry about astral code points since they consume *two* input chars to
+            // generate at most 10 output chars ("&#x10FFFF;"), which equates to 5 output per input.
+            int numCharsWhichMayRequireEncoding = input.Length - idxOfFirstCharWhichRequiresEncoding;
+            int sbCapacity = checked(idxOfFirstCharWhichRequiresEncoding + EncoderCommon.GetCapacityOfOutputStringBuilder(numCharsWhichMayRequireEncoding, worstCaseOutputCharsPerInputChar: 8));
+            Debug.Assert(sbCapacity >= input.Length);
+
+            // Allocate the StringBuilder with the first (known to not require encoding) part of the input string,
+            // then begin encoding from the last (potentially requiring encoding) part of the input string.
+            StringBuilder builder = new StringBuilder(input, 0, idxOfFirstCharWhichRequiresEncoding, sbCapacity);
```
davidfowl: More var

src/Microsoft.AspNet.WebUtilities/Encoders/HtmlEncoder.cs
GrabYourPitchforks: Avoids static factory allocation for a code path that will be used at most once per app.

davidfowl: Is there anywhere else we should be using this pattern? Or is it more a case by case thing? 
 /cc @Eilon

GrabYourPitchforks: Very much a case-by-case thing that depends on how many occurrences of LazyInitializer there are and how often the accessor is likely to be called. For instance, I just stuck with a standard LazyInitializer on UnicodeFilters.None for simplicity.

davidfowl:
We should file a bug to introduce some IHtmlEncoder interface somewhere for app consumption in the higher layers.

src/Microsoft.AspNet.WebUtilities/Encoders/UnicodeHelpers.cs
```diff
@@ -0,0 +1,164 @@
...
+            }
+        }
+
+        private static int GetScalarValueFromUtf16SurrogatePair(char highSurrogate, char lowSurrogate)
+        {
+            Debug.Assert(Char.IsHighSurrogate(highSurrogate));
+            Debug.Assert(Char.IsLowSurrogate(lowSurrogate));
+
+            // See http://www.unicode.org/versions/Unicode6.2.0/ch03.pdf, Table 3.5 for the
+            // details of this conversion. We don't use Char.ConvertToUtf32 because its exception
```
davidfowl: This is interesting. We should file a bug on github.com/dotnet/coreclr

GrabYourPitchforks: Likely not worthwhile. Their API is perfectly fine for its intended use case; our use case is slightly different.

src/Microsoft.AspNet.WebUtilities/Encoders/HexUtil.cs
```diff
@@ -0,0 +1,26 @@
...
+using System;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.AspNet.WebUtilities.Encoders
+{
+    /// <summary>
+    /// Contains helpers for dealing with byte-hex char conversions.
+    /// </summary>
+    internal static class HexUtil
```
dougbu: nit: why not `public` in an `...Internals` namespace?

src/Microsoft.AspNet.WebUtilities/Encoders/HexUtil.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        /// <summary>
+        /// Converts a number 0 - 15 to its associated hex character '0' - 'F'.
+        /// </summary>
+        /// <returns></returns>
```
dougbu: fill in `<returns/>`

src/Microsoft.AspNet.WebUtilities/Encoders/HtmlEncoder.cs
```diff
@@ -0,0 +1,229 @@
...
+        /// </summary>
+        public HtmlEncoder()
+            : this(UnicodeFilters.BasicLatin)
+        {
+        }
+
+        /// <summary>
+        /// Instantiates an encoder using a custom allow list of characters.
+        /// </summary>
+        public HtmlEncoder(params ICodePointFilter[] filters)
```
dougbu: nit: add `<param name="filters"/>` doc

src/Microsoft.AspNet.WebUtilities/Encoders/HtmlEncoder.cs
```diff
@@ -0,0 +1,229 @@
...
+            {
+                if (!IsCharacterAllowed(value[i]))
+                {
+                    return HtmlEncodeImpl(value, i);
+                }
+            }
+            return value;
+        }
+
+        private string HtmlEncodeImpl(string input, int idxOfFirstCharWhichRequiresEncoding)
```
dougbu: nit: toward avoiding abbreviations, how about `HtmlEncodeCore()`?

src/Microsoft.AspNet.WebUtilities/Encoders/HtmlEncoder.cs
```diff
@@ -0,0 +1,229 @@
...
+        {
+            Debug.Assert(idxOfFirstCharWhichRequiresEncoding >= 0);
+            Debug.Assert(idxOfFirstCharWhichRequiresEncoding < input.Length);
+
+            // The worst case encoding is 8 output chars per input char: [input] U+FFFF -> [output] "&#xFFFF;"
+            // We don't need to worry about astral code points since they consume *two* input chars to
+            // generate at most 10 output chars ("&#x10FFFF;"), which equates to 5 output per input.
+            int numCharsWhichMayRequireEncoding = input.Length - idxOfFirstCharWhichRequiresEncoding;
+            int sbCapacity = checked(idxOfFirstCharWhichRequiresEncoding + EncoderCommon.GetCapacityOfOutputStringBuilder(numCharsWhichMayRequireEncoding, worstCaseOutputCharsPerInputChar: 8));
```
dougbu: nit: wrap long line

src/Microsoft.AspNet.WebUtilities/Encoders/HtmlEncoder.cs
```diff
@@ -0,0 +1,229 @@
...
+                    if (IsCharacterAllowed(c))
+                    {
+                        builder.Append(c);
+                    }
+                    else
+                    {
+                        if (c == '<') { builder.Append("&lt;"); }
+                        else if (c == '>') { builder.Append("&gt;"); }
+                        else if (c == '&') { builder.Append("&amp;"); }
+                        else if (c == '\"') { builder.Append("&quot;"); }
```
dougbu: should this map `'\''` to `"&apos;"`?

src/Microsoft.AspNet.WebUtilities/Encoders/UnicodeHelpers.cs
```diff
@@ -0,0 +1,164 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace Microsoft.AspNet.WebUtilities.Encoders
```
dougbu: nits:
- a few long lines in this class can be wrapped
- needs a bit more `var`.  same for last couple of methods in `UnicodeFilters`

test/Microsoft.AspNet.WebUtilities.Tests/Encoders/EncoderCommonTests.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Xunit;
+
+namespace Microsoft.AspNet.WebUtilities.Encoders
+{
+    public class EncoderCommonTests
```
dougbu: nit: guidelines say `...Test` (singular)

test/Microsoft.AspNet.WebUtilities.Tests/Encoders/UnicodeFiltersTests.cs
```diff
@@ -0,0 +1,984 @@
...
+            RunRangeTestImpl(UnicodeFilters.HalfwidthandFullwidthForms, first: '\uFF00', last: '\uFFEF');
+        }
+
+        [Fact]
+        public void Range_Specials()
+        {
+            RunRangeTestImpl(UnicodeFilters.Specials, first: '\uFFF0', last: '\uFFFF');
+        }
+
+        private static void RunRangeTestImpl(ICodePointFilter filter, char first, char last)
```
dougbu: nit: just `RunRangeTest()`

test/Microsoft.AspNet.WebUtilities.Tests/Encoders/HtmlEncoderTests.cs
```diff
@@ -0,0 +1,297 @@
...
+
+namespace Microsoft.AspNet.WebUtilities.Encoders
+{
+    public class HtmlEncoderTests
+    {
+        [Fact]
+        public void Ctor_WithCustomFilters()
+        {
+            // Arrange
+            CustomCodePointFilter filter1 = new CustomCodePointFilter('a', 'b');
```
dougbu: nit: test methods in general aren't `var`ish enough

test/Microsoft.AspNet.WebUtilities.Tests/Encoders/HtmlEncoderTests.cs
```diff
@@ -0,0 +1,297 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using Xunit;
+
+namespace Microsoft.AspNet.WebUtilities.Encoders
+{
+    public class HtmlEncoderTests
+    {
+        [Fact]
+        public void Ctor_WithCustomFilters()
```
dougbu: nit: avoid abbreviations even for `Constructor_...`

dougbu:
:watch:

src/Microsoft.AspNet.WebUtilities/Encoders/HtmlEncoder.cs
GrabYourPitchforks: No - the 'apos' entity is undefined in HTML 4, so most encoders maintain compat by not emitting the entity.

src/Microsoft.AspNet.WebUtilities/Encoders/HexUtil.cs
Eilon: This is an internal API, it doesn't need any docs at all...

PR #189, opened 2015-02-10
Logical AND would never be true; broke into OR statement

ironyx:


msotclas:
Hi __@ironyx__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <p>The agreement was validated by Microsoft Open Technologies, Inc. and real humans are currently evaluating your PR.</p>

TTYL, MSOTBOT;

src/Microsoft.Net.Http.Headers/CookieHeaderParser.cs
```diff
@@ -91,7 +91,7 @@ private static int GetNextNonEmptyOrWhitespaceIndex(string input, int startIndex
 
             if (skipEmptyValues)
             {
-                while ((current < input.Length) && (input[current] == ',') && (input[current] == ';'))
+                while ((current < input.Length) && ((input[current] == ',') || (input[current] == ';')))
```
Eilon: Woah, how weird is that?! @Tratcher does this also mean that there's a bug in the original code this was borrowed from?

Tratcher: This method is copied and customized from the generic version over here: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.Net.Http.Headers/HeaderUtilities.cs#L188

This just looks like a typo in the customization.  We haven't started consuming this new class anywhere.

Eilon: @Tratcher ah good to know, I will merge.

PR #191, opened 2015-02-13
Initial set of AuthenticationType -> AuthenticationScheme renames

HaoK:
- Renamed IAuthTypesContext??? -> IDescribeSchemesContext
- Replaced ClaimsIdentity -> ClaimsPrincipal in context classes

TBD: Still need to work through Sign in with more than one scheme flow, and removing old SignIn(ClaimsIdentity methods)

cc @lodejard @Tratcher @Praburaj @blowdart


msotclas:
Hi __@HaoK__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Hao Kung). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

src/Microsoft.AspNet.Http/Security/AuthenticateResult.cs
```diff
@@ -15,24 +13,21 @@ public class AuthenticationResult
         /// <summary>
         /// Create an instance of the result object
         /// </summary>
-        /// <param name="identity">Assigned to Identity. May be null.</param>
+        /// <param name="principal">Assigned to Principal. May be null.</param>
         /// <param name="properties">Assigned to Properties. Contains extra information carried along with the identity.</param>
         /// <param name="description">Assigned to Description. Contains information describing the authentication provider.</param>
-        public AuthenticationResult(IIdentity identity, [NotNull] AuthenticationProperties properties, [NotNull] AuthenticationDescription description)
+        public AuthenticationResult(ClaimsPrincipal principal, [NotNull] AuthenticationProperties properties, [NotNull] AuthenticationDescription description)
```
Tratcher: Does this need to include the AuthenticationScheme so you can cross-reference  when requesting multiples?

Tratcher:
@brentschmaltz

src/Microsoft.AspNet.Http/Security/AuthenticateResult.cs
HaoK: I kept the old behavior, where that information is in the AuthenticationDescription.AuthenticationScheme still

HaoK:
Switched ClaimsIdentity -> ClaimsPrincipal, initial stab at SignIn, only takes a single ClamsPrincipal to start (need to revisit Accept logic since I think I probably broke it for more than one authentication scheme)

HaoK:
All tests are passing though (so if anything is broken at this layer, we should add new tests)

src/Microsoft.AspNet.Http.Core/DefaultHttpResponse.cs
```diff
@@ -129,55 +126,58 @@ public override void Redirect(string location, bool permanent)
...
-            IEnumerable<string> leftovers = authenticationTypes.Except(challengeContext.Accepted);
+            IEnumerable<string> leftovers = authenticationSchemes.Except(challengeContext.Accepted);
             if (leftovers.Any())
             {
                 throw new InvalidOperationException("The following authentication types were not accepted: " + string.Join(", ", leftovers));
             }
         }
 
-        public override void SignIn(AuthenticationProperties properties, [NotNull] IEnumerable<ClaimsIdentity> identities)
+        public override void SignIn(string authenticationScheme, [NotNull] ClaimsPrincipal principal, AuthenticationProperties properties)
```
lodejard: With this, we've lost the ability to SignIn two schemes at once. That's probably okay if we verify there are no negative side-effects to calling SignIn twice.

HaoK: @lodejard Yeah this was intentional simplification to start with, I'll add some new tests to verify things like SignIn, SignOut, SignIn weaved together do the expected thing.  If we decide to stick to one scheme sign in, should we also nuke the SignOut overloads that take a bunch of schemes at once for consistency?

test/Microsoft.AspNet.Http.Core.Tests/DefaultHttpContextTests.cs
```diff
@@ -78,6 +77,58 @@ public void SignOutWithNoAuthMiddlewareMayThrow()
             Assert.Throws<InvalidOperationException>(() => context.Response.SignOut("Foo"));
         }
 
+        [Fact]
+        public void SignInOutIn()
```
Praburaj: That's good test name :smile:

HaoK:
Updated PR removing overloads of SignOut that take more than one scheme, calling SignOut() or with an empty or null scheme is equivalent to the old behavior of sign out the default/active scheme

HaoK:
Merged de1e8763dd3fbd4ca94aa44171aa34741c536c66

PR #192, opened 2015-02-16
Provide a facility for registering encoder services

GrabYourPitchforks:
/cc @anurse

### DIFF
--- a/src/Microsoft.AspNet.Http.Extensions/Encoders/EncoderServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNet.Http.Extensions/Encoders/EncoderServiceCollectionExtensions.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.WebUtilities.Encoders;
+using Microsoft.Framework.ConfigurationModel;
+
+namespace Microsoft.Framework.DependencyInjection
+{
+    public static class EncoderServiceCollectionExtensions
+    {
+        public static IServiceCollection AddEncoders([NotNull] this IServiceCollection services)
+        {
+            return AddEncoders(services, configuration: null);
+        }
+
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Http.Extensions/Encoders/EncoderServices.cs
```diff
@@ -0,0 +1,24 @@
...
+using System;
+using System.Collections.Generic;
+using Microsoft.Framework.ConfigurationModel;
+using Microsoft.Framework.DependencyInjection;
+
+namespace Microsoft.AspNet.WebUtilities.Encoders
+{
+    public static class EncoderServices
+    {
+        public static IEnumerable<IServiceDescriptor> GetDefaultServices(IConfiguration configuration = null)
```
davidfowl: No more optional parameters.

GrabYourPitchforks: Would the parameterless overload just call the parameterized overload passing null?

davidfowl: Good question, yes pass null.

GrabYourPitchforks: Got it - thanks. :)

src/Microsoft.AspNet.Http.Extensions/Encoders/EncoderServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Http;
+using Microsoft.Framework.ConfigurationModel;
+using Microsoft.Framework.DependencyInjection;
+
+namespace Microsoft.AspNet.WebUtilities.Encoders
```
Praburaj: in `Microsoft.Framework.DependencyInjection` namespace like rest other `IServiceCollection` extensions?

GrabYourPitchforks: Got it. :)

Praburaj:
@ajaybhargavb

Issue #179, opened 2015-01-23
IsSecure is badly named


blowdart:
Request.IsSecure returns true if the scheme is HTTPS. 

HTTS is not a guarantee of security, but merely that the request cannot be eavesdropped (and even then, given null ciphers, this is not actually true)

The the property name is disingenuous and should be removed, or changed to IsHttpsConnection.

bitcrazed:
I agree. The statement IsSecure has a whole lot of connotations [some of which may not be true](http://www.zdnet.com/article/gogo-in-flight-wi-fi-serving-spoofed-ssl-certificates/), even when HTTPS is used.

GrabYourPitchforks:
Related commits:
https://github.com/aspnet/Security/commit/a15cb4ffe559d2c7bb2f0e748e4a10446bd530a6
https://github.com/aspnet/Mvc/commit/cf36a601c61d5aa3b21d3f784b7198b5b4bdfa37

bitcrazed:
Nice! Thanks for taking our concerns into consideration.

PR #194, opened 2015-02-17
Rename HttpRequest.IsSecure -> HttpRequest.IsHttps

GrabYourPitchforks:
Fixes https://github.com/aspnet/HttpAbstractions/issues/179
/cc @blowdart @Tratcher


blowdart:
:shipit:

src/Microsoft.AspNet.Http/HttpRequest.cs
```diff
@@ -32,7 +32,7 @@ public abstract class HttpRequest
         /// Returns true if the owin.RequestScheme is https.
```
Eilon: Can you fix the doc comments in here too? Lots of reference to OWIN.

Issue #193, opened 2015-02-17
Consider moving the Encoders to their own package


GrabYourPitchforks:
Right now the core interfaces and implementations are in WebUtilities. However, since this package has no external dependencies, we can't add the DI-friendly `AddEncoders()` extension method here. Currently this extension method lives in Microsoft.AspNet.Http.Extensions.

We should consider whether the encoders need to be moved to their own package for better separation of concerns.

PR #199, opened 2015-02-22
Refactor Encoders into their own package

GrabYourPitchforks:
Better refactoring per https://github.com/aspnet/HttpAbstractions/issues/193


Eilon:
I approve this message :shipit:

davidfowl:
What about Microsoft.AspNet.Encoders

Eilon:
I'm kind of a fan of the original longer name, though I admit I can't think of what other kind of encoders we'd ever have that wouldn't be "web utilities." Then again, these encoders *are* web utilities...

GrabYourPitchforks:
If they're in the AspNet namespace instead of the Framework namespace, that would imply web, no?

davidfowl:
That's an interesting one. Microsoft.Framework.WebEncoders?

Eilon:
`Microsoft.Framework.WebEncoders` --> that's not bad either. This really has nothing to do with ASP.NET does it? (ASP.NET = server web apps)

GrabYourPitchforks:
WebEncoders works. Leave it in the HttpAbstractions repo for simplicity?

davidfowl:
Leave it there for now.

GrabYourPitchforks:
The most recent commit has the new names.

Issue #117, opened 2014-08-20
Support owin.RequestId


Tratcher:
https://github.com/owin/owin/issues/8
https://katanaproject.codeplex.com/workitem/297

Tratcher:
1. http.sys already has something we can use.
2. We need to find a feature interface to put this on internally.
3. this will be useful in logging.

Praburaj:
The closest one that I can find where we can add to the existing interface is `IHttpConnectionFeature`. or we can add an interface `IHttpRequestIdentifierFeature`.

Praburaj:
Fixed : https://github.com/aspnet/HttpAbstractions/commit/ae23f7c7bc956d40040c7a7a45aa2dedd293ebc4

We now have a feature `IRequestIdentifierFeature` to give this information.

PR #210, opened 2015-03-03
Adding a feature to get the traceidentifier for a request

Praburaj:
Addresses : https://github.com/aspnet/HttpAbstractions/issues/117

Related changes in Helios & weblistener in separate PRs.

@Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Http.Interfaces/IRequestIdentifierFeature.cs
+++ b/src/Microsoft.AspNet.Http.Interfaces/IRequestIdentifierFeature.cs
@@ -1,0 +1,18 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Http.Interfaces
+{
+    /// <summary>
+    /// Feature to identify a request.
+    /// </summary>
+    public interface IRequestIdentifierFeature
+    {
+        /// <summary>
+        /// Identifier to trace a request.
+        /// </summary>
+        Guid TraceIdentifier { get; }
+    }
+}
--- a/src/Microsoft.AspNet.Owin/OwinEnvironment.cs
+++ b/src/Microsoft.AspNet.Owin/OwinEnvironment.cs
@@ -80,7 +80,7 @@ namespace Microsoft.AspNet.Owin
                 { OwinConstants.Security.User, new FeatureMap<IHttpAuthenticationFeature>(feature => feature.User,
                     ()=> null, (feature, value) => feature.User = Utilities.MakeClaimsPrincipal((IPrincipal)value),
                     () => new HttpAuthenticationFeature())
-                },
+                }
             };
 
             // owin.CallCancelled is required but the feature may not be present.
@@ -113,6 +113,9 @@ namespace Microsoft.AspNet.Owin
             }
 
             _context.Items[typeof(HttpContext).FullName] = _context; // Store for lookup when we transition back out of OWIN
+
+            // The request identifier is a string per the spec.
+            _entries[OwinConstants.RequestId] = new FeatureMap<IRequestIdentifierFeature>(feature => feature.TraceIdentifier.ToString());
         }
 
         // Public in case there's a new/custom feature interface that needs to be added.
@@ -369,7 +372,7 @@ namespace Microsoft.AspNet.Owin
             }
 
             public FeatureMap(Func<TFeature, object> getter, Func<object> defaultFactory, Action<TFeature, object> setter)
-                : base(typeof(TFeature), feature => getter((TFeature)feature), defaultFactory,(feature, value) => setter((TFeature)feature, value))
+                : base(typeof(TFeature), feature => getter((TFeature)feature), defaultFactory, (feature, value) => setter((TFeature)feature, value))
             {
             }
 
--- a/src/Microsoft.AspNet.Owin/OwinFeatureCollection.cs
+++ b/src/Microsoft.AspNet.Owin/OwinFeatureCollection.cs
@@ -32,7 +32,8 @@ namespace Microsoft.AspNet.Owin
         IHttpRequestLifetimeFeature,
         IHttpAuthenticationFeature,
         IHttpWebSocketFeature,
-        IOwinEnvironmentFeature
+        IOwinEnvironmentFeature,
+        IRequestIdentifierFeature
     {
         public IDictionary<string, object> Environment { get; set; }
         private bool _headersSent;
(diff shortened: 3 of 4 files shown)


msotclas:
Hi __@Praburaj__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Praburaj Thiagarajan). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

GrabYourPitchforks:
:shipit:
As an aside, do we need a feature along the lines of "tell me the exact time the request started"? These two capabilities usually come hand-in-hand.

Praburaj:
Sure @GrabYourPitchforks. We can add a DateTime field to this interface having that information.

src/Microsoft.AspNet.Http.Interfaces/IRequestIdentifierFeature.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Feature to identify a request.
+    /// </summary>
+    public interface IRequestIdentifierFeature
+    {
+        /// <summary>
+        /// Identifier to trace a request.
+        /// </summary>
+        Guid TraceIdentifier { get; }
```
Eilon: Any reason this is specifically a Guid and not a string? (The OWIN spec doesn't specify a format.)

Eilon: BTW I'm not suggesting it *should* be a string, just asking.

davidfowl:
Any reason it's a guid instead of a string?

src/Microsoft.AspNet.Http.Interfaces/IRequestIdentifierFeature.cs
GrabYourPitchforks: Traditionally these have been GUIDs so that they can be flowed via the CorrelationManager.

Eilon: Yeah that makes sense... but not really :smile: I'm not really opposed to it being a GUID, it just seems so arbitrary (even in CorrelationManager). I mean what's so special about GUIDs anyway - a correlation id is logically an opaque value. So honestly I'm fine either way.

GrabYourPitchforks: GUIDs actually solve the problem nicely because they have a definite ordering so can be indexed. You can't do that with strings without a lot of ceremony: making sure everything is length-prefixed in storage (nvarchar, but then you kill indexability), making sure everybody agrees on which comparer to use (good luck doing this in SQL), and so on.

davidfowl: This should be a string

davidfowl: I can't think of anything else we have in our entire stake that represents an Id that's a guid.

Eilon: I'll concede that GUIDs are right from a *practical* perspective; but academically they are wrong from a *logical* perspective. (There's no absolute rule that says GUIDs are supposed to be ordered.)

GrabYourPitchforks: That they have a fixed, unambiguous, universal representation is a huge plus over strings, even if ordering doesn't matter. I agree that *academically* the difference shouldn't matter, but *practically* it does. I guarantee you'll have a difficult time building a reliable, high-performance correlation engine if your key is an arbitrary structureless string.

davidfowl: Most of the time these Ids are persisted as a string always so I'm not sure what you mean. Are you saying if we don't use the .NET Guid type that it won the performant?

GrabYourPitchforks: The nearly 100% use case is that they're persisted in an ETW log or similar structure, so they're definitely not persisted as strings. Please reread my argument above - I'm not advocating GUIDs for the sake of having GUIDs, I'm advocating it because it's structured, unambiguous, and easy for correlation engines to reason about. If you have another structure in mind that meets these requirements then by all means suggest it.

Praburaj: @Eilon - Just as a side note. `OWIN` spec actually says `owin.RequestId` is a `string`. 

>3.2.1 Request Data
>owin.RequestId	An optional string that uniquely identifies a request. The value is opaque and SHOULD have some level of uniqueness. A Host MAY specify this value. If it is not specified, middleware MAY set it. Once set, it SHOULD NOT be modified.

Eilon: Yep I read the OWIN spec :smile: Honestly I'm OK with forcing GUID because there's nothing *really* wrong with it in practice. It's just kind of annoying when practicality trumps so-called "correctness."

src/Microsoft.AspNet.Owin/OwinFeatureCollection.cs
```diff
@@ -427,6 +428,20 @@ public bool IsReadOnly
...
         }
 
+        Guid IRequestIdentifierFeature.TraceIdentifier
+        {
+            get
+            {
+                var requestId = Prop<string>(OwinConstants.RequestId);
+                if (requestId != null)
+                {
+                    return new Guid(requestId);
```
Tratcher: The OWIN spec say it's an opaque value, we can't assume it will parse as a GUID. TryParse?

src/Microsoft.AspNet.Owin/OwinEnvironment.cs
```diff
@@ -113,6 +113,12 @@ public OwinEnvironment(HttpContext context)
             }
 
             _context.Items[typeof(HttpContext).FullName] = _context; // Store for lookup when we transition back out of OWIN
+
+            if (context.GetFeature<IRequestIdentifierFeature>() != null)
```
Tratcher: I think FeatureMap handles this null check internally.

src/Microsoft.AspNet.Owin/OwinFeatureCollection.cs
Praburaj: sure. I can fix that. Else return `Guid.Empty`?

src/Microsoft.AspNet.Owin/OwinEnvironment.cs
Praburaj: Yeah. You're right. Removed the null check.

Praburaj:
@Tratcher Updated PR with feedback.

src/Microsoft.AspNet.Owin/OwinFeatureCollection.cs
Eilon: Hmm this makes me more worried actually. If we want to support OWIN, and OWIN uses strings, I don't see how we can force GUIDs... that's not to say that if we happen to get a GUID that we shouldn't use it...

Praburaj: OWIN on the top of any of IIS/Systemweb/WebListener will not be a problem. But this can affect any other servers. The alternate that I can think of is change all 'Guid` to `string` and continue to produce `GUID` in our servers so that E2E tracing is possible (and just do a `guid.ToString()`)?

Eilon:
:shipit: (we discussed more in person)

PR #215, opened 2015-03-05
GetXyzEncoder() shouldn't throw if IServiceProvider is null

GrabYourPitchforks:
Also remove dependency on full DI; use Interfaces instead

/cc @Praburaj


Praburaj:
Looks good to me :shipit:

Issue #152, opened 2014-11-14
SignOutContext needs AuthenticationProperties


Tratcher:
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.PipelineCore/Security/SignOutContext.cs
Compare:
http://katanaproject.codeplex.com/SourceControl/latest#src/Microsoft.Owin/Security/AuthenticationResponseRevoke.cs

This was a late addition to Katana. WsFed & OIDC have need for sign-out properties in order to customize single-signout redirects.

Also update HttpResponse.SignOut(...).

Praburaj:
https://github.com/aspnet/HttpAbstractions/commit/58c45cd379774cf1fd683a12465cb52094b7193d

PR #218, opened 2015-03-07
SignOutContext needs AuthenticationProperties

Praburaj:
Fixes https://github.com/aspnet/HttpAbstractions/issues/152

@Tratcher @HaoK


Tratcher:
@brentschmaltz This is for OIDC & WsFed.

src/Microsoft.AspNet.Http.Core/DefaultHttpResponse.cs
```diff
@@ -162,11 +160,11 @@ public override void SignIn(string authenticationScheme, [NotNull] ClaimsPrincip
             }
         }
 
-        public override void SignOut(string authenticationScheme)
+        public override void SignOut(string authenticationScheme, AuthenticationProperties properties)
         {
             var handler = HttpAuthenticationFeature.Handler;
 
-            var signOutContext = new SignOutContext(authenticationScheme);
+            var signOutContext = new SignOutContext(authenticationScheme, properties == null ? null : properties.Dictionary);
```
PinpointTownes: `properties?.Dictionary`?

Praburaj: sure. fixed.

davidfowl:
We don't do optional parameters anymore.

Praburaj:
@davidfowl Removed optional parameters on this method.

Praburaj:
Updated PR. 

/CC : @muratg

muratg:
Changes look good to me, I'd ping @brentschmaltz to see if he sees any problems, otherwise :shipit:

Praburaj:
@brentschmaltz?

brentschmaltz:
Adding some more context after speaking with Praburaj. there can be a lot of other parameters for signout: id_token, session_state, sid, idp, state etc.  I just saw the spec on Thursday afternoon and am digesting it. We may want to have first class properties similar to how OIDCMessage and WsFedMessage work. It may mean <templates> or derived types. Not sure. Signout it not as simple as we think.

Praburaj:
Merged https://github.com/aspnet/HttpAbstractions/commit/58c45cd379774cf1fd683a12465cb52094b7193d

Issue #208, opened 2015-02-28
Rename methods on CodePointFilter


GrabYourPitchforks:
`AllowBlock` is weirdly named, even if the doc comments are clear as to what it does. Consider renaming for clarity.

Potential candidates:
`AllowUnicodeBlock` (verbose, still slightly weird)
`AllowRange` (would require renaming `UnicodeBlock` to `UnicodeRange`, which is incorrect)
`Allow` (will have many overloads, but this should be fine)

PR #220, opened 2015-03-07
API cleanup: Rename UnicodeBlock -> UnicodeRange

GrabYourPitchforks:
The term "block" was confusing since I had intended for it to refer to a defined block from the Unicode spec, but it caused confusion in the filter APIs since as a verb it is a synonym for "forbid". This PR renames "block" to "range" throughout to avoid confusion. Also took this opportunity to clean up a bunch of doc comments.

Fixes https://github.com/aspnet/HttpAbstractions/issues/208.


GrabYourPitchforks:
Fixes https://github.com/aspnet/HttpAbstractions/issues/222, https://github.com/aspnet/HttpAbstractions/issues/208.

PR #225, opened 2015-03-09
Using new encoders over the old encoders.

Praburaj:
Fixes https://github.com/aspnet/HttpAbstractions/issues/198

@GrabYourPitchforks

### DIFF
--- a/src/Microsoft.AspNet.Http.Core/project.json
+++ b/src/Microsoft.AspNet.Http.Core/project.json
@@ -1,35 +1,34 @@
-
-{
-  "version": "1.0.0-*",
-  "description": "ASP.NET 5 HTTP feature implementations.",
-  "dependencies": {
-    "Microsoft.AspNet.FeatureModel": "1.0.0-*",
-    "Microsoft.AspNet.Http": "1.0.0-*",
-    "Microsoft.AspNet.Http.Interfaces": "1.0.0-*",
-    "Microsoft.AspNet.WebUtilities": "1.0.0-*",
-    "Microsoft.Net.Http.Headers": "1.0.0-*"
-
-  },
-  "frameworks": {
-    "dnx451": {},
-    "dnxcore50": {
-      "dependencies": {
-        "Microsoft.Net.WebSocketAbstractions": "1.0.0-*",
-        "System.Collections": "4.0.10-beta-*",
-        "System.ComponentModel": "4.0.0-beta-*",
-        "System.Diagnostics.Debug": "4.0.10-beta-*",
-        "System.Diagnostics.Tools": "4.0.0-beta-*",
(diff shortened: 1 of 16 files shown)


src/Microsoft.AspNet.Http.Core/project.json
```diff
@@ -1,35 +1,34 @@
...
+        "Microsoft.Framework.WebEncoders": "1.0.0-*",
+        "Microsoft.Net.Http.Headers": "1.0.0-*"
+    },
+    "frameworks": {
+        "dnx451": {},
+        "dnxcore50": {
+            "dependencies": {
+                "Microsoft.Net.WebSocketAbstractions": "1.0.0-*",
+                "System.Collections": "4.0.10-beta-*",
+                "System.ComponentModel": "4.0.0-beta-*",
```
Praburaj: I'll fix the formatting issues.

GrabYourPitchforks:
Looking at this, this makes me think that we need a static method that will allow changing HtmlEncoder.Default (and related). Otherwise I'm thinking this PR really doesn't have much benefit since developers still can't control the built-in implementation. :/

davidfowl:
So looking at the dependencies that Encoders have I wonder if we should split it out. It's not clear what things need an replacable implementation of the I*Encoder and which ones are using it as an implementation detail. 

We need to be careful where we introduce dependencies in the lower layers of the stack. I don't feel comfortable with our default Http implementations relying on DI.

GrabYourPitchforks:
In ASP.NET, all calls to UrlEncode (including calls made internally by the framework) would go through a centralized replaceable encoder. This was added due to very large customer demand. The general idea is to have parity in the new web stack.

We could absolutely split the package such that the base package has no dependencies. Then layer something else on top that understands DI and Options.

davidfowl:
Sure, I'm not pushing back on replaceable encoding. I'm more worried about how things are layered. It's a balance between ease of use in DI vs number of dependencies to get that going.

Praburaj:
>Looking at this, this makes me think that we need a static method that will allow changing HtmlEncoder.Default (and related).

@GrabYourPitchforks - Shall I file a work item to introduce a `set` on `HtmlEncoder.Default` property?

GrabYourPitchforks:
Sure, file away. Thanks! :)

GrabYourPitchforks:
@Praburaj Try adding a package reference to WebEncoders.Core instead of WebEncoders. That should solve the layering issues that concerned David.

Praburaj:
@GrabYourPitchforks sure I'll update the PR.

Praburaj:
@GrabYourPitchforks rebased updated and squashed.

Praburaj:
@GrabYourPitchforks signed off the change while near my office. I'm pushing this assuming no more concerns.

PR #230, opened 2015-03-13
AuthN cleanup

HaoK:
- Rename Security folder -> Authentication
- Change Authenticate to only take one scheme to match other APIs, the
params overload did not make it any cleaner to consume (since it didn't
produce a combined ClaimsPrincipal anyways)

cc @lodejard @Praburaj @Tratcher


davidfowl:
Looks cleaner

HaoK:
Merged 08ddbe853130fe10bdde6f8b529e5a885e0974be

dougbu:
@HaoK this change appears to be incomplete.  the `AuthorizationPolicy` class still exposes a collection of schemes.  it doesn't look possible to fix the current build break in MVC where it's calling `HttpContext.AuthenticateAsync(Policy.ActiveAuthenticationSchemes)` without fixing that gap.  what do you recommend as a quick fix for MVC?

HaoK:
I’ll fix it

Issue #231, opened 2015-03-13
New Url encoders difference in behavior from Url.EscapeDataString


Praburaj:
@GrabYourPitchforks 

```
Uri.EscapeDataString("https://www.googleapis.com/auth/plus.login") = "https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fplus.login"
```

With new encoder:
```
https:%2F%2Fwww.googleapis.com%2Fauth%2Fplus.login
```

GrabYourPitchforks:
Per RFC 3986 (Appendix A) / RFC 3987 (Sec. 2.2), the segment, query, and fragment portions of a URI are all allowed to contain the colon character. I originally suspected that the Uri.EscapeDataString behavior was for compliance with RFC 2396, but even that older RFC allows colons pretty much everywhere (including in the userinfo). RFC 3986 was extended to allow colons in the host name to support IPv6 addresses. Will follow up with that team to find out the history behind this behavior.

Praburaj:
https://github.com/aspnet/HttpAbstractions/commit/1e9d57f80ca883881804292448fff4de8b112733

PR #235, opened 2015-03-17
UrlEncoder should always encode the U+003A COLON character

GrabYourPitchforks:
Provides extra defense-in-depth in case an application is using this API to encode a relative URL, otherwise the part before the colon could inadvertently be treated as a scheme.

Addresses https://github.com/aspnet/HttpAbstractions/issues/231.

/cc @HaoK @Praburaj


HaoK:
:shipit:

Issue #246, opened 2015-04-04
Test failure on mono: Microsoft.AspNet.Http.Core.FormFeatureTests.ReadFormAsync_MultipartWithFieldAndFile_ReturnsParsedFormCollection [FAIL]


davidfowl:
https://travis-ci.org/aspnet/HttpAbstractions/builds/57113537

Looks like a functional issue

/cc @Tratcher @Praburaj

Tratcher:
https://github.com/aspnet/HttpAbstractions/blob/dev/test/Microsoft.AspNet.Http.Core.Tests/FormFeatureTests.cs#L100

I assume Git is setup to download files using OS specific line endings? I think that's changing the new lines in these multi-line C# strings. Multipart requires CRLF, Windows uses CRLF, but Linux uses LF (or CR?).

Recommended fix: Don't use multi-line C# strings when the new line characters actually matter.

PR #247, opened 2015-04-06
Fix multipart test on linux.

Tratcher:
#246


davidfowl:
The build got alot further this time so :shipit: 

There are still a bunch more issues though. 

https://travis-ci.org/aspnet/HttpAbstractions/builds/57361737#L3669

Tratcher:
It's the same issue in a different set of tests, so ill just fix it now.
________________________________

davidfowl:
@Tratcher great!

PR #248, opened 2015-04-06
Short circuit empty inputs.

Tratcher:
Convert.FromBase64CharArray on Mono does not allow count=0.

This optimization is already in place for the other code path.

### DIFF
--- a/test/Microsoft.Net.Http.Headers.Tests/RangeConditionHeaderValueTest.cs
+++ b/test/Microsoft.Net.Http.Headers.Tests/RangeConditionHeaderValueTest.cs
@@ -112,7 +112,7 @@ namespace Microsoft.Net.Http.Headers
         [InlineData("Sun, 06 Nov 1994 08:49:37 GMT \"x\"")]
         [InlineData(null)]
         [InlineData("")]
-        [InlineData(" Wed 09 Nov 1994 08:49:37 GMT")]
+        // [InlineData(" Wed 09 Nov 1994 08:49:37 GMT")] // Succeeds on Mono.
         [InlineData("\"x")]
         [InlineData("Wed, 09 Nov")]
         [InlineData("W/Wed 09 Nov 1994 08:49:37 GMT")]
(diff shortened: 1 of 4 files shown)


Tratcher:
@davidfowl The tests all pass now. The build is still broken for some reason around json parsing though.

davidfowl:
One more change required:

https://github.com/aspnet/HttpAbstractions/blob/dev/test/Microsoft.Framework.WebEncoders.Tests/project.json#L13

```JSON
{
    "compilationOptions": {
        "allowUnsafe": "true"
    },
}
```

true needs to be in quotes

Tratcher:
@davidfowl All clean now.

test/Microsoft.Net.Http.Headers.Tests/RangeConditionHeaderValueTest.cs
```diff
@@ -112,7 +112,7 @@ public void Parse_SetOfValidValueStrings_ParsedCorrectly()
         [InlineData("Sun, 06 Nov 1994 08:49:37 GMT \"x\"")]
         [InlineData(null)]
         [InlineData("")]
-        [InlineData(" Wed 09 Nov 1994 08:49:37 GMT")]
+        // [InlineData(" Wed 09 Nov 1994 08:49:37 GMT")] // Succeeds on Mono.
```
Eilon: Delete this line?

test/Microsoft.Net.Http.Headers.Tests/RangeConditionHeaderValueTest.cs
```diff
@@ -141,7 +141,7 @@ public void TryParse_SetOfValidValueStrings_ParsedCorrectly()
...
```
Eilon: Delete?

test/Microsoft.Net.Http.Headers.Tests/RangeConditionHeaderValueTest.cs
Tratcher: It really was supposed to fail. I left it in for follow-up investigation. I've filed https://github.com/aspnet/HttpAbstractions/issues/250 to track this.

Eilon: Ah ok!

PR #252, opened 2015-04-08
Replace WebSocket dependency. Remove chained dependencies.

Tratcher:
Microsoft.Net.WebSocketAbstractions can now be replaced with System.Net.WebSockets.

@davidfowl @Eilon


Eilon:
:shipit:

PR #254, opened 2015-04-08
Simplify Challenge flow

HaoK:
Sugar overloads still allow challenging multiple schemes, but each
ChallengeContext is only dealing with one scheme at a time

cc @Tratcher @lodejard

### DIFF
--- a/src/Microsoft.AspNet.Http.Core/Authentication/ChallengeContext.cs
+++ b/src/Microsoft.AspNet.Http.Core/Authentication/ChallengeContext.cs
@@ -4,33 +4,34 @@
 using System;
 using System.Collections.Generic;
 using Microsoft.AspNet.Http.Authentication;
-using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Http.Core.Authentication
 {
     public class ChallengeContext : IChallengeContext
     {
-        private List<string> _accepted;
+        private bool _accepted;
 
-        public ChallengeContext([NotNull] IEnumerable<string> authenticationSchemes, IDictionary<string, string> properties)
+        public ChallengeContext(string authenticationScheme, IDictionary<string, string> properties)
         {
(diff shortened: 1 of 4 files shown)


Tratcher:
It will be a little less efficient re-traversing the chain of handlers for every challenge, but how often do you plan on challenging more than one?

Lots of this sugar seems like it could be moved off to extension methods and stop cluttering up the main type.

HaoK:
I'm all for killing the multiple authentication scheme overloads entirely if you think they aren't needed.  I don't think we challenge with multiple schemes today anyways.  

But MVC ChallengeResult allows for this, but we could move the for loop to ChallengeResult instead of polluting HttpAbstractions itself.  

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ActionResults/ChallengeResult.cs

Tratcher:
Do you see any examples of ChallengeResult getting used with multiple challenges?

In classic HTTP auth it would make sense to challenge for NTLM, Basic, Digest, etc., but we don't directly implement any of those. We would only be challenging for OIDC, or WsFed, or MSA, etc. one at a time.

HaoK:
No I don't believe we do today as challenge result was pushed into MVC from the auth templates which currently only do one at a time challenges.  I think we can just kill multiple challenges at once all together then?

Tratcher:
For now, yes. They can always demand it back later.

HaoK:
Great, updated with those overloads removed, will send out matching security/MVC prs shortly

HaoK:
Meh, looks like I introduced a usage for ChallengeResult's multiple challenge support as part of the new AuthorizeFilter which challenges for all of the schemes in the AuthorizationPolicy.  But I think its ok if we move this support into ChallengeResult for now

HaoK:
c69c289abf85a42c5eff156f1a571efd26ce407f

Issue #167, opened 2015-01-15
Integrate cookie APIs with new strongly typed header values


Tratcher:
The RequestCookiesFeature internal parser can be replaced with the CookieHeaderValue parser.

Similarly the ResponseCookies class could be greatly simplified by delegating to the SetCookieHeaderValue class as a serializer.

PR #258, opened 2015-04-10
Update cookie APIs to use shared core

Tratcher:
#167 
We added the CookieHeaderValue and SetCookieHeaderValue types a while ago, but had not yet taken advantage of them in our cookie APIs. This now consumes them internally without affecting any public APIs.

### DIFF
--- a/src/Microsoft.AspNet.Http.Core/Collections/RequestCookiesCollection.cs
+++ b/src/Microsoft.AspNet.Http.Core/Collections/RequestCookiesCollection.cs
@@ -4,8 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using Microsoft.AspNet.Http;
-using Microsoft.AspNet.Http.Core.Infrastructure;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Http.Core.Collections
 {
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNet.Http.Core/RequestCookiesFeature.cs
```diff
@@ -44,18 +44,23 @@ public IReadableStringCollection Cookies
                 }
 
                 var headers = _request.Fetch(_features).Headers;
-                string cookiesHeader = ParsingHelpers.GetHeader(headers, Constants.Headers.Cookie) ?? string.Empty;
+                string[] values;
+                if (!headers.TryGetValue(HeaderNames.Cookie, out values))
+                {
+                    values = new string[0];
+                }
+                var cookiesHeader = string.Join(", ", values);
```
davidfowl: Why do we need this?

Tratcher: We use it to check for changes next time you access the collection. Doing Equals on a string[] is harder, and you never know if someone is going to change your own array references.  If it shows up on perf we could reduce it to a fancy hash code check.

davidfowl: It's not harder:

```
Enumerable.Equals(a,b)
```

davidfowl: And yes, we should profile

Tratcher:
Updated.

davidfowl:
:shipit:

src/Microsoft.AspNet.Http.Core/Collections/RequestCookiesCollection.cs
```diff
@@ -73,12 +72,20 @@ public IList<string> GetValues(string key)
...
         {
             _dictionary.Clear();
-            ParsingHelpers.ParseDelimited(cookiesHeader, SemicolonAndComma, AddCookieCallback, _dictionary);
+
+            IList<CookieHeaderValue> cookies;
+            if (CookieHeaderValue.TryParseList(values, out cookies))
+            {
+                foreach (var cookie in cookies)
+                {
+                    var name = Uri.UnescapeDataString(cookie.Name.Replace('+', ' '));
```
davidfowl: How many places do we need to do this? Can we use @GrabYourPitchforks  de/encoder?

Tratcher: He only left us with an encoder, not a decoder.

davidfowl: Damn

GrabYourPitchforks: I ended up not checking in the decoder because it wasn't significantly different from UnescapeDataString.

davidfowl: UnescapeDataString has weird limitations that can only be overridden in a config file.

Issue #149, opened 2014-11-08
Supported way of getting RemoteIpAddress


avanderhoorn:
In OWIN, `RemoteIpAddress` via the `HttpRequest`. In vNext world, I can see that `IHttpConnectionFeature` has the IpAddress, and I can see that `DefaultHttpRequest` has the `HttpConnectionFeature` has a private property, but that property is never used and `IHttpRequest` doesn't have a `RemoteIpAddress` property.

sandorfr:
It is frequent to use client ip adresss to approximate geolocation (analytics or choose an appropriate default culture). 

An example : server side call to mixpanel analytics, if we do not provide client IP, mixpanel takes our server IP to compute some attributes.

I definitely need this :). Thanks for the workaround  @tugberkugurlu .

avanderhoorn:
Just wondering if there is any progress on this one?

davidfowl:
We should add this to HttpContext /cc @lodejard @glennc @muratg

PR #263, opened 2015-04-14
Expose connection information as 1st class.

Tratcher:
#149 
IPAddresses, ports, and client certificates.
@avanderhoorn 
@davidfowl 
@muratg

### DIFF
--- a/src/Microsoft.AspNet.Http/ConnectionInfo.cs
+++ b/src/Microsoft.AspNet.Http/ConnectionInfo.cs
@@ -1,0 +1,27 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Net;
+using System.Security.Cryptography.X509Certificates;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Http
+{
+    public abstract class ConnectionInfo
+    {
+        public abstract IPAddress RemoteIpAddress { get; set; }
+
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNet.Http/ConnectionInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Net;
+using System.Security.Cryptography.X509Certificates;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Http
+{
+    public abstract class ConnectionInfo
```
muratg: Not sure if this needs to be an abstract class, could be an interface. @Tratcher made it so that it matches other properties hanging off of HttpContext.  @davidfowl your thoughts?

Tratcher: One reason that I remember we choose this was so that we could more easily add new (virtual) items without them being official breaking changes.

davidfowl:
Does this need the ITlsBindingXX interface?

Tratcher:
I don't see any immediate need to expose it, it's a very advanced scenario.

src/Microsoft.AspNet.Http/ConnectionInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        public abstract HttpContext HttpContext { get; }
```
davidfowl: I don't think this makes sense here.

davidfowl:
:shipit:

avanderhoorn:
Looks good!

Issue #270, opened 2015-04-14
ApiReview: AuthenticationDescription, AuthenticationProperties


Tratcher:
Rename the underlying collection referenced by these wrappers to be `items`, instead of `properties` or `dictionary`.

Issue #266, opened 2015-04-14
ApiReview: Move all authentication APIs to HttpContext.Authentication


Tratcher:
Authenticate, SignIn, SignOut, Challenge

But not User.

Issue #273, opened 2015-04-14
ApiReview: Favor abstract base classes over interfaces for context objects


Tratcher:
IWebSocketAcceptContext
IAuthenticateContext
IChallengeContext
IDescribeSchemesContext
ISignInContext
ISignOutContext

Tratcher:
This only works if all the POCOs are put into Http.Interfaces.

Issue #267, opened 2015-04-14
ApiReview: Move WebSocket APIs to HttpContext.WebSockets


Tratcher:
```
WebSocketRequestedProtocols { get; }
AcceptWebSocketAsync()
AcceptWebSocketAsync(string subProtocol)
```

But not `IsWebSocketRequest { get; }`

Tratcher:
I think it makes sense to also move IsWebSocketRequest in with everything else. Just accessing HttContext.WebSockets doesn't imply any operations. Compare to HttpContext.Authentication.

davidfowl:
context.WebSockets.IsWebSocketRequest? /cc @lodejard

Tratcher:
Also do with #273 for IWebSocketAcceptContext.

PR #282, opened 2015-04-17
Consolidate auth APIs

Tratcher:
#266 - Consolidate all the auth APIs to HttpContext.Authentication
#270 - Naming
@davidfowl @muratg 
/cc @HaoK


Tratcher:
Update to include #273 - POCO context objects.

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
```diff
@@ -0,0 +1,147 @@
...
+                new AuthenticationProperties(authenticateContext.Properties),
+                new AuthenticationDescription(authenticateContext.Description));
+        }
+
+        public override void Challenge(AuthenticationProperties properties, string authenticationScheme)
+        {
+            HttpResponseFeature.StatusCode = 401;
+            var handler = HttpAuthenticationFeature.Handler;
+
+            var challengeContext = new ChallengeContext(authenticationScheme, properties == null ? null : properties.Items);
```
HaoK: properties?.Items

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            // Verify all types ack'd
+            if (!signInContext.Accepted)
+            {
+                throw new InvalidOperationException("The following authentication scheme was not accepted: " + authenticationScheme);
+            }
+        }
+
+        public override void SignOut(string authenticationScheme, AuthenticationProperties properties)
```
HaoK: Should we introduce a new POCO that contains both the scheme and AuthenticationProperties, they always seem to be used together...  Or maybe just push the Scheme inside of the properties as a special property that's required?

```
public class AuthenticationProperties {
   AuthenticationProperties(string scheme)

   string Scheme { get; }
   Dictionary<string, string> Items { get; }
}

Comment (Tratcher) [inline src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs]: I've seen enough examples of scheme being used without properties that I think it's more convenient to keep them separate.

File: src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs [hunk 3]
```diff
@@ -0,0 +1,147 @@
...
+            var authenticateContext = new AuthenticateContext(authenticationScheme);
+            if (handler != null)
+            {
+                await handler.AuthenticateAsync(authenticateContext);
+            }
+
+            // Verify all types ack'd
+            if (!authenticateContext.Accepted)
+            {
+                throw new InvalidOperationException("The following authentication scheme was not accepted: " + authenticationScheme);
```
muratg: Nit: $"The following.... {authenticationScheme}.."

Ditto other exception messages below.

Tratcher:
Updated and rebased.

Tratcher:
Updated with related changes for WebSockets. #267

src/Microsoft.AspNet.Http/DefaultWebSocketManager.cs
```diff
@@ -0,0 +1,65 @@
...
+        private IHttpWebSocketFeature WebSocketFeature
+        {
+            get { return _webSockets.Fetch(_features); }
+        }
+
+        public override bool IsWebSocketRequest
+        {
+            get
+            {
+                var webSocketFeature = WebSocketFeature;
```
muratg: Why not use the property directly? Do we need to assign it to a variable?

src/Microsoft.AspNet.Http/DefaultWebSocketManager.cs
```diff
@@ -0,0 +1,65 @@
...
+            get
+            {
+                return ParsingHelpers.GetHeaderUnmodified(HttpRequestFeature.Headers,
+                    HeaderNames.WebSocketSubProtocols) ?? EmptyList;
+            }
+        }
+
+        public override Task<WebSocket> AcceptWebSocketAsync(string subProtocol)
+        {
+            var webSocketFeature = WebSocketFeature;
```
muratg: Looks like this is not used.

muratg:
Code changes look good to me and reflects the discussions from the API review.  

@davidfowl does new API surface look good to you?

src/Microsoft.AspNet.Http.Core/Authentication/AuthenticationManager.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            Challenge(properties, "");
+        }
+
+        public virtual void Challenge(string authenticationScheme)
+        {
+            Challenge(properties: null, authenticationScheme: authenticationScheme);
+        }
+
+        public abstract void Challenge(AuthenticationProperties properties, string authenticationScheme);
```
HaoK: I agree with Brent's issue, we should probably switch scheme to be the first parameter here to be consistent with sign in/signout

Issue #265, opened 2015-04-14
ApiReview: IApplicationBuilder & extensions


Tratcher:
Remove IApplicationBuilder Properties & Server setters.
Remove Run extensions that take services.
Remove Map string overload (and add implicit conversion between PathString & String).
Remove Map PredicateAsync support.

Tratcher:
@muratg @davidfowl @lodejard What about the Use extensions that take services? Those are still there.
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http.Abstractions/Extensions/UseExtensions.cs#L37

brockallen:
Were these removed just to reduce the API surface area and/or for simplicity, or was there some larger reason?

davidfowl:
Simplicity and usability.

PR #283, opened 2015-04-17
Builder and extensions cleanup

Tratcher:
#265
Hosting changes to follow.
@davidfowl @muratg

### DIFF
--- a/src/Microsoft.AspNet.Http.Core/Extensions/MapWhenExtensions.cs
+++ b/src/Microsoft.AspNet.Http.Core/Extensions/MapWhenExtensions.cs
@@ -5,12 +5,11 @@ using System;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Builder.Extensions;
+using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Builder
 {
-    using Microsoft.Framework.Internal;
     using Predicate = Func<HttpContext, bool>;
-    using PredicateAsync = Func<HttpContext, Task<bool>>;
 
     /// <summary>
     /// Extension methods for the MapWhenMiddleware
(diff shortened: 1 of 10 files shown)


Tratcher:
https://github.com/aspnet/Hosting/commit/7acee4d54550991e6380d6a94c7c37b8fb8ef0f1

src/Microsoft.AspNet.Http.Core/Extensions/MapWhenMiddleware.cs
```diff
@@ -20,27 +20,13 @@ public MapWhenMiddleware([NotNull] RequestDelegate next, [NotNull] MapWhenOption
 
         public async Task Invoke([NotNull] HttpContext context)
         {
-            if (_options.Predicate != null)
+            if (_options.Predicate(context))
```
Eilon: This doesn't need a null check anymore? The setter for this property doesn't seem to guard against null.

Tratcher: The NotNull guard is here:
https://github.com/aspnet/HttpAbstractions/pull/283/files#diff-93bfbc9b10d86e70f618ce3182ccfbeaL27

It used to check before because there were two different options, sync or async.

Eilon: That's a guard on the method, not on the actual options class. Someone can still go to the options and make it null, no?

davidfowl:
Where's the implicit conversion from string -> PathString

src/Microsoft.AspNet.Http.Core/Extensions/MapWhenMiddleware.cs
Tratcher: With effort. They would have to do `app.Use(next => new MapWhenMiddleware(next, options).Invoke);` since we don't provide an extension that takes the options directly.

Tratcher:
@davidfowl I didn't find any, just methods like this: https://github.com/aspnet/HttpAbstractions/pull/283/files#diff-d9042ff419b4052d360932870b783853L21

davidfowl:
Part of the removal of the string methods was about adding that conversion. Did that get lost as well? /cc @glennc @muratg

Tratcher:
Ah, I misread that feedback. I thought it was asking to remove any implicit type conversions between string and PathString.  I'll update the bug and the PR.

Tratcher:
Updated with implicit string converters.

src/Microsoft.AspNet.Http.Core/Extensions/MapWhenMiddleware.cs
Eilon: Hmm, true. Can we add a `[NotNull]` to the setter anyway? That way there would be no doubt.

Tratcher: Updated.

Eilon:
:shipit:

davidfowl:
:shipit:

src/Microsoft.AspNet.Http.Core/Extensions/MapWhenExtensions.cs
```diff
@@ -5,12 +5,11 @@
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Builder.Extensions;
+using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Builder
```
davidfowl: We totally need to change this namespace

PR #286, opened 2015-04-18
Support Set<T> & Get<T> in ISessionCollection

hishamco:
Provide a generic version of ```ISessionCollection.Set<T>``` and ```ISessionCollection.Get<T>``` which allow the developers to implement ```ISessionFormatter```. Based on David Fowler suggestion we can plug ```BondSessionFormatter``` or ```ProtoBufSessionFormatter``` as well.
/cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Http.Extensions/SessionExtensions.cs
+++ b/src/Microsoft.AspNet.Http.Extensions/SessionExtensions.cs
@@ -51,5 +51,20 @@ namespace Microsoft.AspNet.Http
             session.TryGetValue(key, out value);
             return value;
         }
+
+        public static void Set(this ISessionCollection session, string key, byte[] value)
+        {
+            session.Set(key, new ArraySegment<byte>(value));
+        }
+        
+        public static T Get<T>(this ISessionCollection session, string key, ISessionFormatter<T> formatter)
+        {
+            return formatter.Deserialize(session.Get(key));
+        }
+
+        public static void Set<T>(this ISessionCollection session, string key, T value, ISessionFormatter<T> formatter)
+        {
+            session.Set(key, formatter.Serialize(value));
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


msotclas:
Hi __@hishamco__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <p>The agreement was validated by Microsoft Open Technologies, Inc. and real humans are currently evaluating your PR.</p>

TTYL, MSOTBOT;

src/Microsoft.AspNet.Http.Extensions/SessionCollectionExtensions.cs
```diff
@@ -56,5 +56,15 @@ public static void Set(this ISessionCollection session, string key, byte[] value
...
+        {
+            return formatter.Deserialize(session.Get(key));
+        }
+
+        public static void Set<T>(this ISessionCollection session, string key, T value, ISessionFormatter<T> formatter)
+        {
+            session.Set(key, formatter.Serialize(value));
+        }
     }
 }
```
jasonwilliams200OK: When committing from Windows to GitHub: always a line-break at the end of each file, so the diff does not show this:
github: no newline before the end of line

Read more on [No newline before the end of file - Stackoverflow](http://stackoverflow.com/questions/5813311/no-newline-at-end-of-file).

hishamco: @jasonwilliams200OK I added two methods without adding or modifying line-break

davidfowl:
Currently we're not comfortable making this part of the built in interface (even though they are extension methods). The current recommendation is to put something like this into your application code and use whatever serializer/deserialize best fits your needs. We might revisit this in the future.

hishamco:
Thanks @davidfowl

Issue #274, opened 2015-04-14
ApiReview: Consolidate extensions


Tratcher:
HeaderDictionaryTypeExtensions - Move all of these extensions (except GetTypedHeaders) to public methods on Request/ResponseHeaders.
HttpResponseSendingExtensions - Cut all of these unless find examples where they are very useful (MVC?)

PR #290, opened 2015-04-22
Reorganize the strong header type extensions. Remove SendAsync extensions

Tratcher:
#274 
HeaderDictionaryTypeExtensions: "Remove everything except GetTypedHeaders. Everything else goes to  public method on Either ReuqestHeaders or ResponseHeaders depending on who uses them"
HttpResponseSendingExtensions: "Cut all of these. If there is significant or painful to work around uses then come back to us."


Tratcher:
@muratg @davidfowl

src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryTypeExtensions.cs
```diff
@@ -23,12 +23,14 @@ public static ResponseHeaders GetTypedHeaders(this HttpResponse response)
             return new ResponseHeaders(response.Headers);
         }
 
-        public static DateTimeOffset? GetDate([NotNull] this IHeaderDictionary headers, [NotNull] string name)
+        // These are all shared helpers used by both RequestHeaders and ResponseHeaders
+
```
muratg: nit: blank line

muratg:
RequestHeaders and ResponseHeaders seem to share some code. Would it make sense to create an abstract class that covers the shared functionality, and move the functionality in the HeaderDictionaryTypeExtensions (which are currently becoming Internal with this change) to this shared abstract class?

Tratcher:
I tried that in the initial feature work but it was rejected in code review.

muratg:
:shipit:

Issue #295, opened 2015-04-29
ApiReview: Rename packages (again)


Tratcher:
Http.Abstractions – HttpContext (abstract), referenced by middleware, Http, and applications
Http – DefaultHttpContext (implementation), referenced by Hosting (no change)
Http.Features – Feature interfaces & POCOs, referenced by Http and Servers

It might even make sense to merge the FeatureModel package into Http.Features.

Tratcher:
Moving the FeatureModel work to its own bug: https://github.com/aspnet/HttpAbstractions/issues/297

PR #296, opened 2015-04-29
Rename Core->Abstractions and Interfaces->Features.

Tratcher:
#295
@muratg @BrennanConroy 

I'm delaying the proposal to merge FeatureModel into Features until we can decide how it would impact the namespaces.


davidfowl:
> I'm delaying the proposal to merge FeatureModel into Features until we can decide how it would impact the namespaces.

Good! I was looking for that change at first. We also need to verify if everyone that consumes the features needs to have the feature model types (or is that just server authors).

BrennanConroy:
Looks good

Issue #269, opened 2015-04-14
ApiReview: QueryString & Builder


Tratcher:
Consolidate QueryBuilder into QueryString
Remove the `QueryString(string name, string value)` constructor. Consider changing it to a Create method with the Builder APIs?

Consider renaming QueryString (needs design)

Tratcher:
Usage 1:
```c#
var queryBuilder = new QueryBuilder() 
{ 
  { "grant_type", "authorization_code" }, 
  //...
};
var result = queryBuilder.ToString();
```
https://github.com/aspnet/Security/blob/87c31c5526edee557997f03c7b13d46cb5d1d893/src/Microsoft.AspNet.Authentication.Facebook/FacebookAuthenticationHandler.cs#L30-L39
https://github.com/aspnet/Security/blob/63fc18b945bb30b217de20d2aec7c042f70a68ae/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs#L228-L236

Usage 2:
```c#
var queryBuilder = new QueryBuilder(); 
foreach (var kvp in values) 
{
  //...
  queryBuilder.Add(kvp.Key, kvp.Value); 
} 
return queryBuilder.ToString(); 
```
https://github.com/aspnet/Routing/blob/3626900bc9bbb3186e2e47d27dc18661081d9919/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs#L259-L277

Tratcher:
QueryString cannot efficiently implement either pattern shown above. It would be like trying to merge String and StringBuilder.  I think QueryBuilder needs to remain as is.  I will just update the QueryString constructor.

PR #298, opened 2015-04-30
Clean up QueryString APIs, add tests.

Tratcher:
This makes QueryString easier to work with but does not replace the QueryBuilder scenarios.
#269 
@muratg @davidfowl @loudej


muratg:
Looks good and :+1: on tests!  Rebase to get the updated (c) header and :shipit:

Issue #275, opened 2015-04-14
ApiReview: Make UriHelper only static


Tratcher:
Remove everything except the static Create methods that return string.

PR #300, opened 2015-04-30
Reduce UriHelper to static methods.

Tratcher:
#275

### DIFF
--- a/src/Microsoft.AspNet.Http.Extensions/ResponseHeaders.cs
+++ b/src/Microsoft.AspNet.Http.Extensions/ResponseHeaders.cs
@@ -126,20 +126,20 @@ namespace Microsoft.AspNet.Http.Headers
             }
         }
 
-        public UriHelper Location
+        public Uri Location
         {
             get
             {
                 Uri uri;
                 if (Uri.TryCreate(Headers[HeaderNames.Location], UriKind.RelativeOrAbsolute, out uri))
                 {
-                    return new UriHelper(uri);
+                    return uri;
                 }
                 return null;
             }
             set
             {
-                Headers.Set(HeaderNames.Location, value);
+                Headers.Set(HeaderNames.Location, value == null ? null : UriHelper.Encode(value));
             }
         }
 
--- a/src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
+++ b/src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
@@ -8,107 +8,68 @@ namespace Microsoft.AspNet.Http.Extensions
     /// <summary>
     /// A helper class for constructing encoded Uris for use in headers and other Uris.
     /// </summary>
-    public class UriHelper
+    public static class UriHelper
     {
-        public UriHelper()
-        {
-        }
-
-        public UriHelper(HttpRequest request)
-        {
-            Scheme = request.Scheme;
-            Host = request.Host;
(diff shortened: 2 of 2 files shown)


Tratcher:
@muratg @davidfowl

src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
```diff
@@ -8,107 +8,46 @@ namespace Microsoft.AspNet.Http.Extensions
...
-            return Scheme + "://" + Host + path + Query + Fragment;
-        }
-
-        public override string ToString()
-        {
-            return IsFullUri ? GetFullUri() : GetPartialUri();
-        }
-
-        public static string Create(PathString pathBase,
+        public static string Encode(PathString pathBase,
```
davidfowl: Doc comment? What does this do exactly

Tratcher:
Updated.

src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
```diff
@@ -8,107 +8,68 @@ namespace Microsoft.AspNet.Http.Extensions
...
-
-        public override string ToString()
-        {
-            return IsFullUri ? GetFullUri() : GetPartialUri();
-        }
-
-        public static string Create(PathString pathBase,
+        /// <summary>
+        /// Combines the given uri components into a string that is properly encoded for use in HTTP headers.
+        /// </summary>
```
muratg: Doc nit: lowercase "uri" should probably be URI (or Uri)

src/Microsoft.AspNet.Http.Extensions/ResponseHeaders.cs
```diff
@@ -126,20 +126,20 @@ public EntityTagHeaderValue ETag
...
                 {
-                    return new UriHelper(uri);
+                    return uri;
                 }
                 return null;
             }
             set
             {
-                Headers.Set(HeaderNames.Location, value);
+                Headers.Set(HeaderNames.Location, value == null ? null : UriHelper.Encode(value));
```
muratg: Consider using null coalescing (?? operator)

src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
```diff
@@ -8,107 +8,68 @@ namespace Microsoft.AspNet.Http.Extensions
...
-                Query = query,
-                Fragment = fragment
-            }.GetPartialUri();
+            string combinePath = (pathBase.HasValue || path.HasValue) ? (pathBase + path).ToString() : "/";
+            return combinePath + query + fragment;
         }
 
-        public static string Create(string scheme,
+        /// <summary>
+        /// Combines the given uri components into a string that is properly encoded for use in HTTP headers.
```
muratg: Ditto

src/Microsoft.AspNet.Http.Extensions/ResponseHeaders.cs
Tratcher: That only works for the reverse case `value == null ? DoSomethingElse() : null;` We only want to call Encode if it is not null.

muratg: Ah, right.

muratg:
:shipit:

Issue #281, opened 2015-04-17
ApiReview: Move most of Http into Http.Internal namespace


Tratcher:
Move everything but HttpContext/Request/Response, and then pull back items as needed.

PR #302, opened 2015-05-06
Reorganize files, namespaces

Tratcher:
#265 Leftover cleanup to remove Use extensions that take services.
#281 Reorganize namespaces and files.
- Items used only internally or for tests are moved to .Internal namespace
- Features moved to the Features namespace. They are similar to internal.
- Update folder structure to mirror namespaces

@muratg @davidfowl


muratg:
:shipit:

Issue #272, opened 2015-04-14
ApiReview: Make most properties settable on HttpContext, HttpRequest, HttpResponse, etc.


Tratcher:
Anything that makes sense to set during testing.

Yes:
Items (& IItemsFeature.Items)
RequestAborted
Request.Cookies (& IRequestCookiesFeature.Cookies)
Request.IsHttps
Request.Query (& IQueryFeature.Query)

Maybe:
Request, Response
Request.Headers
Response.Headers
Response.HeadersSent
Response.Cookies - not a collection

No:
IsWebSocketRequest (You have to replace the entire feature for testing)
WebSocketRequestedProtocols
Request.HasFormContentType (set the Form property instead)

Tratcher:
Remove DefaultHttpContext.Revision.

Tratcher:
HttpContext.Session can't be easily set. The property is an ISessionCollection, but the feature stores an ISession that the collection wraps. We may need to think about the layering there.

PR #305, opened 2015-05-11
Make more properties settable 

Tratcher:
#272 (Items, RequestAborted, IsHttps, Query, Cookies).
That should cover everything but Session which cannot be easily set with it's current design.
@muratg @davidfowl

### DIFF
--- a/src/Microsoft.AspNet.Http.Abstractions/QueryString.cs
+++ b/src/Microsoft.AspNet.Http.Abstractions/QueryString.cs
@@ -111,6 +111,7 @@ namespace Microsoft.AspNet.Http
         /// </summary>
         /// <param name="name">The un-encoded parameter name</param>
         /// <param name="value">The un-encoded parameter value</param>
+        /// <returns>The resulting QueryString</returns>
         public static QueryString Create(string name, string value)
         {
             return new QueryString("?" + UrlEncoder.Default.UrlEncode(name) + '=' + UrlEncoder.Default.UrlEncode(value));
@@ -120,7 +121,7 @@ namespace Microsoft.AspNet.Http
         /// Creates a query string composed from the given name value pairs.
         /// </summary>
         /// <param name="parameters"></param>
-        /// <returns></returns>
+        /// <returns>The resulting QueryString</returns>
         public static QueryString Create(IEnumerable<KeyValuePair<string, string>> parameters)
         {
             var builder = new StringBuilder();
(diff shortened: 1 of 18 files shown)


src/Microsoft.AspNet.Http/Constants.cs
```diff
@@ -5,6 +5,7 @@ namespace Microsoft.AspNet.Http.Internal
 {
     internal static class Constants
     {
+        internal const string Http = "HTTP";
```
Eilon: Any good reason why these are all caps?

Tratcher: No, they should probably be lower case per spec.

Eilon: Yeah it looks wrong with caps... fix here I guess? Or new PR? Up to you.

Tratcher:
Updated.

src/Microsoft.AspNet.Http.Abstractions/QueryString.cs
```diff
@@ -137,6 +137,30 @@ public static QueryString Create(IEnumerable<KeyValuePair<string, string>> param
             return new QueryString(builder.ToString());
         }
 
+        /// <summary>
+        /// Creates a query string composed from the given name value pairs.
+        /// </summary>
+        /// <param name="parameters"></param>
+        /// <returns></returns>
```
muratg: Returns QueryString?

src/Microsoft.AspNet.Http/Features/QueryFeature.cs
```diff
@@ -41,13 +42,22 @@ public IReadableStringCollection Query
                 }
 
                 var queryString = _request.Fetch(_features).QueryString;
-                if (_query == null || _queryString != queryString)
+                if (!string.Equals(_queryString, queryString, StringComparison.Ordinal) || _query == null)
```
muratg: Consider putting null check before string comparison. If you make that change, in case of _query being null, there wouldn't be a string comparison at all.

src/Microsoft.AspNet.Http/Features/RequestCookiesFeature.cs
```diff
@@ -50,19 +50,37 @@ public IReadableStringCollection Cookies
...
 
-                if (_cookiesCollection == null)
+                if (_cookieHeaders == null || !Enumerable.SequenceEqual(_cookieHeaders, values, StringComparer.Ordinal))
                 {
                     _cookieHeaders = values;
-                    _cookiesCollection = new RequestCookiesCollection();
+                    if (_cookiesCollection == null)
+                    {
+                        _cookiesCollection = new RequestCookiesCollection();
+                        _cookies = _cookiesCollection;
```
muratg: Q: missing a cast?

muratg:
Nit: Test names have a little bit of inconsistency, but not a big deal.

After addressing other comments, :shipit:

src/Microsoft.AspNet.Http/Features/RequestCookiesFeature.cs
Tratcher: Nope, IRequestCookiesCollection implements IReadableStringCollection.

Issue #176, opened 2015-01-21
Add a copy constructor to MediaTypeHeaderValue


Tratcher:
Scenario:
MVC uses some lists of MTHVs to determine what formats are supported. They select one from the list and use it to configure the response. Sometimes this involves modifying the value (e.g. adding charset), however this affects the value in the list and may impact matching and output for subsequent requests.

To prevent modifications they need to make a copy of the MTHV. Right now they are using ToString and Parse to make a copy, which is least efficient way possible:
https://github.com/aspnet/Mvc/blob/eda4b16cc5dfc11528116c51e3a119c1b513e986/src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs#L181

Add a copy constructor to avoid the all the string generation, string parsing, and parameter validation. NameValueHeaderValue will also require a copy constructor.

PR #306, opened 2015-05-15
Add Clone() to MediaTypeHeaderValue and NameValueHeaderValue.

Tratcher:
#176 
MVC needs a more efficient way to copy this type to avoid accidental modifications. Right now they do it the worst possible way by doing ToString and then reparsing the value.

We can add this to other types as needed.

@muratg


Tratcher:
@kichalla

kichalla:
:shipit:

PR #316, opened 2015-05-28
Session API review changes

kichalla:
@Tratcher (This PR builds on top of the previous PR related to referencing Http.Features in Http.Abstractions)


src/Microsoft.AspNet.Http.Extensions/HttpContextExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Http.Features;
+
+namespace Microsoft.AspNet.Http
+{
+    public static class HttpContextExtensions
+    {
+        public static async Task<ISession> GetSessionAsync(this HttpContext httpContext)
```
Tratcher: Needs a comment on when/why you would use this.

Tratcher: Actually, do you need this now that LoadAsync is directly exposed?  `HttpContext.Session.LoadAsync();`

src/Microsoft.AspNet.Http.Features/ISession.cs
```diff
@@ -1,20 +1,20 @@
...
-using System;
 using System.Collections.Generic;
+using System.Threading.Tasks;
 
 namespace Microsoft.AspNet.Http.Features
 {
     public interface ISession
     {
-        void Load();
+        Task LoadAsync();
```
Tratcher: We should add a state field to indicate if the session has been loaded, and one to show if it has been modified since loaded. Right now you can't tell if you need to call LoadAsync or CommitAsync.

Tratcher:
:watch:

src/Microsoft.AspNet.Http.Features/ISession.cs
kichalla: But shouldn't that state be internal to the session implementations itself?...like from middleware we just commit and the implementation internally decides if it needs to commit or not...

src/Microsoft.AspNet.Http.Extensions/HttpContextExtensions.cs
kichalla: Good catch, we could remove this now.

src/Microsoft.AspNet.Http.Features/ISession.cs
Tratcher: Things are a little different now that it's async. How does the performant user know if they need to call it? Or do they just always call it at at the beginning of their code?

Issue #317, opened 2015-05-29
Remove dictionary that tracks string -> Type in FeatureCollection


davidfowl:
We don't support duck typing in the feature collection anymore so tracking interface types by full name is pointless and wastes memory.

PR #319, opened 2015-05-29
Removed bookeeping for duck typing feature interfaces

davidfowl:
- Removed interlocked increment since the version was
being incremented within the lock

#317 

/cc @Tratcher @lodejard


Tratcher:
Doesn't interlocked also flush cpu caches?

PR #323, opened 2015-06-03
Auth API changes - Async only and add challenge behavior

HaoK:
Changes needed for Authentication changes in security:
- IAuthenticationHandler is now mostly async, AuthenticationManager.Authencate is a sync wrapper
- Http.Abstractions now depends on Http.Features to reuse AuthenticateContext in AuthManager
- Nuke AuthenticateResult, with Authenticate api changed to use AuthenicateContext
- Add sugar methods which just take scheme
- Remove 401 setting behavior for challenge in manager (Handler will be responsible for everything)
- Add ChallengeBehavior enum/property to ChallengeContext
- Add new Forbidden API which maps to (Challenge with Forbidden(403) behavior)

cc @lodejard @davidfowl @Tratcher @divega


src/Microsoft.AspNet.Http.Abstractions/project.json
```diff
@@ -2,6 +2,7 @@
   "version": "1.0.0-*",
   "description": "ASP.NET 5 HTTP object model. HttpContext and family.",
   "dependencies": {
+    "Microsoft.AspNet.Http.Features": "1.0.0-*",
```
Tratcher: @kichalla Didn't you already add this?

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
```diff
@@ -11,9 +12,24 @@ public abstract class AuthenticationManager
     {
         public abstract IEnumerable<AuthenticationDescription> GetAuthenticationSchemes();
 
-        public abstract AuthenticationResult Authenticate(string authenticationScheme);
+        public abstract void Authenticate(AuthenticateContext context);
 
-        public abstract Task<AuthenticationResult> AuthenticateAsync(string authenticationScheme);
+        public abstract Task AuthenticateAsync(AuthenticateContext context);
```
PinpointTownes: Retrieving the `AuthenticationProperties` with this new method will be really cumbersome...

HaoK: That is fairly uncommon which is why we are leaning this way to start (Identity will be using the properties, so when I get there, if its super bad, don't worry I'll definitely add sugar :)

But most of the time, Authenticate() giving you back a claimsprincipal really all you want, without having to dig through the old Result tuple...

src/Microsoft.AspNet.Http.Abstractions/project.json
kichalla: @Tratcher : I did not check this in actually...i wanted to combine Session related changes with the checkin...

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
PinpointTownes: I'm not sure it's that uncommon: I've been asked at least 3 or 4 times on JabbR how to retrieve the expiration date of a cookie or a token.

HaoK: yeah but its not that hard...

```
var context = new AuthenticateContext(scheme);
httpContext.Authentication.Authenticate(context);
var properties = context.Properties;
```

PinpointTownes: I agree, but it's not really intuitive. And you now have to ensure `Accepted` is true:

```csharp
var context = new AuthenticateContext(scheme);
httpContext.Authentication.Authenticate(context);
if (context.Accepted) {
    var principal = context.Principal;
    ...
}
```

... which is not really simpler than the old API:

```csharp
var principal = (await context.Authentication.AuthenticateAsync(scheme))?.Principal;
if (principal != null) {
    ...
}
```

HaoK: Nah, you don't have to ensure accepted is true, the AuthenticationManager.Authenticate ensures that, or it throws.  That's an implementation detail, I can possibly make that internal

PinpointTownes: Ah yeah, you just have to ensure that the principal is not null, just like the old approach.

HaoK: Yep, the motivation for this change was to make it easier to access the principal, with the intentional tradeoff that accessing Properties/Description is harder.

darrelmiller:
Would it be possible for someone to explain the reasoning behind changing to this style of API?  Personally I find the new style a major regression from the point of view of clarity, readability and discovery.

Instead of a function where we pass the Scheme to Authenticate() and get a result we now must construct a context object that will be mutated by the Authenticate method.  

This approach forces me to ask a ton of questions about API:
- How should the context object be constructed? Is the ctor parameter enough? Do I also need to set properties on it? Does it need to be passed to another method first for other mutations?
- How do I determine failure?  Will the method throw if Authentication fails?  What property on the context reflects success?
- What properties will success update on the context object?  Which properties are inputs, which are outputs?  Is that context object thread safe?  What happens if I pass it elsewhere and some other code is also passing it to an another Authenticate method?

Functions/Methods have signatures for a reason.  They have input parameters and they have return values.  This allows to reason about the effects and non-effects of calling them.  This use of context objects completely eliminates these benefits.  Why are we doing it?

HaoK:
Just to be clear, there still is the following API (which now directly give syou the principal):

```
ClaimsPrincipal Authenticate(string authenticationScheme)
```

Its only if you are using the less commonly accessed AuthenticationProperties/Description where you have to use the context

darrelmiller:
@HaoK But that's syntactic sugar over the primary API which uses the context object, right?

HaoK:
Depends on how far you want to take that line of argument, the whole AuthenticationManager API is 'sugar' on top of the HttpFeature which has always looked this way (AuthenticateContext)...  The most commonly used API should be the new simpler ClaimsPrincipal Authenticate.  We removed the intermediate API which returned a tuple of 3 things from the Context.  The intent is to guide people to the new simplier API.

darrelmiller:
I agree that the simpler ClaimsPrincipal Authenticate is better than the API that returned AuthenticationResult.  And if the objective is to guide people towards the simpler API for the majority of cases then I would agree that is a good thing.
I'm just frustrated by context objects throughout the framework being the primary interaction method.  Glad to hear that's not the case here.

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
```diff
@@ -117,6 +95,18 @@ public override void Challenge(string authenticationScheme, AuthenticationProper
             }
         }
 
+        // You are not allowed access
+        public void Forbidden(string authenticationScheme, AuthenticationProperties properties)
```
davidfowl: Deny?

HaoK: Are we changing the ChallengeBehavior enum to match the naming as well?

davidfowl: Forbid?

PinpointTownes:
> Functions/Methods have signatures for a reason. They have input parameters and they have return values. This allows to reason about the effects and non-effects of calling them. This use of context objects completely eliminates these benefits. Why are we doing it?

I couldn't agree more with you, sir :+1:

HaoK:
Renamed Forbidden => Forbid and added overloads to abstract base class to make it friendlier (optional AuthenticationProperties, and exposed the raw abstract Challenge overload which takes ChallengeBehavior which will make @lodejard happy)

HaoK:
One thing I couldn't figure out how to do yet is remove the 401 setting inside of Challenge, @lodejard was adamant that this behavior belongs in the IAuthenticationHandler instead, I haven't quite gotten that to work yet.

HaoK:
Per discussion with @Tratcher made AuthManager apis async to try and eliminate duplicate code paths in AuthHandlers.

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
```diff
@@ -11,41 +14,93 @@ public abstract class AuthenticationManager
...
+        public abstract Task SignOutAsync(string authenticationScheme, AuthenticationProperties properties);
 
-        public abstract void SignOut(string authenticationScheme, AuthenticationProperties properties);
+        private static readonly TaskFactory _myTaskFactory = new TaskFactory(CancellationToken.None,
+            TaskCreationOptions.None, TaskContinuationOptions.None, TaskScheduler.Default);
+
+        private static TResult RunSync<TResult>(Func<Task<TResult>> func)
+        {
+            // REVIEW: Do we need to flow culture info still?
+            return _myTaskFactory.StartNew(() => func()).Unwrap().GetAwaiter().GetResult();
```
PinpointTownes: OMFG, what's that? :scream:

It's bad, bad, bad, horribly bad: you're offloading an **async** delegate (`() => func()`) to the thread pool before unwrapping it and then blocking on this new task :smile: 

This clearly creates unnecessary pressure on the thread pool, specially since you use `StartNew` on an `Task`-returning delegate, which is supposed to be non-blocking.

HaoK: Yeah well this was the only safe way to do this historically in asp.net (calling async methods synch). Other less nasty ways would deadlock in some situations. I'm not sure this is still needed in the new world.

Tratcher: I'd be really surprised if this was still necessary now that we no longer have the System.Web call context.

HaoK: Ok cool, so its definitely not as gross with out this...

@davidfowl What do you think about this change?  (Making everything async only for auth, and only having a sync public wrapper in the manager?

HaoK:
After discussing with @lodejard @Tratcher @Eilon concluded that we are going with async only, removing prototype tag, these should be the changes needed for security.

src/Microsoft.AspNet.Owin/OwinEnvironment.cs
```diff
@@ -61,7 +61,7 @@ public OwinEnvironment(HttpContext context)
                 { OwinConstants.ResponseReasonPhrase, new FeatureMap<IHttpResponseFeature>(feature => feature.ReasonPhrase, (feature, value) => feature.ReasonPhrase = Convert.ToString(value)) },
                 { OwinConstants.ResponseHeaders, new FeatureMap<IHttpResponseFeature>(feature => feature.Headers, (feature, value) => feature.Headers = (IDictionary<string, string[]>)value) },
                 { OwinConstants.ResponseBody, new FeatureMap<IHttpResponseFeature>(feature => feature.Body, () => Stream.Null, (feature, value) => feature.Body = (Stream)value) },
-                { OwinConstants.CommonKeys.OnSendingHeaders, new FeatureMap<IHttpResponseFeature>(feature => new Action<Action<object>, object>(feature.OnSendingHeaders)) },
+                { OwinConstants.CommonKeys.OnSendingHeaders, new FeatureMap<IHttpResponseFeature>(feature => new Action<Func<object, Task>, object>(feature.OnResponseStarting)) },
```
PinpointTownes: I'm not sure you can change the signature of `server.OnSendingHeaders` without breaking tons of OWIN middleware that expect a `Action<Action<object>, object>`.

http://owin.org/spec/spec/CommonKeys.html

src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
```diff
@@ -22,11 +23,11 @@ public abstract class HttpResponse
 
         public abstract IResponseCookies Cookies { get; }
 
-        public abstract bool HeadersSent { get; }
+        public abstract bool HasStarted { get; }
 
-        public abstract void OnSendingHeaders(Action<object> callback, object state);
+        public abstract void OnResponseStarting(Func<object, Task> callback, object state);
```
PinpointTownes: Yeah, does that mean OSH callbacks will be async only? :smile: 
Is it worth re-opening the ticket in aspnet/Security then?

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
```diff
@@ -11,41 +13,66 @@ public abstract class AuthenticationManager
...
-        public virtual void Challenge(string authenticationScheme)
+        public virtual Task ChallengeAsync([NotNull] string authenticationScheme)
         {
-            Challenge(authenticationScheme: authenticationScheme, properties: null);
+            return ChallengeAsync(authenticationScheme: authenticationScheme, properties: null);
         }
 
-        public abstract void Challenge(string authenticationScheme, AuthenticationProperties properties);
+        // Leave it up to authentication handler to do the right thing for the challenge
+        public Task ChallengeAsync([NotNull] string authenticationScheme, AuthenticationProperties properties)
```
Tratcher: You're using `[NotNull]` here but then passing in null from the overloads. This can only end well...

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
```diff
@@ -11,41 +13,66 @@ public abstract class AuthenticationManager
...
+        {
+            return ChallengeAsync(authenticationScheme, properties, ChallengeBehavior.Automatic);
+        }
 
-        public void SignIn(string authenticationScheme, ClaimsPrincipal principal)
+        public Task SignInAsync([NotNull] string authenticationScheme, ClaimsPrincipal principal)
```
Tratcher: nit, organization: Move this down to be with the rest of SignIn

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
```diff
@@ -11,41 +13,66 @@ public abstract class AuthenticationManager
...
+            return SignOutAsync(authenticationScheme: null, properties: null);
+        }
+
+        public Task SignOutAsync([NotNull] string authenticationScheme)
+        {
+            return SignOutAsync(authenticationScheme, properties: null);
+        }
 
-        public abstract void SignOut(string authenticationScheme, AuthenticationProperties properties);
+        public abstract Task SignOutAsync([NotNull] string authenticationScheme, AuthenticationProperties properties);
```
Tratcher: Out of curiosity, what are you expecting [NotNull] to do on an abstract method? I found EF doing the same thing recently.

src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
```diff
@@ -24,9 +25,9 @@ public abstract class HttpResponse
 
         public abstract bool HasStarted { get; }
 
-        public abstract void OnResponseStarting(Action<object> callback, object state);
+        public abstract void OnResponseStarting(Func<object, Task> callback, object state);
```
Tratcher: Do we want sync sugar for these? And while we're at it we could have sugar that doesn't require state.

davidfowl: Documentation

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
HaoK: Whoops I had a bad feeling when I added them yesterday while rebasing to dev. Now I know why!  Good catch :)

HaoK: Updated PR to use [NotNull] everywhere, Challenge sugar methods now pass string.Empty since we were already doing string.IsNullOrWhitespace checks, this shouldn't affect anything.  Also nuked the overloads of SignIn/SignOut that didn't require a scheme, this was on the todo list from earlier API reviews with Lou

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
HaoK: I think its so tools like resharper can warn you

src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
HaoK: I filed https://github.com/aspnet/HttpAbstractions/issues/333 to track.

Lou said no sync sugar anywhere as part of agreeing to full async.  I guess we are strongly encouraging async this way. I'll add a stateless sugar overload now

src/Microsoft.AspNet.Owin/OwinEnvironment.cs
```diff
@@ -61,7 +61,7 @@ public OwinEnvironment(HttpContext context)
                 { OwinConstants.ResponseReasonPhrase, new FeatureMap<IHttpResponseFeature>(feature => feature.ReasonPhrase, (feature, value) => feature.ReasonPhrase = Convert.ToString(value)) },
                 { OwinConstants.ResponseHeaders, new FeatureMap<IHttpResponseFeature>(feature => feature.Headers, (feature, value) => feature.Headers = (IDictionary<string, string[]>)value) },
                 { OwinConstants.ResponseBody, new FeatureMap<IHttpResponseFeature>(feature => feature.Body, () => Stream.Null, (feature, value) => feature.Body = (Stream)value) },
-                { OwinConstants.CommonKeys.OnSendingHeaders, new FeatureMap<IHttpResponseFeature>(feature => new Action<Action<object>, object>(feature.OnResponseStarting)) },
+                { OwinConstants.CommonKeys.OnSendingHeaders, new FeatureMap<IHttpResponseFeature>(feature => new Action<Func<object, Task>, object>(feature.OnResponseStarting)) },
```
PinpointTownes: So, what did you decide for OWIN?

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
PinpointTownes: IMHO, using `string.Empty` is a bit ugly. And removing `SignIn/SignOut` overloads that don't take a specific scheme while keeping the `Challenge` method that takes no parameter is a bit weird, no?

src/Microsoft.AspNet.Owin/OwinEnvironment.cs
HaoK: @Tratcher mentioned something about adapting this, I haven't had a chance to take a look at this yet.  I'm guessing we don't have any test coverage of this?

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
HaoK: Naked Challenge is required for naked Authorize which is pretty important for MVC.

SignIn/SignOut with no arguments where any automatic middleware picks that up felt like a pretty unnecessary feature.  

We are going to try and make it explicit so authenticationScheme is more understood since its always there...

PinpointTownes: Well, `SignIn/SignOut with no arguments` are a great way for the underlying application to keep the separation of concerns clear. For instance, controllers shouldn't have to know which schemes have been configured in Startup.cs request to globally sign in or sign out the user.

HaoK: That's what identity is for.  The feedback has been pretty consistent that AutomaticAuthentication is confusing.  We need it for naked challenge and the automatic cookie/bearer flow.  But for everything else we want to dial that back is our current thinking.

src/Microsoft.AspNet.Owin/OwinEnvironment.cs
Tratcher: Yeah, this will require some adaptation. OWIN middleware on ASP.NET servers will be easy because we can wrap an `Action<object>` in a `Func<object, Task>`. However, running ASP.NET components on OWIN servers will be a little awkward because we have to wrap a  `Func<object, Task>` in a `Action<object>` and block.

PinpointTownes: Nice, that's exactly what I had in mind :smile:

src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
```diff
@@ -24,9 +25,14 @@ public abstract class HttpResponse
...
 
-        public abstract void OnResponseCompleted(Action<object> callback, object state);
+        public void OnResponseStarting(Func<object, Task> callback)
+        {
+            OnResponseStarting(callback, state: null);
+        }
+
```
HaoK: @Tratcher is this sweet enough?

HaoK: Perhaps we should add one for OnResponseCompleted too?

src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
Tratcher: Lou said no sync sugar that calls async and Wait. Here you'd just wrap the Action in a Func and return a completed Task.

That said, does it confuse the intellisense for anonymous delegates if you have overloads that differ only by return value?

src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
Tratcher: How about `OnResponseStarting(Func<Task> callback) `?

Sure, add the same for OnResponseCompleted too.

PinpointTownes: We definitely need more C# 6 goodness: 

```csharp
public void OnResponseStarting(Func<Task> callback) => OnResponseStarting(state => callback(state), state: null);
```

src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
HaoK: Fair enough, updated with all the new overloads

src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
```diff
@@ -24,14 +25,33 @@ public abstract class HttpResponse
...
-        public abstract void OnResponseCompleted(Action<object> callback, object state);
+        public virtual void OnResponseStarting(Func<Task> callback) => OnResponseStarting(s => callback(), state: null);
 
-        public virtual void Redirect(string location)
-        {
-            Redirect(location, permanent: false);
-        }
+        public virtual void OnResponseStarting(Action callback) => OnResponseStarting(s => callback(), state: null);
+
+        public virtual void OnResponseStarting(Action<object> callback, object state) =>
```
Tratcher: Does the intellisense get confused with anonymous delegates between Action and Func?

src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
```diff
@@ -24,14 +25,33 @@ public abstract class HttpResponse
...
+        public abstract void OnResponseStarting(Func<object, Task> callback, object state);
 
-        public abstract void OnResponseCompleted(Action<object> callback, object state);
+        public virtual void OnResponseStarting(Func<Task> callback) => OnResponseStarting(s => callback(), state: null);
 
-        public virtual void Redirect(string location)
-        {
-            Redirect(location, permanent: false);
-        }
+        public virtual void OnResponseStarting(Action callback) => OnResponseStarting(s => callback(), state: null);
```
Tratcher: Extra closure. You can pass the callback into a static Func as state.

src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
HaoK: Yeah looks like Func wins, so it doesn't let you provide a naked Action, says wrong return type for the void Action oh well

src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
HaoK: I'm not sure I follow, aren't all of these sugar methods supposed to be delegating to abstract OnResponseStarting?

PinpointTownes: I guess @Tratcher suggests something like that:

```csharp
public void OnResponseStarting(Func<Task> callback) => OnResponseStarting(cb => ((Func<Task>) cb)(), state: callback);
```

Tratcher: Yes, but as a static method/delegate to prevent allocations for every call.

HaoK: Not following I think still, can you outline the static method that you are envisioning?

PinpointTownes:
```csharp
private static readonly Func<object, Task> _onResponseStartingDelegate = callback => ((Func<Task>) callback)();

public void OnResponseStarting(Func<Task> callback) => OnResponseStarting(_onResponseStartingDelegate, state: callback);
```

HaoK: Ah got it, so are we still not able to use this overload in Security AuthHandler after the change to a static delegate?

HaoK: Updated nuking the Action overloads, and switching both sugar methods to use the same static delegate

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
davidfowl: docs yep

HaoK:
Added a new `DisposeOnCompleted(IDisposable)` since MVC was using this all over.  Seemed generally useful.  @Tratcher what do you think about shortening the name of the OnResponseXyz methods to simply OnXyz()?  They already hang off of httpResponse, it seems a bit redundant to say response.OnResponseStarted/Completed.  

response.OnStarted/OnCompleted seems just as clear...  I'm touching all of the signatures already, so if we want to rename these, its a free opportunity

Tratcher:
Yes, OnStarted/Completed is better.

HaoK:
Updated with OnStarted/OnCompleted/OnCompetedDispose

src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
```diff
@@ -24,14 +32,17 @@ public abstract class HttpResponse
...
-        public abstract void OnResponseCompleted(Action<object> callback, object state);
+        public virtual void OnStarting(Func<Task> callback) => OnStarting(_callbackDelegate, callback);
 
-        public virtual void Redirect(string location)
-        {
-            Redirect(location, permanent: false);
-        }
+        public abstract void OnCompleted(Func<object, Task> callback, object state);
+
+        public virtual void OnCompletedDispose(IDisposable disposable) => OnStarting(_disposeDelegate, disposable);
```
Tratcher: This should be OnCompleted, not OnStarting.

HaoK: That would explain the functional failures I was seeing :)

Tratcher:
Yes, that looks right.

src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
```diff
@@ -24,14 +32,17 @@ public abstract class HttpResponse
...
-        public abstract void OnResponseCompleted(Action<object> callback, object state);
+        public virtual void OnStarting(Func<Task> callback) => OnStarting(_callbackDelegate, callback);
 
-        public virtual void Redirect(string location)
-        {
-            Redirect(location, permanent: false);
-        }
+        public abstract void OnCompleted(Func<object, Task> callback, object state);
+
+        public virtual void OnCompletedDispose(IDisposable disposable) => OnCompleted(_disposeDelegate, disposable);
```
Tratcher: `[NotNull]` for everything but `state`.

Tratcher:
:shipit:

HaoK:
5fe8037281bb826e0708abdcdafbc76571dc21f5

PR #327, opened 2015-06-12
Using 'nameof' operator instead of magic strings

hishamco:


src/Microsoft.Net.Http.Headers/RangeHeaderValue.cs
```diff
@@ -33,7 +33,7 @@ public string Unit
             get { return _unit; }
             set
             {
-                HeaderUtilities.CheckValidToken(value, "value");
+                HeaderUtilities.CheckValidToken(value, nameof(_unit));
```
rtpHarry: Just glancing over this and not very familiar with project but this looks like it will change between the original string `"value"` and `nameof(_unit)` - is it an issue?

davidfowl: Looks like an issue.

src/Microsoft.Net.Http.Headers/CookieHeaderValue.cs
```diff
@@ -39,7 +39,7 @@ public string Name
             get { return _name; }
             set
             {
-                CheckNameFormat(value, "name");
```
hishamco: @rtpHarry I replaced the string literal to the original backend field after I saw that line of code

dougbu: @hishamco the majority of the original code changed in this PR (using `"value"`) was correct. Should not `throw` `Exceptions` containing names of backing fields or other private implementation details. E.g. this one should have been `"value"` and now `nameof(value)`.

hishamco: @dougbu so I will change them and I will update the PR now

Tratcher:
:shipit:

src/Microsoft.AspNet.Owin/WebSockets/WebSocketAdapter.cs
```diff
@@ -134,7 +134,7 @@ private Task RedirectSendToCloseAsync(ArraySegment<byte> buffer, CancellationTok
                     _webSocket.Abort();
                     break;
                 default:
-                    throw new ArgumentOutOfRangeException("state", _webSocket.State, string.Empty);
+                    throw new ArgumentOutOfRangeException(nameof(_webSocket.State), _webSocket.State, string.Empty);
```
Eilon: This seems like it's not even the right exception type... there's no "argument" (i.e. a method parameter) here.

Tratcher:
Rebased and Merged.
@Eilon I changed it to a NotSupportedException.

Issue #310, opened 2015-05-18
IRequestIdentifierFeature cleanup


Tratcher:
- Rename to IHttpRequestIdenfitierFeature for consistency with other features.
- Make it settable
- Make the value a string instead of Guid.

davidfowl:
yay

PR #330, opened 2015-06-16
Refactor IRequestIdentifierFeature.

Tratcher:
#310 
Rename to IHttpRequestIdenfitierFeature for consistency with other features.
Make it settable
Make the value a string instead of Guid.

This will require minor updates in Hosting, WebListener, and Helios.

@muratg @davidfowl


muratg:
:shipit:

Yay strings!

Issue #177, opened 2015-01-21
Figure out how to make immutable header value objects


Tratcher:
Related: https://github.com/aspnet/HttpAbstractions/issues/176

Scenario:
MVC uses some lists of MediaTypeHeaderValues to determine what formats are supported. They select one from the list and use it to configure the response. Sometimes this involves modifying the value (e.g. adding charset), however this affects the value in the list and may impact matching and output for subsequent requests. The correct code would make a copy of the header value first.

Figure out how to make immutable header value object to prevent erroneous modifications.

Some ideas:
- Add a method (`MakeReadOnly()`) or property (`IsReadOnly`) to lock a header value object into read-only mode so it throws if anyone tries to make inappropriate modifications.
- Have an abstract base class with mutable and immutable implementations.

Tratcher:
@rynowak

PR #334, opened 2015-06-24
Immutable HeaderValue objects.

Tratcher:
#177 
@muratg @rynowak

### DIFF
--- a/src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
+++ b/src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Globalization;
+using System.Linq;
 using System.Text;
 
 namespace Microsoft.Net.Http.Headers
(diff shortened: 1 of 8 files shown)


src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -20,8 +21,9 @@ public class MediaTypeHeaderValue
             = new GenericHeaderParser<MediaTypeHeaderValue>(true, GetMediaTypeLength);
 
         // Use list instead of dictionary since we may have multiple parameters with the same name.
```
rynowak: Comment is stale?

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -250,22 +268,47 @@ public bool IsSubsetOf(MediaTypeHeaderValue otherMediaType)
 
         /// <summary>
         /// Performs a deep copy of this object and all of it's NameValueHeaderValue sub components,
-        /// while avoiding the cost of revalidating the components.
+        /// while avoiding the cost of re-validating the components.
```
rynowak: it was better before

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -250,22 +268,47 @@ public bool IsSubsetOf(MediaTypeHeaderValue otherMediaType)
...
+            }
+            return other;
+        }
+
+        /// <summary>
+        /// Performs a deep copy of this object and all of it's NameValueHeaderValue sub components,
+        /// while avoiding the cost of re-validating the components. This copy is read-only.
+        /// </summary>
+        /// <returns>A deep, read-only, copy.</returns>
+        public MediaTypeHeaderValue CopyAsReadOnly()
```
muratg: Perhaps `CloneAsReadOnly`?

rynowak:
:shipit: from me, seems like it would solve the issue we have.

muratg:
:shipit:

Issue #238, opened 2015-03-24
QueryHelpers.AddQueryString should support # in the URL


sornaks:
Currently, if you try to add a query string like `foo=bar` to `http://someurl.com/hello#world` it returns `http://someurl.com/hello#world?foo=bar` which should actually be `http://someurl.com/hello?foo=bar#world`

Eilon:
Feel free to assign to @sornaks when triaging.

muratg:
@sornaks are you working on this for Beta 6?

sornaks:
@muratg - yes. sent a PR.

sornaks:
Checked in. https://github.com/aspnet/HttpAbstractions/commit/25ea93de9e03f1b0be8922bf727163f8ac3be499

brockallen:
Would it make more sense to have a separate API called AddHashFragment?

Tratcher:
@brockallen We're not trying to add a hash, we're trying to append query params to a url that may already have a hash.

brockallen:
Ah, got ya.

PR #337, opened 2015-06-29
Making QueryHelpers.AddQueryString support # in the URL.

sornaks:
https://github.com/aspnet/HttpAbstractions/issues/238


test/Microsoft.AspNet.WebUtilities.Tests/QueryHelpersTests.cs
```diff
@@ -52,5 +52,39 @@ public void ParseQueryWithEmptyKeyWorks()
...
+        [Theory]
+        [InlineData("http://contoso.com/", "http://contoso.com/?hello=world")]
+        [InlineData("http://contoso.com/someaction", "http://contoso.com/someaction?hello=world")]
+        [InlineData("http://contoso.com/someaction?q=test", "http://contoso.com/someaction?q=test&hello=world")]
+        [InlineData(
+            "http://contoso.com/someaction?q=test#anchor",
+            "http://contoso.com/someaction?q=test&hello=world#anchor")]
+        [InlineData("http://contoso.com/someaction#anchor", "http://contoso.com/someaction?hello=world#anchor")]
+        [InlineData("http://contoso.com/#anchor", "http://contoso.com/?hello=world#anchor")]
+        public void AddQueryStringWithKeyAndValue(string uri, string expectedUri)
```
Tratcher: Try `http://contoso.com/someaction?q=test#anchor?value` and `http://contoso.com/someaction?q=test#anchor#value`

Tratcher: Split is not your friend. IndexOf is more appropriate.

sornaks: Just confirming - Is the expectation in `http://contoso.com/someaction?q=test#anchor?value` case `http://contoso.com/someaction?q=test#anchor?value&sometext`?

Tratcher: No, `http://contoso.com/someaction?q=test#anchor?value` + `name=value` => `http://contoso.com/someaction?q=test&name=value#anchor?value`. Only the first `?` and `#` in a url are special.
Also consider:
`http://contoso.com/someaction#anchor?stuff` + `name=value` => `http://contoso.com/someaction?name=value#anchor?stuff`

sornaks:
Updated.. Please have a look.

src/Microsoft.AspNet.WebUtilities/QueryHelpers.cs
```diff
@@ -43,7 +64,8 @@ public static string AddQueryString([NotNull] string uri, [NotNull] IDictionary<
                 sb.Append(UrlEncoder.Default.UrlEncode(parameter.Value));
                 hasQuery = true;
             }
-            return sb.ToString();
+
+            return sb.ToString() + anchorText;
```
Eilon: Should probably append `anchorText` to the `sb` and then call `.ToString()` on it to prevent the extra string being created.

sornaks: Yes. That makes sense.

src/Microsoft.AspNet.WebUtilities/QueryHelpers.cs
```diff
@@ -20,8 +20,7 @@ public static class QueryHelpers
         /// <returns>The combined result.</returns>
         public static string AddQueryString([NotNull] string uri, [NotNull] string name, [NotNull] string value)
         {
-            bool hasQuery = uri.IndexOf('?') != -1;
-            return uri + (hasQuery ? "&" : "?") + UrlEncoder.Default.UrlEncode(name) + "=" + UrlEncoder.Default.UrlEncode(value);
+            return AddQueryString(uri, new Dictionary<string, string> { { name, value } });
```
Tratcher: Perf: dictionaries aren't cheep. Do not create a new one just to add one key-value pair.

davidfowl: +1

src/Microsoft.AspNet.WebUtilities/QueryHelpers.cs
```diff
@@ -29,12 +28,34 @@ public static string AddQueryString([NotNull] string uri, [NotNull] string name,
...
+                anchorText = uri.Substring(anchorIndex);
+                uriToBeAppended = uri.Substring(0, anchorIndex);
+            }
+
+            var queryIndex = uriToBeAppended.IndexOf('?');
+            var hasQuery = queryIndex != -1;
+
+            // If there is a second query, then the query string must be appended to the first occurance.
+            // eg. http://contoso.com/someaction?hello?world + name=value
+            // => http://contoso.com/someaction?hello&name=value?world
```
Tratcher: This sample is wrong, it should be:
```
http://contoso.com/someaction?hello?world + name=value =>
http://contoso.com/someaction?hello?world&name=value
```
Only the first `?` is significant.

src/Microsoft.AspNet.WebUtilities/QueryHelpers.cs
Eilon: They're not cheap, but we also don't want to duplicate a lot of code... I'm not sure that this method is much of a hot path.

The only other thought I have that doesn't duplicate code is to create a new degenerate dictionary type called `SingleValueDictionary` that takes in `string key, string value` and basically hard-codes all the `IDictionary` stuff to just be aware of the single key/value pair. Probably not a lot of code to write, but not sure it's worth it.

Tratcher: You don't need a custom `IDictionary` to share code, you just need an `IEnumerable<KeyValuePair...>` which you could get with a `IList<T>`.

Eilon: O I C. That would mean having both the public methods call into a new 3rd private method that worked off of `IEnumerable<KVP<,>>` right? That does seem reasonable to me.

sornaks:
Updated..

Tratcher:
:shipit:

sornaks:
Checked in. Thanks! https://github.com/aspnet/HttpAbstractions/commit/25ea93de9e03f1b0be8922bf727163f8ac3be499

PR #344, opened 2015-07-09
Introducing IHtmlContent - an interface which lets any content define how to write itself.

sornaks:


src/Microsoft.AspNet.Html.Abstractions/IHtmlContent.cs
```diff
@@ -0,0 +1,25 @@
...
+using Microsoft.Framework.WebEncoders;
+
+namespace Microsoft.AspNet.Html.Abstractions
+{
+    /// <summary>
+    /// Represents a content which can write itself.
+    /// </summary>
+    /// <remarks>
+    /// This interface provides a uniform way for deferred content-building.
+    /// </remarks>
```
rynowak: Doesn't mean anything, just leave it out.

src/Microsoft.AspNet.Html.Abstractions/IHtmlContent.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+using Microsoft.Framework.WebEncoders;
+
+namespace Microsoft.AspNet.Html.Abstractions
+{
+    /// <summary>
+    /// Represents a content which can write itself.
+    /// </summary>
```
rynowak: Eww 'represents' rears its ugly head again. You can take it out and the meaning is unchanged.

I'd say "HTML content which can be written to a TextWriter"

src/Microsoft.AspNet.Html.Abstractions/project.json
```diff
@@ -0,0 +1,20 @@
...
+    },
+    "dependencies": {
+        "Microsoft.Framework.WebEncoders.Core": "1.0.0-*"
+    },
+    "frameworks": {
+        "net45" : { },
+        "dnx451": { },
+        "dnxcore50": {
+            "dependencies": {
+                "System.Diagnostics.Debug": "4.0.10-beta-*"
```
rynowak: ???

src/Microsoft.AspNet.Html.Abstractions/project.json
```diff
@@ -0,0 +1,20 @@
+﻿{
+    "version": "1.0.0-*",
+    "description": "ASP.NET 5 HTML content interface.",
+    "repository": {
+        "type": "git",
+        "url": "git://github.com/aspnet/httpabstractions"
+    },
+    "dependencies": {
+        "Microsoft.Framework.WebEncoders.Core": "1.0.0-*"
```
rynowak: Why core? Isn't there an abstractions package for this?

rynowak:
:watch:

src/Microsoft.AspNet.Html.Abstractions/project.json
sornaks: You mean `WebUtilities` package? I could use that, but this has all the essential interfaces. And I found the `Core` package to be used in other Abstractions projects as well..

sornaks:
Updated..

src/Microsoft.AspNet.Html.Abstractions/project.json
rynowak: Ok, I think you're right then.

rynowak:
:shipit:

sornaks:
Checked in. Thanks! https://github.com/aspnet/HttpAbstractions/commit/24f90cc91427e89aeea6ac5b127014dc09f9ff6b

Issue #339, opened 2015-07-01
NotImplementedExceptions in FeatureCollection


ardalis:
Why does FeatureCollection have NotImplementedExceptions when it is implementing IFeatureCollection, for which one would expect perfect parity? If there are parts of that interface that FeatureCollection doesn't use, why are they on the interface?

davidfowl:
NotImplementedException is most likely wrong and should be not supported exception.

ardalis:
I'm not sure that's an improvement. Why not adjust IFeatureCollection, which is literally in the same project and folder, to match FeatureCollection's capabilities (or vice versa). Yes, IFeatureCollection inherits from a collection interface, but if that collection interface has too many features, choose a different collection interface to inherit from. Otherwise this is violating Liskov Substitution Principle and resulting in extra bloat in FeatureCollection.

Tratcher:
Those methods are all inherited from IDictionary. They weren't implemented because nobody needed them yet. We never attempt to consume the collection as an IDictionary and I don't think we need anything beyond TryGetValue and Set/indexer.

Recommendation: Remove IDictionary from IFeatureCollection and just add the methods we need.

ardalis:
@Tratcher +1 to your recommendation.

muratg:
Makes sense to me too.

davidfowl:
@Tratcher We want to preserve enumeration also.

lodejard:
Yep, it started as a get/set interface and dictionary was added when enumeration was needed. Out of curiosity, would it make sense to implement the idictionary interface instead of declaring a get/set/enumerable subset?

Tratcher:
IDictionary has 10x more functionality/complexity than we need. This collection is also never exposed directly to the application, it only exists between the server and Hosting/DefaultHttpContext. I'm drafting a subset API list now, but so far the only APIs we actually use are Get and Set. Enumeration is a nice to have, but it's not actually used anywhere or available to the app.

Keeping the surface area to a minimum also makes implementing https://github.com/aspnet/HttpAbstractions/issues/353 so much easier and more efficient.

lodejard:
I agree, but lets keep enumeration (of Type) available to support diagnostic scenarios of displaying all features and writing out all reflected properties.

Sent from mobile device
________________________________

PR #356, opened 2015-07-28
Reduce IFeatureCollection surface area.

Tratcher:
#339 
@muratg @lodejard @ardalis @davidfowl


muratg:
Does this break other repos?

Tratcher:
Just the servers, and those should be easy to fix.

src/Microsoft.AspNet.Http.Features/FeatureCollection.cs
```diff
@@ -78,89 +64,39 @@ IEnumerator IEnumerable.GetEnumerator()
...
         }
 
-        public ICollection<Type> Keys
+        private class KeyComparer : IEqualityComparer<KeyValuePair<Type, object>>
         {
-            get { throw new NotImplementedException(); }
-        }
+            public bool Equals(KeyValuePair<Type, object> x, KeyValuePair<Type, object> y)
+            {
+                return x.Key.Equals(y.Key);
```
muratg: Just comparing keys is enough to call these "Equals"?  It feels a little weird, i.e. {"x":1} and {"x":2} would be equal per this.

Tratcher: This is only used on line 77 to filter out features that exist in both collections, so we only want to match by key.

ardalis: It's public. It may only be used by this class for that purpose, but what if others choose to use it?

Tratcher: No, the KeyComparer class is private.

ardalis: So it is. Never mind. :)

Tratcher:
Updated to lazy initialize the feature storage. In the average case we'll only be reading from the default/fallback, not making changes.

Tratcher:
Re: https://github.com/aspnet/KestrelHttpServer/commit/f681599e4807f1332e760c1aebfd0c32b3ee1037

Tratcher:
Re: https://github.com/aspnet/WebListener/commit/161fdbb4b6f476a58e178b07eec96af574cfa642

Tratcher:
@Eilon to review, but we don't want to check in the break until we get Helios building again.

Eilon:
:shipit: from me. If you wanted to avoid breaking changes you could probably keep the `.Add` method at first, check this change in, then update Kestrel and WebListener to use the indexer, and finally remove the `.Add` method. But I don't care much either way.

Tratcher:
@Eilon it doesn't matter if I break Helios now, Andrew checked in a breaking change to ILibraryManager on Friday so the old Helios assemblies won't run anymore.

Eilon:
@Tratcher thanks, I'm fine either way.

Tratcher:
https://github.com/aspnet/Helios/commit/1ee51e137becb7e3ae03d98b568fec0862657d52

Tratcher:
FYI: Updated WebListener to use the same style static feature collection as Helios: https://github.com/aspnet/WebListener/commit/e69cb5c911b8dfaef217d0912fade5b510ccd816

PR #358, opened 2015-08-03
Changing QueryHelpers.AddQueryString to use Array instead of List.

sornaks:


Eilon:
The code is :shipit: 

Did it show up in a perf profile?

sornaks:
No. I did not profile this.

Eilon:
The change is fine, but let's try to focus on stuff we see in profiles.

Issue #341, opened 2015-07-06
New method: HttpRequest.GetFullUrl 


muratg:
Add a new extension method HttpRequest (GetFullUrl) that returns a string. 

- GetEncodedUrl
- GetDecodedUrl

Tratcher:
We've won't-fixed this request a few times already. What's different now?
#110, #308

davidfowl:
1. It's commonly asked for
2. It's a string so we avoid all of the Uri problems with round tripping

Tratcher:
No, we still have many of the round tripping problems.  What format should all of the components be in? Human readable or encoded for use in headers? I suspect you need at least two methods.

davidfowl:
Not sure why we'd have roundtripping problems. It's just a concat of the parts of the URL that we already have

Tratcher:
The values have already been processed by multiple layers and are in different formats. e.g. Paths are unescaped, query is still escaped, who knows what state the Host is in, etc.. Just concatenating them back together gives you a Frankenstein url. You have to consider both the current format of the data and the desired format/usage of the result.

This is one of the reasons we haven't done this. The only reasonable request I've heard so far for the full url was for logging.  What other usages have been requested? Almost everybody that has asked for this ends up just wanting the pieces, not the full url.

borgdylan:
Would this include the query string as well? While developing I print each request path to the console to make sure stuff gets called by JS. Having the query string as well would be nice.

TomGroeneboer:
I think you can get the querystring with `HttpContext.GetFeature<IHttpRequestFeature>() ` https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http.Features/IHttpRequestFeature.cs

Tratcher:
@TomGroeneboer You can get it from `HttpContext.Request.QueryString`.

Eilon:
We should re-discuss the names of these. Those the current names are technically correct, I don't believe they will lead to correct usage. The doc comments explain what's what, but I'm sure they'll be often ignored. I recommend perhaps changing `GetDecodedUrl()` to be something more like `GetDisplayUrl()`, or perhaps even getting rid of it entirely.

PR #359, opened 2015-08-03
Add HttpReqeuest GetEncodedUrl and GetDecodedUrl extensions.

Tratcher:
#341 
@davidfowl @Eilon

### DIFF
--- a/src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
+++ b/src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
@@ -71,5 +71,27 @@ namespace Microsoft.AspNet.Http.Extensions
                 return uri.GetComponents(UriComponents.SerializationInfoString, UriFormat.UriEscaped);
             }
         }
+
+        /// <summary>
+        /// Returns the combined components of the request URL in a fully escaped form suitable for use in HTTP headers
+        /// and other HTTP operations.
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns></returns>
+        public static string GetEncodedUrl(this HttpRequest request)
+        {
+            return Encode(request.Scheme, request.Host, request.PathBase, request.Path, request.QueryString);
+        }
+
+        /// <summary>
+        /// Returns the combined components of the request URL in a fully un-escaped form (except for the QueryString)
+        /// suitable only for display. This format should not be used in HTTP headers or other HTTP operations.
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns></returns>
+        public static string GetDisplayUrl(this HttpRequest request)
+        {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
```diff
@@ -71,5 +71,27 @@ public static string Encode(Uri uri)
...
+
+        /// <summary>
+        /// Returns the combine components of the request URL in a fully un-escaped form (except for the QueryString)
+        /// suitable only for display. This format should not be used in HTTP headers or other HTTP operations.
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns></returns>
+        public static string GetDecodedUrl(this HttpRequest request)
+        {
+            return request.Scheme + "://" + request.Host.Value + request.PathBase.Value + request.Path.Value + request.QueryString.Value;
```
Eilon: Does all this concatenation work when some sections are missing? E.g. no query string or empty path etc.?

Eilon:
Also commented in the bug:

> We should re-discuss the names of these. Those the current names are technically correct, I don't believe they will lead to correct usage. The doc comments explain what's what, but I'm sure they'll be often ignored. I recommend perhaps changing  GetDecodedUrl()  to be something more like  GetDisplayUrl() , or perhaps even getting rid of it entirely.

Tratcher:
Renamed GetDecodedUrl to GetDisplayUrl.

src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
```diff
@@ -71,5 +71,27 @@ public static string Encode(Uri uri)
                 return uri.GetComponents(UriComponents.SerializationInfoString, UriFormat.UriEscaped);
             }
         }
+
+        /// <summary>
+        /// Returns the combine components of the request URL in a fully escaped form suitable for use in HTTP headers
```
Eilon: `combined`

src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
```diff
@@ -71,5 +71,27 @@ public static string Encode(Uri uri)
...
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns></returns>
+        public static string GetEncodedUrl(this HttpRequest request)
+        {
+            return Encode(request.Scheme, request.Host, request.PathBase, request.Path, request.QueryString);
+        }
+
+        /// <summary>
+        /// Returns the combine components of the request URL in a fully un-escaped form (except for the QueryString)
```
Eilon: `combined`

Eilon:
:shipit: after the comment fix.

PR #369, opened 2015-08-12
Move HttpContextAccessor to HttpAbstractions

HaoK:
cc @davidfowl @Tratcher


src/Microsoft.AspNet.Http.Abstractions/HttpContextAccessor.cs
```diff
@@ -0,0 +1,47 @@
...
+
+using System;
+#if DNX451
+using System.Runtime.Remoting.Messaging;
+using System.Runtime.Remoting;
+#elif DNXCORE50
+using System.Threading;
+#endif
+
+namespace Microsoft.AspNet.Http
```
Tratcher: Http.Internal?

Tratcher:
:shipit:

src/Microsoft.AspNet.Http.Abstractions/HttpContextAccessor.cs
HaoK: Hosting is going to need to use this still, but I don't know what the guidance is on the internal namespace.

HaoK: Seems a little weird to put it in Internal when its another Repo using it though?

Tratcher: The internal namespace is for framework infrastructure that's probably not used by the general public.
Compare to DefaultHttpContext which Hosting also uses: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http/DefaultHttpContext.cs#L15

HaoK: Sounds reasonable, but there's nothing else in Internal in HttpAbstractions currently, but I can create the folder and move it there if you think its the right thing do

HaoK:
Updated PR moving implementation to internal namespace, I didn't create a folder since there is no folder in the other projects in Http currently

src/Microsoft.AspNet.Http.Abstractions/HttpContextAccessor.cs
```diff
@@ -0,0 +1,47 @@
...
+
+using System;
+#if DNX451
+using System.Runtime.Remoting.Messaging;
+using System.Runtime.Remoting;
+#elif DNXCORE50
+using System.Threading;
+#endif
+
+namespace Microsoft.AspNet.Http.Internal
```
Tratcher: Oh, I just realized this was in the Abstractions package. This implementation should move to the Http package.

HaoK: Ok that makes things feel much nicer :)

Tratcher:
:shipit: :shipit:

PR #372, opened 2015-08-15
Add Api for HTTP/2 PushPromise

### LINKED ISSUE #371: Add Api for HTTP/2 PushPromise
Implement the PushPromise api as virtual NoOp in the abstract class HttpResponse. 
## Background

This API is to support applications sending push promises to HTTP 2.0 clients.

[Current .NET v4.6 Api](https://msdn.microsoft.com/en-us/library/dn823340%28v=vs.110%29.aspx)

``` csharp
public void PushPromise( string path )

public void PushPromise(
    string path,
    string method,
    NameValueCollection headers
)
```

**Parameters**
_path_ Type: System.String
The URL of the push request. It should be the virtual path of the relative resource that the server wants to push to the client.

_method_ Type: System.String
Http request method that would be used by the push request.

_headers_ Type: System.Collections.Specialized.NameValueCollection
Http request header that would be used by the push request.

Suggested addition to [Microsoft.AspNet.Http.Abstractions/HttpResponse.cs](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs)

**Remarks**
PushPromise is non-deterministic and applications shouldn't have logic that depends on it. Its only purpose is performance advantage in some cases. There are many conditions (protocol and implementation) that may cause to ignore the push requests completely. The expectation is based on fire-and-forget.
## Rationale

Its _out-of-band_ data like `headers` and `cookies` and is part of the response so should live at the `HttpResponse` level.

It is a core feature of http/2 rather than a extended addition that might be `HttpContext` related.

Its now standardised and implemented in all browsers - so is something I assume will be implemented in ASP.NET 5.

It already comes with two huge caveats: 
- _applications shouldn't have logic that depends on it_
- _There are many conditions (protocol and implementation) that may cause to ignore the push requests completely._

Which means it doesn't actually have to do anything at this time.
## Suggested Api

``` csharp
public virtual void PushPromise(string path);
public virtual void PushPromise(string path, string method, IHeaderDictionary headers);
```

benaadams:
Resolves #371


davidfowl:
It's dangerous to add a public API like this without validating the implementation on various servers. This isn't something we'd do before actually implementing it. We can always add APIs but we can't remove them, so it's safer to leave it until the rest is done

benaadams:
Fair enough; will tweak the Issue a little then :)

PR #377, opened 2015-08-19
Fix #365 - Make IFormFileCollection implmenent IReadOnlyList<T>

### LINKED ISSUE #365: Consider making IFormFileCollection read-only
`IFormCollection` is readonly, `IFormFileCollection` is not.

https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http.Abstractions/IFormCollection.cs
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http.Abstractions/IFormFileCollection.cs

This is a bit of a speedbump for fixing aspnet/Mvc#2527 and would otherwise require us to duplicate the code in `FormFileCollection`.

rynowak:
/cc @Tratcher


Tratcher:
:shipit:

Issue #361, opened 2015-08-04
Add StringValues struct for collections


lodejard:
Streamlines all of the cases where string,string[] collections cause friction

Also has side-effect of reducing number of allocations per request

Tratcher:
What did the CLR team have to say about this?

Also, layering. Ideally this type would be in a BCL package, not one of ours.

Eilon:
As with many types, we should start out with our own and see where it goes.

davidfowl:
+1

lodejard:
I'm not sure this is *necessarily* a BCL/CLR asset, we certainly don't need to be gated by that for packaged code. Nothing else in HttpAbstractions and web server feature interfaces was punched down that far.

davidfowl:
@Tratcher is concerned because of the layering of WebListener. The lower layer doesn't depend on anything except bcl types today

Eilon:
WebListener is a DNX-only component in practice, but either way, having it depend on another package is really not so bad.

Tratcher:
@lodejard Do you have your prototype code in a branch? Or paste it here?

lodejard:
yep, https://github.com/aspnet/HttpAbstractions/tree/lodejard/stringvalues

PR #378, opened 2015-08-21
Introduce StringValues to replace string[] usage.

Tratcher:
#361
Open issues:
- Name and location of the package containing StringValues. It needs to make sense to reference even from Microsoft.Net.Http.Server (WebListener's low level API without Asp.Net dependencies). Microsoft.AspNet.Primitives is just a placeholder.

Related breaking changes:
- IHeaderDictionary:
 - Removal of AppendValues, Get*, and Set*
 - Append now functions like AppendValues instead of AppendCommaSeperatedValues.
- IReadableStringCollection
 - Removal of Get*, use the indexer instead.

Here are some of the necessary changes in downstream repos. Please use them only as references to evaluate this PR, lets avoid full reviews on them until we have settled open issues on this one.
- https://github.com/aspnet/Hosting/commit/49b8b293027e9432fb6c44cd6c43d0f875b7d814
- https://github.com/aspnet/Helios/commit/724cb5823199d93987a4cb97f8fdfffbb498bde4
- https://github.com/aspnet/WebListener/commit/d858bf73fd2f8477858e47f7280a1f6785e9b0a5
- https://github.com/aspnet/KestrelHttpServer/commit/2abc40692577227ad0599b19915c5885ac979338
- https://github.com/aspnet/Security/commit/25af3ee9f71447f5de42add2d4aa8bce7cfb7ba1
- https://github.com/aspnet/Mvc/commit/6331a9c1f4a2e6fff43a2535976ce2dde43bcaa6

@lodejard @davidfowl @muratg @rynowak @halter73 @Eilon

### DIFF
--- a/samples/SampleApp/Program.cs
+++ b/samples/SampleApp/Program.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Diagnostics;
+using Microsoft.Framework.Primitives;
+
+namespace SampleApp
+{
+    public class Program
+    {
+        public void Main(string[] args)
+        {
+            for (int i = 0; i < 10; i++)
+            {
+                Stopwatch timer = new Stopwatch();
+                timer.Start();
+                string myString;
+                string[] myArray;
+                StringValues myValues;
+                for (int j = 0; j < 100000000; j++)
+                {
+                    myString = new string('a', 40);
+                    myArray = new[] { myString };
+                    // myValues = new StringValues(myString);
+                    myValues = new StringValues(myArray);
+                }
+                timer.Stop();
+                Console.WriteLine(timer.Elapsed + ", " + Environment.WorkingSet);
+            }
+        }
+    }
+}
--- a/samples/SampleApp/project.json
+++ b/samples/SampleApp/project.json
@@ -1,0 +1,15 @@
+﻿{
+  "version": "1.0.0-*",
+
+  "dependencies": {
+    "Microsoft.AspNet.Http": "1.0.0-*"
+  },
+
+  "commands": {
+    "SampleApp": "SampleApp"
+  },
+
+  "frameworks": {
+    "dnx451": { }
+  }
+}
--- a/src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryTypeExtensions.cs
+++ b/src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryTypeExtensions.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Reflection;
 using Microsoft.AspNet.Http.Headers;
 using Microsoft.Framework.Internal;
+using Microsoft.Framework.Primitives;
 using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Http
@@ -50,7 +51,7 @@ namespace Microsoft.AspNet.Http
             }
             else
             {
-                headers.SetValues(name, values.Select(value => value.ToString()).ToArray());
+                headers[name] = values.Select(value => value.ToString()).ToArray();
             }
         }
 
@@ -98,7 +99,7 @@ namespace Microsoft.AspNet.Http
             }
 
             var value = headers[name];
-            if (string.IsNullOrWhiteSpace(value))
+            if (StringValues.IsNullOrEmpty(value))
             {
                 return default(T);
             }
@@ -112,11 +113,11 @@ namespace Microsoft.AspNet.Http
             if (KnownListParsers.TryGetValue(typeof(T), out temp))
             {
                 var func = (Func<IList<string>, IList<T>>)temp;
-                return func(headers.GetValues(name));
+                return func(headers[name]);
             }
 
-            var values = headers.GetValues(name);
-            if (values == null || !values.Any())
+            var values = headers[name];
+            if (StringValues.IsNullOrEmpty(values))
             {
                 return null;
             }
@@ -158,7 +159,7 @@ namespace Microsoft.AspNet.Http
             return default(T);
         }
 
-        private static IList<T> GetListViaReflection<T>(IList<string> values)
+        private static IList<T> GetListViaReflection<T>(StringValues values)
         {
             // TODO: Cache the reflected type for later? Only if success?
             var type = typeof(T);
--- a/src/Microsoft.Framework.Primitives/StringValues.cs
+++ b/src/Microsoft.Framework.Primitives/StringValues.cs
@@ -1,0 +1,233 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.Framework.Primitives
+{
+    /// <summary>
+    /// Represents zero/null, one, or many strings in an efficient way.
+    /// </summary>
+    public struct StringValues : IList<string>
+    {
+        private static readonly string[] EmptyArray = new string[0];
+        public static readonly StringValues Empty = new StringValues(EmptyArray);
+
+        private readonly string _value;
+        private readonly string[] _values;
+
+        public StringValues(string value)
+        {
+            _value = value;
+            _values = null;
+        }
+
+        public StringValues(string[] values)
+        {
+            _value = null;
+            _values = values;
+        }
+
+        public static implicit operator StringValues(string value)
+        {
+            return new StringValues(value);
+        }
+
+        public static implicit operator StringValues(string[] values)
+        {
+            return new StringValues(values);
+        }
+
+        public static implicit operator string (StringValues values)
+        {
+            return values.GetStringValue();
+        }
+
+        public static implicit operator string[] (StringValues value)
+        {
+            return value.GetArrayValue();
+        }
+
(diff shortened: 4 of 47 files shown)


dougbu:
@Tratcher note aspnet/Mvc@6331a9c needs to be updated now that @rynowak has pushed aspnet/Mvc@6d365e9

samples/SampleApp/Program.cs
```diff
@@ -0,0 +1,30 @@
...
+                Stopwatch timer = new Stopwatch();
+                timer.Start();
+                string myString;
+                string[] myArray;
+                StringValues myValues;
+                for (int j = 0; j < 100000000; j++)
+                {
+                    myString = new string('a', 40);
+                    myArray = new[] { myString };
+                    // myValues = new StringValues(myString);
```
dougbu: ??

samples/SampleApp/project.json
```diff
@@ -0,0 +1,15 @@
...
+  "dependencies": {
+    "Microsoft.AspNet.Http": "1.0.0-*"
+  },
+
+  "commands": {
+    "SampleApp": "SampleApp"
+  },
+
+  "frameworks": {
+    "dnx451": { }
```
dougbu: Why not run on .NET Core 5.0?

Tratcher:
@dougbu updated: https://github.com/aspnet/Mvc/commit/b42d5015e4010537da5f1c578771d3ff837424a1

src/Microsoft.AspNet.Http.Abstractions/WebSocketManager.cs
```diff
@@ -11,7 +11,7 @@ public abstract class WebSocketManager
     {
         public abstract bool IsWebSocketRequest { get; }
 
-        public abstract IList<string> WebSocketRequestedProtocols { get; }
+        public abstract StringValues WebSocketRequestedProtocols { get; }
```
davidfowl: Why?

Tratcher: It's an http header with 0-many values.

Tratcher:
Offline review notes:

Rename AspNet.Primitives -> Framework.Primitives

Add StringValues.ToArray() for explicit typing (instead of casting), mirrors ToString

IHeaderDictionary:
- Move to extension methods:
 - StringValues GetCommaSeparatedValues(string key); - Return `IList<string>`
 - void Append(string key, StringValues value);
 - void AppendCommaSeparatedValues(string key, params string[] values);
 - void SetCommaSeparatedValues(string key, params string[] values);

src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryTypeExtensions.cs
```diff
@@ -112,11 +113,11 @@ internal static IList<T> GetList<T>([NotNull] this IHeaderDictionary headers, st
...
             {
                 var func = (Func<IList<string>, IList<T>>)temp;
-                return func(headers.GetValues(name));
+                return func(headers[name]);
             }
 
-            var values = headers.GetValues(name);
-            if (values == null || !values.Any())
+            var values = headers[name];
+            if (StringValues.IsNullOrEmpty(values))
```
davidfowl: I think this usage shows some of the benefits of the API. Checking for null or !any is pretty annoying.

src/Microsoft.Framework.Primitives/StringValues.cs
```diff
@@ -0,0 +1,233 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.Framework.Primitives
+{
+    /// <summary>
+    /// Represents zero/null, one, or many strings in an efficient way.
+    /// </summary>
+    public struct StringValues : IList<string>
+    {
+        private static readonly string[] EmptyArray = new string[0];
```
KrzysztofCwalina: You might want to use Array.Empty. I thought @stephentoub added it a couple of month ago.

src/Microsoft.Framework.Primitives/StringValues.cs
```diff
@@ -0,0 +1,233 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.Framework.Primitives
+{
+    /// <summary>
+    /// Represents zero/null, one, or many strings in an efficient way.
+    /// </summary>
+    public struct StringValues : IList<string>
```
KrzysztofCwalina: I wonder if we should not simply write a general purpose collection generic collection that does that, i.e. OneOrMoreList<T>. @terrajobst, any thoughts on this?

src/Microsoft.Framework.Primitives/StringValues.cs
```diff
@@ -0,0 +1,233 @@
...
+        {
+            return new StringValues(values);
+        }
+
+        public static implicit operator string (StringValues values)
+        {
+            return values.GetStringValue();
+        }
+
+        public static implicit operator string[] (StringValues value)
```
KrzysztofCwalina: This returns the private array. Might be surprising to many people. Maybe it should be a method or the return type should be one of the red-only collection wrappers?

src/Microsoft.Framework.Primitives/StringValues.cs
Tratcher: @KrzysztofCwalina Array.Empty isn't available in .NET 4.5.1.

src/Microsoft.Framework.Primitives/StringValues.cs
Tratcher: @KrzysztofCwalina  It's a mirror of the incoming converter. string[] -> StringValues -> string[]. If the user just wants a read-only IList they can use StringValues directly.

src/Microsoft.Framework.Primitives/StringValues.cs
Tratcher: @KrzysztofCwalina A BCL generic would be nice. The only special cases here are the static IsNullOrEmpty and Concat methods.

Tratcher: Actually not Concat, I was thinking of ToString which appends comma array values with commas.

Tratcher:
Updated.

Tratcher:
@davidfowl @lodejard Can we get this wrapped up today so we can start pushing it through the other repos?

src/Microsoft.Framework.Primitives/StringValues.cs
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        public string[] ToArray()
+        {
+            return GetArrayValue() ?? EmptyArray;
+        }
+
+        private string[] GetArrayValue()
+        {
+            if (_value != null)
```
lodejard: This method should probably check for _values == null instead of _value != null. In all of the other methods the _values being null or not is the decision maker

src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryTypeExtensions.cs
lodejard: :heart: IsNullOrEmpty

src/Microsoft.Framework.Primitives/StringValues.cs
```diff
@@ -0,0 +1,233 @@
...
+        {
+            return values.GetStringValue();
+        }
+
+        public static implicit operator string[] (StringValues value)
+        {
+            return value.GetArrayValue();
+        }
+
+        public int Count => _values?.Length ?? (_value != null ? 1 : 0);
```
davidfowl: I love this line of code :smile:

rynowak: I had this in my VPR PR and @pranavkm hated it.

src/Microsoft.Framework.Primitives/StringValues.cs
Tratcher: In this case you still need to check if _value != null because you don't want to do return new[] { _value }; with a null _value:
```
            if (_values == null)
            {
                if (_value != null)
                {
                    return new[] { _value };
                }
            }
            return _values;
```
or
```

            if (_value == null)
            {
                return _values;
            }
            return new[] { _value };
```
or
```

            if (_value != null)
            {
                return new[] { _value };
            }
            return _values;
```

Tratcher:
Offline :shipit: from @lodejard

davidfowl:
it is time :shipit:

lodejard:
Finally found some cellular data! :shipit:

src/Microsoft.Framework.Primitives/StringValues.cs
lodejard: const keyword on arrays would have been nice :)

src/Microsoft.Framework.Primitives/StringValues.cs
lodejard: I dunno, I'd be really nervous about having one method with a different logical order to determine state. It's `if values !=null then values, else if value!=null then string[1]{value}, else string[0]`. See also Count property. A comment next to the fields would have been nice, sorry about that :/

Tbh that means the string ctor could really detect "null" and assign values to a `string[1]{null}` instead.

That said, it can be refined after the merge, so treat this like a followup remark and keep running with the shipit

PR #379, opened 2015-08-22
Change "AddMiddleware" to upcoming "BuildFunc"

khellang:
This PR changes the `AddMiddleware` delegate to the `BuildFunc` that'll (hopefully) end up in the OWIN v1.1.0 spec.

The difference between the two is that `BuildFunc` (and `MidFactoryFunc`, which replaces the `CreateMiddleware` delegate) takes in the startup properties in order to construct the actual `MidFunc`.

I'm not sure how you feel about pulling this before anything has landed in an official spec, but I thought it'd be nice to open this PR as early as possible.
This will break middleware which already has a dependency on the `AddMiddleware` delegate, like Nancy, so it's better to pull it sooner rather than later IMO.

An alternative to breaking the middleware already working with the current `AddMiddleware` delegate could be to add this as an overload. That might be worth exploring.


Tratcher:
I haven't seen much discussion on  the subject recently. Do you have a reference? We'd prefer not to update until it stabilizes, but forking to prototype is encouraged.

khellang:
Not really. We talked about it on an OWIN call on wednesday. @serialseb is working on the 1.1.0 version of the spec, including `MidFunc`, at https://github.com/openrasta/owin.spec/ and I was hoping to get the conversation going (again) to include a `BuildFunc` into that as well. I raised an issue at https://github.com/owin/owin/issues/39. Let's see if anyone cares :stuck_out_tongue_closed_eyes:

davidfowl:
@khellang any updates?

khellang:
The PR including BuildFunc is here; https://github.com/owin/owin/pull/41.

muratg:
I don't see much discussion on this in either PR.  If this needs to happen in ASP.NET 5, we need to reach a decision soon.

serialseb:
The spec and it's proposal have been stable for a year now, I think it's safe to say that there will not be signature changes. Feedback from implementers on the spec however are more than welcome, as they have been for a year.

serialseb:
That said if it helps things on your side for "stability", then we can trigger a vote and pass the spec as an RC next week, I'm not aware of any significant reason why this would not happen.

Tratcher:
By all means @serialseb, it would be nice to see it finalized on your side.

panesofglass:
I thought we *did* finalize the type signature, if not the exactness of the full wording. Draft is here: http://owin.org/html/spec/middleware-1.0.0-draft.1.html ([source](https://github.com/owin/owin/blob/master/spec/middleware-1.0.0-draft.1.md)).

khellang:
Doing some spring cleaning on my PR backlog. Don't see this going anywhere... Closing.

PR #380, opened 2015-08-28
Fix #343 - Avoid going to disk when reading the form

### LINKED ISSUE #343: Reading the Form data goes to disk every time
Found while profiling simple model binding scenarios in MVC

`BufferingHelper.EnableRewind` will do a check for the existence of the temp directly on each request, even when there isn't enough data to force a write to disk. 
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http/BufferingHelper.cs#L40

Discussed with @Tratcher - we can have this wait until a write is actually needed to do the write, or even ask forgiveness instead of asking permission.

rynowak:
This change tries to avoid looking up the TEMP directory for most reads of
the form data. We'll now only hit the disk when necessary.


NuGet.Config
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
davidfowl: undo

rynowak:
I went with both caching the lookup of the TEMP directory **and** making it deferred. Let me know if you want to back away from either of these.

/cc @Tratcher  @davidfowl

NuGet.Config
rynowak: This is super-star level BOM. why would you not want that?

Tratcher:
:shipit:

davidfowl:
:shipit:

src/Microsoft.AspNet.Http/BufferingHelper.cs
```diff
@@ -12,21 +12,31 @@ public static class BufferingHelper
...
+                    {
+                        // TODO: ???
+                        throw new DirectoryNotFoundException(temp);
+                    }
+
+                    _tempDirectory = temp;
                 }
 
-                return temp;
+
```
mikaelm12: Remove newline

src/Microsoft.AspNet.WebUtilities/FileBufferingReadStream.cs
```diff
@@ -88,6 +101,13 @@ public override long Seek(long offset, SeekOrigin origin)
 
         private Stream CreateTempFile()
         {
+            if (_tempFileDirectory == null)
```
mikaelm12: Will this ever evaluate to true with the [NotNull] attribute in the ctor on tempFileDirectory?

rynowak: Yes, I added a new constructor to this class which takes a `Func<string>`, in that case `_tempFileDirectory` will be null.

src/Microsoft.AspNet.Http/BufferingHelper.cs
```diff
@@ -12,21 +12,31 @@ public static class BufferingHelper
...
                 {
-                    // TODO: ???
-                    throw new DirectoryNotFoundException(temp);
+                    // Look for folders in the following order.
+                    var temp = Environment.GetEnvironmentVariable("ASPNET_TEMP") ??     // ASPNET_TEMP - User set temporary location.
+                               Path.GetTempPath();                                      // Fall back.
+
+                    if (!Directory.Exists(temp))
+                    {
+                        // TODO: ???
```
dougbu: Not blocking this PRT at all. But is there an issue corresponding to this TODO?

Issue #367, opened 2015-08-12
Should HttpContext implement IFeatureCollection?


davidfowl:
Why doesn't it? It would allow users to enumerate features as well. Although Revision would be kinda weird to expose. Some of the properties could be implicit implementation

Tratcher:
I think a direct implementation would be awkward, like the indexer that takes a type and returns an object.  Features are an advanced topic that most users won't ever deal with. Why not just add a `IFeatureCollection Features { get; }` property? And move Get/SetFeature<T> to extensions on IFeatureCollection.

davidfowl:
Sounds better.

Tratcher:
Dropping out of backlog because it involves breaking changes on the abstract base class.

Tratcher:
This keeps coming up as a useful way to create the structures for calling TestServer.Invoke directly.

davidfowl:
Ah to avoid going through the HttpClient?

Tratcher:
Yes. HttpClient is convenient but it can only handle simple scenarios. Katana had a separate API that let you set up the server data structures directly.

PR #383, opened 2015-08-30
Add HttpContext.Features, move Get/SetFeature.

Tratcher:
#367
@davidfowl @muratg @pranavkm

### DIFF
--- a/src/Microsoft.AspNet.Http.Features/FeatureCollectionExtensions.cs
+++ b/src/Microsoft.AspNet.Http.Features/FeatureCollectionExtensions.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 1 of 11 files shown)


davidfowl:
I'm not sure if it's worth it to make the change in 2 passes? Don't delete the old methods in the first pass and just update the call sites. This will make the ci happy 😀

davidfowl:
Besides that the change looks good :shipit:

src/Microsoft.AspNet.Http.Features/FeatureCollectionExtensions.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Http.Features
+{
+    public static class FeatureCollectionExtensions
+    {
+        public static T Get<T>(this IFeatureCollection features)
```
pranavkm: `TFeature` + documentation

PR #389, opened 2015-09-07
StringValues Improvements

justinvp:
The first commit adds more `StringValues` tests.

The second commit includes the following improvements:

1. Add public struct `Enumerator` and public `GetEnumerator` method. This avoids enumerator allocations when `foreach`ing over `StringValues`.

2. Implement `IReadOnlyList<string>`, because why not? It's a subset of `IList<T>`, which `StringValues` already implements, and aligns with other collections that implement both `IList<T>` and `IReadOnlyList<T>`, like `T[]` and `List<T>`.

3. Rename the indexer's parameter from `key` to `index` to be consistent with the naming used on other (all?) collections, like `T[]` and `List<T>`.

4. More efficient `IList<T>.IndexOf`: it was previously implemented using `foreach` over itself, which was resulting in an enumerator allocation. Instead, it's more performant to loop over the `_values` array directly, when `_values` isn't null, avoiding the enumerator overhead altogether.

5. More efficient `ICollection<T>.Contains`: it was previously casting itself to `IList<string>` to call `IList<string>.IndexOf`, which results in a box allocation. Now, `IndexOf` has been separated out into a private method that `Contains` calls directly, without boxing.

6. More efficient `ICollection<T>.CopyTo`: similar to the `IndexOf` improvement, it was previously using `foreach` over itself, resulting in an enumerator allocation. Now it makes use of the faster `Array.CopyTo` when `_values` isn't null.

7. More efficient `Concat`: similar to `IndexOf` and `CopyTo` this was using `foreach` over two `StringValues`, resulting in two enumerator allocations. Now it uses the more efficient `CopyTo` to copy the values directly into the new array.


src/Microsoft.Framework.Primitives/StringValues.cs
```diff
@@ -114,28 +115,64 @@ private string[] GetArrayValue()
...
+            if (_value != null)
             {
-                array[arrayIndex + i] = this[i];
+                if (arrayIndex < 0)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(arrayIndex));
+                }
+                if (array.Length - arrayIndex < 1)
+                {
+                    throw new ArgumentException(
```
davidfowl: Pass the name of the argument? Also why not just let the underlying operation throw instead of throwing your own exception?

justinvp: This just matches the behavior of Array.Copy above. The param name isn't specified in this case (as with Array.Copy) because it's checking two parameters (array.Length and arrayIndex) so it doesn't know which one is at fault.

src/Microsoft.Framework.Primitives/StringValues.cs
```diff
@@ -114,28 +115,64 @@ private string[] GetArrayValue()
...
-            foreach (var value in this)
+            return IndexOf(item);
+        }
+
+        private int IndexOf(string item)
+        {
+            if (_values != null)
             {
-                if (string.Equals(value, item, StringComparison.Ordinal))
+                string[] values = _values;
```
muratg: nit: `var`

src/Microsoft.Framework.Primitives/StringValues.cs
muratg: I agree, this looks like a more descriptive message.  Though I agree that you should pass the nameof the argument instead of hardcoding it in the string.

src/Microsoft.Framework.Primitives/StringValues.cs
```diff
@@ -218,16 +246,65 @@ public static StringValues Concat(StringValues values1, StringValues values2)
...
-                combined[index++] = value;
+                _values = values;
+                _current = null;
+                _index = 0;
             }
-            foreach (var value in values2)
+
+            public bool MoveNext()
+            {
+                string[] values = _values._values;
```
muratg: nit: `var`

src/Microsoft.Framework.Primitives/StringValues.cs
```diff
@@ -218,16 +246,65 @@ public static StringValues Concat(StringValues values1, StringValues values2)
...
+                        _current = values[_index];
+                        _index++;
+                        return true;
+                    }
+
+                    _current = null;
+                    return false;
+                }
+
+                string value = _values._value;
```
muratg: nit: `var`

src/Microsoft.Framework.Primitives/StringValues.cs
```diff
@@ -4,13 +4,14 @@
...
 using System.Collections.Generic;
+using Microsoft.Framework.Internal;
 
 namespace Microsoft.Framework.Primitives
 {
     /// <summary>
     /// Represents zero/null, one, or many strings in an efficient way.
     /// </summary>
-    public struct StringValues : IList<string>
+    public struct StringValues : IList<string>, IReadOnlyList<string>
```
pranavkm: Could this just be an `IReadOnlyList<>`? We'd lose `ICollection.CopyTo` when we make the change - is there anything else that's interesting that might be affected with that change?

Tratcher: No, we want to maintain interop with `IList<string>` because that's what string[] implements. It avoids more conversions. Many more APIs take IList than IReadOnlyList, including most of Linq.

src/Microsoft.Framework.Primitives/StringValues.cs
```diff
@@ -114,28 +115,64 @@ private string[] GetArrayValue()
...
+            return IndexOf(item) >= 0;
         }
 
         void ICollection<string>.CopyTo(string[] array, int arrayIndex)
         {
-            for(int i = 0; i < Count; i++)
+            CopyTo(array, arrayIndex);
+        }
+
+        private void CopyTo([NotNull] string[] array, int arrayIndex)
```
Tratcher: Don't use NotNull here. We're updating our guidelines for this soon.

src/Microsoft.Framework.Primitives/StringValues.cs
Tratcher: why make a local at all? It's not going to change on you.

src/Microsoft.Framework.Primitives/StringValues.cs
```diff
@@ -218,16 +246,65 @@ public static StringValues Concat(StringValues values1, StringValues values2)
...
 
             var combined = new string[count1 + count2];
-            var index = 0;
-            foreach (var value in values1)
+            values1.CopyTo(combined, 0);
+            values2.CopyTo(combined, count1);
+            return new StringValues(combined);
+        }
+
+        public struct Enumerator : IEnumerator<string>
```
Tratcher: private?

justinvp: The whole point of the struct enumerator is to avoid an enumerator when `foreach` over `StringValues` directly. As such, it (along with the public `GetEnumerator` method that returns it) cannot be private, otherwise it would result in a box allocation.

src/Microsoft.Framework.Primitives/StringValues.cs
justinvp: FWIW, the string didn't have any hardcoded argument names in it, it just described the two arguments without referring to them by name. This was [the same exact string](https://github.com/dotnet/coreclr/blob/4df67c74b4f35ce32f6afb41ffd813f28d8aed27/src/mscorlib/src/mscorlib.txt#L204) used by `Array.Copy`. However, given your comments, I went ahead and modified it here to include the argument names. Now, developers will get a slightly better message when hitting this case with a `StringValues` of length one.

justinvp:
I addressed the feedback. Not sure what's going on with the Travis CI build, looks like its failing for unrelated reasons. Any way to force it to try another build?

src/Microsoft.Framework.Primitives/StringValues.cs
```diff
@@ -4,13 +4,14 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using Microsoft.Framework.Internal;
```
Tratcher: nit: I don't think you need this using anymore.

src/Microsoft.Framework.Primitives/StringValues.cs
justinvp: The JIT is likely to produce better code when its cached in a local. See comment by jkotas: https://github.com/dotnet/coreclr/pull/1231#discussion_r34603505

test/Microsoft.Framework.Primitives.Tests/StringValuesTests.cs
```diff
@@ -23,51 +90,34 @@ public void IsReadOnly_True()
...
-        public void Constructor_NullStringArray_ExpectedValues()
+        [Theory]
+        [MemberData(nameof(DefaultOrNullStringValues))]
+        [MemberData(nameof(EmptyStringValues))]
+        public void DefaultNullOrEmpty_ExpectedValues(StringValues stringValues)
         {
-            var stringValues = new StringValues((string[])null);
             Assert.Equal(0, stringValues.Count);
             Assert.Null((string)stringValues);
+            Assert.Equal((string)null, stringValues);
```
Tratcher: Won't StringValues(string.Empty) implicitly convert to string.Empty here? How is null == string.Empty?

justinvp: It's not testing `string.Empty` here.

Tratcher: Ah, my bad, it's StringValues.Empty.

davidfowl:
@Tratcher :shipit:

src/Microsoft.Framework.Primitives/StringValues.cs
muratg: Cool, thanks!

Tratcher:
Rebased and merged.

Issue #366, opened 2015-08-07
Form reader should rewind buffer after parsing


Tratcher:
Reading the request body as a for fully buffers the body (https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http/Features/FormFeature.cs#L88), but does not rewind after parsing.

After parsing the form, rewind the buffer to the beginning so that if someone else can read the body directly without knowing if they can/should rewind. Here? https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http/Features/FormFeature.cs#L150

Otherwise all of the callers need to remember do this:
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs#L269

PR #392, opened 2015-09-08
Rewind the request buffer after parsing the form.

Tratcher:
#366 
@muratg @BrennanConroy

### DIFF
--- a/src/Microsoft.AspNet.Http/Features/FormFeature.cs
+++ b/src/Microsoft.AspNet.Http/Features/FormFeature.cs
@@ -152,6 +152,9 @@ namespace Microsoft.AspNet.Http.Features.Internal
                 }
             }
 
+            // Rewind so later readers don't have to.
+            _request.Body.Seek(0, SeekOrigin.Begin);
+
             Form = new FormCollection(formFields, files);
             return Form;
         }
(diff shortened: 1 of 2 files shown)


muratg:
:shipit:

BrennanConroy:
:shipit:

davidfowl:
Why Travis fail

src/Microsoft.AspNet.Http/Features/FormFeature.cs
```diff
@@ -152,6 +152,9 @@ public IFormCollection ReadForm()
                 }
             }
 
+            // Rewind so later readers don't have to.
+            _request.Body.Seek(0, SeekOrigin.Begin);
```
kichalla: curious should we check if the request body is seekable?...like one could disable request buffering with `IHttpBufferingFeature`

Tratcher: The request body was made seekable on line 89 above.

Tratcher:
@davidfowl Restore hung.

src/Microsoft.AspNet.Http/Features/FormFeature.cs
kichalla: ah ok, thanks!

Issue #360, opened 2015-08-04
Add Response.CanClear and Response.Clear()


lodejard:
```
bool HttpResponse.CanClear { get{
  true if Response.Body.CanSeek && !ResponseHeadersSent
}}
void HttpResponse.Clear() {
  reset body and headers.
}
```

Tratcher:
I think they'd go into Microsoft.AspNet.Http.Extensions.

Tratcher:
```
                context.Response.StatusCode = 200;
                context.Response.Headers.Clear();
                context.Response.Body.SetLength(0);
```

JonTheMon:
When using Clear(), should we be expected to do a separate check to ensure the body has been emptied/disposed?   Namely, when getting an exception, the DeveloperExceptionPage calls Clear() but returns Html already in the stream, then the exception page.

Tratcher:
@JonTheMon what component is buffering the body? The servers don't do that by default. What kind of buffer stream is in use? Clear calls SetLength(0), which may only work for MemoryStream.

JonTheMon:
The stream is the HttpContext.Response.Body, or apparently a System.IO.Stream of type VisualStudio.Web.BrowserLink.Runtime.ScriptInjectionFilterStream.  Instead of the SetLength(0) that was mentioned, I've taken to calling Body.Dispose().

Tratcher:
Sounds like a browserlink bug. UseBrowserLink is after UseDeveloperExceptionPage in your Startup, correct? It should have a try/finally in their middleware that removes their stream if exceptions are thrown. Something like this: https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionMiddleware.cs#L89-L91
Dispose is not recommended.
I'll followup with Browerlink.

JonTheMon:
UseBrowserLink is actually before UseDeveloperExceptionPage.  However, removing UseBrowserLink seems to not show the Developer Exception page anymore.

Tratcher:
Ah, move UseBrowserLink after UseDeveloperExceptionPage to see if that makes a difference.
Note you don't see the exception page when browserlink is removed because no other components are buffering, so exceptions part way through the response cannot be rendered.
https://github.com/aspnet/Templates/blob/dev/src/BaseTemplates/StarterWeb/Startup.cs#L42-L43

JonTheMon:
[...]

PR #394, opened 2015-09-08
Response.Clear().

Tratcher:
#360 
@lodejard @muratg @davidfowl

### DIFF
--- a/src/Microsoft.AspNet.Http.Extensions/ResponseExtensions.cs
+++ b/src/Microsoft.AspNet.Http.Extensions/ResponseExtensions.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Http.Features;
+
+namespace Microsoft.AspNet.Http.Extensions
+{
+    public static class ResponseExtensions
+    {
+        public static void Clear(this HttpResponse response)
+        {
+            if (response.HasStarted)
+            {
+                throw new InvalidOperationException("The response cannot be cleared, it has already started sending.");
+            }
+            response.StatusCode = 200;
+            response.HttpContext.Features.Get<IHttpResponseFeature>().ReasonPhrase = null;
+            response.Headers.Clear();
+            if (response.Body.CanSeek)
+            {
+                response.Body.SetLength(0);
+            }
+        }
+    }
+}
--- a/test/Microsoft.AspNet.Http.Extensions.Tests/ResponseExtensionTests.cs
+++ b/test/Microsoft.AspNet.Http.Extensions.Tests/ResponseExtensionTests.cs
@@ -1,0 +1,64 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Http.Features;
+using Microsoft.AspNet.Http.Internal;
+using Microsoft.Framework.Primitives;
+using Xunit;
+
+namespace Microsoft.AspNet.Http.Extensions
+{
+    public class ResponseExtensionTests
+    {
+        [Fact]
+        public void Clear_ResetsResponse()
+        {
+            var context = new DefaultHttpContext();
+            context.Response.StatusCode = 201;
+            context.Response.Headers["custom"] = "value";
+            context.Response.Body.Write(new byte[100], 0, 100);
+
+            context.Response.Clear();
+
+            Assert.Equal(200, context.Response.StatusCode);
+            Assert.Equal(string.Empty, context.Response.Headers["custom"].ToString());
+            Assert.Equal(0, context.Response.Body.Length);
+        }
+
+        [Fact]
+        public void Clear_AlreadyStarted_Throws()
+        {
+            var context = new DefaultHttpContext();
+            context.Features.Set<IHttpResponseFeature>(new StartedResponseFeature());
+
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNet.Http.Extensions/ResponseExtensions.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Http.Features;
+
+namespace Microsoft.AspNet.Http.Extensions
+{
+    public static class ResponseExtensions
```
davidfowl: Should this be in HttpAbstractions? It feels kinda core

src/Microsoft.AspNet.Http.Extensions/ResponseExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public static bool CanClear(this HttpResponse response)
+        {
+            return !response.HasStarted && response.Body.CanSeek;
+        }
+
+        public static void Clear(this HttpResponse response)
+        {
+            response.StatusCode = 200;
```
davidfowl: Will this throw if CanClear returns false? Should we throw?

src/Microsoft.AspNet.Http.Extensions/ResponseExtensions.cs
Tratcher: No, it's all sugar. About the only place it will get used is here:
https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/ErrorHandlerMiddleware.cs#L61
https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/ErrorPageMiddleware.cs#L75

src/Microsoft.AspNet.Http.Extensions/ResponseExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+            response.HttpContext.Features.Get<IHttpResponseFeature>().ReasonPhrase = null;
+            response.Headers.Clear();
+            response.Body.SetLength(0);
```
Tratcher: Note to self, check CanSeek.

src/Microsoft.AspNet.Http.Extensions/ResponseExtensions.cs
Tratcher: All of these APIs should already throw. The only reason to throw ourselves is if we want a more specific error message.

src/Microsoft.AspNet.Http.Extensions/ResponseExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using Microsoft.AspNet.Http.Features;
+
+namespace Microsoft.AspNet.Http.Extensions
+{
+    public static class ResponseExtensions
+    {
+        public static bool CanClear(this HttpResponse response)
+        {
+            return !response.HasStarted && response.Body.CanSeek;
```
Tratcher: Note to self, move CanSeek into Clear.

Tratcher:
Rebased and updated. I dropped CanClear because it ended up just being !HasStarted.

src/Microsoft.AspNet.Http.Extensions/ResponseExtensions.cs
davidfowl: Those links both 404 :smile:

davidfowl: I think we should throw a custom exception if we know we're in an invalid state.

src/Microsoft.AspNet.Http.Extensions/ResponseExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.AspNet.Http.Features;
+
+namespace Microsoft.AspNet.Http.Extensions
+{
+    public static class ResponseExtensions
+    {
+        public static void Clear(this HttpResponse response)
+        {
+            response.StatusCode = 200;
+            response.HttpContext.Features.Get<IHttpResponseFeature>().ReasonPhrase = null;
```
davidfowl: We don't have ReasonPhrase on the HttpResponse? Isn't that weird?

Tratcher: No, we consider reason phrase obsolete/useless. 99% of users just set it to the default value for the given status code. It has also been removed in HTTP/2.

src/Microsoft.AspNet.Http.Extensions/ResponseExtensions.cs
Tratcher: https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/ExceptionHandler/ExceptionHandlerMiddleware.cs#L61-L69
https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs#L75-L82

Tratcher:
Updated.

test/Microsoft.AspNet.Http.Extensions.Tests/ResponseExtensionTests.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using Microsoft.AspNet.Http.Internal;
+using Xunit;
+
+namespace Microsoft.AspNet.Http.Extensions
+{
+    public class ResponseExtensionTests
+    {
+        [Fact]
+        public void Clear_ResetsResponse()
```
davidfowl: Add a test for the clear throwing.

davidfowl:
:shipit: after CI passes and after testing the exceptional case.

PR #396, opened 2015-09-11
Add IHtmlContentBuilder interface

rynowak:
Common interface for things that allow appending content (TagBuilder,
BufferedHtmlContent). We want to be able to expose this from APIs for
users to add content.

See discussion: https://github.com/aspnet/Mvc/issues/3087


pranavkm:
:shipit:

PR #403, opened 2015-09-14
Add AppendEncoded to the builder

rynowak:
This is needed because a builder may have an optimized path for an
unencoded string. There's also no 'common' encoded string implementation
so it's much more straightforward to put this on the interface.


dougbu:
:shipit:

Issue #402, opened 2015-09-14
[Perf] High Allocator DefaultHttpContext


benaadams:
For 2000 plaintext requests DefaultHttpContext allocates 12,078 objects:

![image](http://aoa.blob.core.windows.net/aspnet/dhttp_alloc_objects.png)

Clocking in at 927,624 bytes:

![image](http://aoa.blob.core.windows.net/aspnet/dhttp_alloc_bytes.png)

Which would mean if you were performing 1M rps the GC would have to flush 6,039,000 objects and 464 Mbytes per second, 927 Mbytes per second for 2M rps and 3.2GBytes (42.3M objects) per second for 7M rps (~10GbE saturation) or 192 GBytes per minute.

rynowak:
@benaadams - images aren't loading

benaadams:
@rynowak better? (Have a PR for this)

rynowak:
yeah images are showing up now.

benaadams:
Have PR that reduces this to 136 objects

![allocs](http://aoa.blob.core.windows.net/aspnet/dhttp_alloc_after.png)

And 20434 bytes total for the lifetime of the app

![allocs](http://aoa.blob.core.windows.net/aspnet/dhttp_alloc_bytes_after.png)

benaadams:
Using the Test aspnet site: https://github.com/benaadams/IllyriadGames.Server.Web.River.Test on the server https://www.nuget.org/packages/IllyriadGames.Server.Web.River

Before this change and https://github.com/aspnet/DependencyInjection/pull/289 running a 10 second test allows 311,043 rps:

![10 sec run](http://aoa.blob.core.windows.net/aspnet/before_10s.png)

However the GC load is very high so running the test for 3 minutes causes the network to periodically flatline as it goes 100% on GC; which means it can only do 61,468 rps with very high latencies:

![3 min run](http://aoa.blob.core.windows.net/aspnet/before_3m.png)

With these two (three PRs) changes running the same server for 3 minutes is very consistent network performance and is performing 1,422,499 rps with much much lower latency:

![3 min after](http://aoa.blob.core.windows.net/aspnet/after_3m.png)

For reference the [RIO server](https://github.com/aspnet/benchmarks/pull/12) on this hardware+network is performing 1,991,008 rps

![3 min rio](http://aoa.blob.core.windows.net/aspnet/rio_3m.png)

And this is kinda the addition of the aspnet to that server (still not parsing headers :smile:

[...]

PR #404, opened 2015-09-15
Allow DefaultHttpContext to be pooled [Design]

benaadams:
Half of  #402 needs change to Hosting also


src/Microsoft.AspNet.Http/DefaultHttpContext.cs
```diff
@@ -168,8 +185,26 @@ public override void Abort()
 
         public override void Dispose()
         {
-            // REVIEW: is this necessary? is the environment "owned" by the context?
```
benaadams: Question has moved to -> https://github.com/aspnet/HttpAbstractions/pull/404/files#diff-ad404d59bc5860d71b7403b8e0a665b2R102

src/Microsoft.AspNet.Http.Features/FeatureReference.cs
```diff
@@ -34,5 +34,11 @@ public T Update(IFeatureCollection features, T feature)
             _revision = features.Revision;
             return feature;
         }
+
+        public void Reset()
```
davidfowl: Do you really need to hold onto FeatureReference structs?

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
```diff
@@ -13,14 +13,19 @@
 
 namespace Microsoft.AspNet.Http.Authentication.Internal
 {
-    public class DefaultAuthenticationManager : AuthenticationManager
+    public class DefaultAuthenticationManager : AuthenticationManager, IDisposable
```
davidfowl: I'm not a fan of using IDisposable to release memory for the pooled object, feels like mismatch...

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
```diff
@@ -109,5 +114,13 @@ public override IEnumerable<AuthenticationDescription> GetAuthenticationSchemes(
                 throw new InvalidOperationException($"The following authentication scheme was not accepted: {authenticationScheme}");
             }
         }
+
+        public void Dispose()
```
davidfowl: Reset

src/Microsoft.AspNet.Http/DefaultConnectionInfo.cs
```diff
@@ -72,5 +78,13 @@ public override Task<X509Certificate2> GetClientCertificateAsync(CancellationTok
         {
             return TlsConnectionFeature.GetClientCertificateAsync(cancellationToken);
         }
+
+        public void Dispose()
```
davidfowl: Reset

src/Microsoft.AspNet.Http/DefaultHttpContext.cs
```diff
@@ -51,6 +58,16 @@ public DefaultHttpContext(IFeatureCollection features)
             _session = FeatureReference<ISessionFeature>.Default;
         }
 
+        public void Reinitalize(IFeatureCollection features)
```
davidfowl: Initialize

src/Microsoft.AspNet.Http/DefaultHttpContext.cs
```diff
@@ -168,8 +185,26 @@ public override void Abort()
 
         public override void Dispose()
```
davidfowl: Reset

benaadams:
See https://github.com/aspnet/HttpAbstractions/issues/402#issuecomment-140336990 for details on server run improvements (61.5k rps -> 1.49M rps)

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
benaadams: Perhaps IReusable? With .Clear() method?

src/Microsoft.AspNet.Http.Features/FeatureReference.cs
benaadams: Need to check; they are heap allocated - so would be another chunk of memory but they are only a pointer and an int...

src/Microsoft.AspNet.Http/DefaultHttpResponse.cs
```diff
@@ -116,5 +121,14 @@ public override void Redirect(string location, bool permanent)
 
             Headers[HeaderNames.Location] = location;
         }
+
+        public void Dispose()
+        {
```
benaadams: Reset

src/Microsoft.AspNet.Http/DefaultHttpContext.cs
benaadams: Only one that doesn't fit as is already Disposable?

davidfowl:
Closing this as discussed with @benaadams

PR #405, opened 2015-09-15
Extension methods for IHtmlContentBuilder

rynowak:
Resolves a bunch of duplication between TagHelperContent and
BufferedHtmlContent, plus adds a few more convenience overloads we don't
have in one of the two places.


NTaylorMullen:
:shipit:

Issue #368, opened 2015-08-12
KeyValueAccumulator performance improvements


davidfowl:
- It allocates 2 dictionaries
- It should probably be a struct

davidfowl:
Even better would be to make the KeyValueAccumulator a call back model so there was 0 allocation (besides the delegate :smile: )

muratg:
If you have any proof of large perf issues with this, bring it back

benaadams:
@muratg this was completed in https://github.com/aspnet/HttpAbstractions/pull/411#issuecomment-153189423

PR #411, opened 2015-09-20
Less alloc/wrapping/boxing for Query, Forms, Cookies

benaadams:
* Lower allocation for empty dictionaries of types ```HeaderDictionary```, ```QueryCollection```, ```FormCollection```, ```RequestCookies```
* No allocation for empty Form parsing or Query parsing
* Read/Write request data for middleware editing/processing (e.g. 3rd party auth that removes its data; or data transforms along the pipeline)
* Removed ```IReadableStringCollection```, ```ReadableStringCollection``` -> ```IQueryCollection```, ```IRequestCookies```,  ```IFormCollection``` each implementing ```IDictionary<string, StringValues>``` 
* Added ```IQueryCollection```, ```IRequestCookies```,  ```IFormCollection``` doesn't throw on get contract, non-boxing enumerator in Empty paths.
* ```KeyValueAccumulator``` from ```class``` to ```struct```
* Removed a bunch of boxing paths
* Reduced allocs in some paths (e.g. [```ToUriComponent```](https://github.com/aspnet/HttpAbstractions/pull/411/files#diff-acd9903bfe4362b3b8480b33de592c20L66) or [```ParseQuery```](https://github.com/aspnet/HttpAbstractions/pull/411/files#diff-cb271f343ac8ee259123a6e42be32e89L114))
* Changes to support non-boxing ```Equality``` for ```StringValues``` (see PR https://github.com/aspnet/Common/pull/55)

Also cleaned up and added some intellisense code comments; but they still need work, so this is not "done"


benaadams:
May need changes to Kestrel's custom IDictionaries dealing with empty keys
Also Content-Length bonding to header

src/Microsoft.AspNet.Http/Features/HttpRequestFeature.cs
```diff
@@ -12,7 +12,7 @@ public class HttpRequestFeature : IHttpRequestFeature
     {
         public HttpRequestFeature()
```
Tratcher: FYI: This implementation is used exclusively in unit tests, no optimizations required. We should add a doc comment to that effect.

src/Microsoft.AspNet.Http/Features/HttpResponseFeature.cs
```diff
@@ -6,23 +6,51 @@
 using System.IO;
 using System.Threading.Tasks;
 using Microsoft.Framework.Primitives;
+using Microsoft.AspNet.Http.Internal;
+using Microsoft.AspNet.WebUtilities;
 
 namespace Microsoft.AspNet.Http.Features.Internal
 {
     public class HttpResponseFeature : IHttpResponseFeature
```
Tratcher: Same for this one, just for unit tests.

src/Microsoft.AspNet.Http/Features/FormFeature.cs
```diff
@@ -67,16 +68,6 @@ public bool HasFormContentType
 
         public IFormCollection ReadForm()
         {
-            if (Form != null)
```
davidfowl: What happened to this code?

benaadams: It then calls ```ReadFormAsync``` which repeats these checks; no point in doing them twice?

davidfowl: It avoids the state machine and task allocation.

src/Microsoft.AspNet.Http/Features/HttpResponseFeature.cs
davidfowl: Yes I would undo the changes in these implementations.

src/Microsoft.AspNet.Http/FormCollection.cs
```diff
@@ -10,29 +12,28 @@ namespace Microsoft.AspNet.Http.Internal
...
-            }
-
-            Files = files;
+            // can be null
+            Store = store;
+            _files = files;
         }
 
-        public IFormFileCollection Files { get; }
+        public IFormFileCollection Files {
```
davidfowl: The formatter here looks messed up.

src/Microsoft.AspNet.Http/HeaderDictionary.cs
```diff
@@ -1,199 +0,0 @@
...
-        }
-
-        /// <summary>
-        /// Get or sets the associated value from the collection as a single string.
-        /// </summary>
-        /// <param name="key">The header name.</param>
-        /// <returns>the associated value from the collection as a StringValues or StringValues.Empty if the key is not present.</returns>
-        public StringValues this[string key]
-        {
-            get { return ParsingHelpers.GetHeader(Store, key); }
```
davidfowl: Where did this logic go? Seems like none of the tests failed. Which either means we have no coverage or this is equivalent to looking up the key in the dictionary.

src/Microsoft.AspNet.Http/ParsingHelpers.cs
```diff
@@ -549,16 +550,11 @@ public static void SetHeader(IDictionary<string, StringValues> headers, string k
             {
                 throw new ArgumentNullException(nameof(headers));
             }
-
-            if (key == null)
```
davidfowl: Why did this check get removed?

src/Microsoft.AspNet.Http/ParsingHelpers.cs
```diff
@@ -592,8 +588,10 @@ private static string DeQuote(string value)
 
             const NumberStyles styles = NumberStyles.AllowLeadingWhite | NumberStyles.AllowTrailingWhite;
             long value;
-            var rawValue = headers[HeaderNames.ContentLength];
```
davidfowl: This should not throw. That's one of the reasons we introduced IReadableStringCollection in the first place. Indexers throwing for missing entries is annoying as hell.

src/Microsoft.AspNet.Http/DefaultHttpResponse.cs
```diff
@@ -54,21 +56,20 @@ public override Stream Body
...
         {
-            get
-            {
-                return ParsingHelpers.GetContentLength(Headers);
-            }
-            set
-            {
-                ParsingHelpers.SetContentLength(Headers, value);
-            }
+            get { return HttpResponseFeature.ContentLength; }
```
davidfowl: Isn't this broken? Don't you still have to support setting the header? Can this and the header get out of sync?

src/Microsoft.AspNet.Http/DefaultHttpResponse.cs
```diff
@@ -54,21 +56,20 @@ public override Stream Body
...
+            set { HttpResponseFeature.ContentLength = value; }
         }
 
         public override string ContentType
         {
             get
             {
-                return Headers[HeaderNames.ContentType];
```
davidfowl: This shouldn't throw. Revert this change.

src/Microsoft.AspNet.Http/Features/HttpResponseFeature.cs
davidfowl: Yep revert these changes.

src/Microsoft.AspNet.Http/ParsingHelpers.cs
davidfowl: Why would we support null keys?

src/Microsoft.AspNet.Http/ParsingHelpers.cs
```diff
@@ -592,8 +588,10 @@ private static string DeQuote(string value)
...
-            if (rawValue.Count == 1 &&
+
+            StringValues rawValue;
+            if (headers.TryGetValue(HeaderNames.ContentLength, out rawValue) &&
```
davidfowl: revert change

src/Microsoft.AspNet.Owin/OwinFeatureCollection.cs
```diff
@@ -14,13 +14,13 @@
...
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Features;
 using Microsoft.AspNet.Http.Features.Authentication;
 using Microsoft.Framework.Primitives;
 
 namespace Microsoft.AspNet.Owin
 {
+    using Net.Http.Headers;
```
davidfowl: Should be outside the namespace

src/Microsoft.AspNet.WebUtilities/LowAllocationDictionary.cs
```diff
@@ -0,0 +1,227 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.WebUtilities
```
davidfowl: This type should only be used internally. I'm not sure than we should be deriving from it. It's also a LazyDictionary not a LowAllocationDictionary. This should probably move to the Common repository if we take this change.

src/Microsoft.AspNet.Http.Features/IHttpResponseFeature.cs
```diff
@@ -13,6 +13,7 @@ public interface IHttpResponseFeature
     {
         int StatusCode { get; set; }
         string ReasonPhrase { get; set; }
+        long? ContentLength { get; set; }
```
davidfowl: Undo this change

src/Microsoft.AspNet.Http/ParsingHelpers.cs
benaadams: Is already covered by check below: ```string.IsNullOrWhiteSpace(key)```

davidfowl: yep just saw it

src/Microsoft.AspNet.Http/Features/FormFeature.cs
benaadams: True. Will put back in the form take out the exception throw.

Why support the non-async method, when you clearly don't? Can avoid state machine with a slightly different pattern for cached results; but put the choice and onus of "Sync over Async" into user code?

davidfowl: Unfortunate synchronous call sites in user code.

benaadams: Which can just as happily also call: ```ReadFormAsync().GetAwaiter().GetResult()``` or ```ReadFormAsync().Result()``` but it doesn't come with the incorrect assumption that the synchronous method is not [Sync over Async](http://blogs.msdn.com/b/pfxteam/archive/2012/04/13/10293638.aspx)?

davidfowl: I agree, we should have a proper sync version.

benaadams: I was going for kill the sync version... :stuck_out_tongue_closed_eyes:

src/Microsoft.AspNet.Http/HeaderDictionary.cs
benaadams: It moved; tests are testing it.

src/Microsoft.AspNet.Http/DefaultHttpResponse.cs
benaadams: No; not with the code here; however it does rely on the ```engine/server``` implementer of ```IHttpResponseFeature``` maintaining this behaviour. Middleware wrapping that implementation will still be safe.

src/Microsoft.AspNet.Http.Features/IHttpResponseFeature.cs
benaadams: Puts the onus on the IHttpResponseFeature to sync

benaadams: Takes the allocations to zero if the ContentLength is maintained as the source of the header and the header set/get is intercepted; rather than otherway around.

davidfowl: I get it but I don't think it belongs on this PR

src/Microsoft.AspNet.Http/DefaultHttpResponse.cs
```diff
@@ -41,9 +43,18 @@ public override int StatusCode
             set { HttpResponseFeature.StatusCode = value; }
         }
 
-        public override IHeaderDictionary Headers
+        public override IDictionary<string, StringValues> Headers
         {
-            get { return new HeaderDictionary(HttpResponseFeature.Headers); }
+            get
+            {
+                if (HttpResponseFeature.Headers is SafeDictionary<string, StringValues>)
```
benaadams: Looked at using a local struct wrapper; but the Header collection is RO here, but R/W through the feature; so without a notification on the feature that its been changed it has to be something like this; and SafeDictionary a class as it will get boxed so no saving.

Headers Dict can be made R/W here safely.

src/Microsoft.AspNet.Http/DefaultHttpRequest.cs
```diff
@@ -121,12 +123,21 @@ public override string Protocol
...
         }
 
-        public override IHeaderDictionary Headers
+        public override IDictionary<string, StringValues> Headers
         {
-            get { return new HeaderDictionary(HttpRequestFeature.Headers); }
+            get
+            {
+                if (HttpRequestFeature.Headers is SafeDictionary<string, StringValues>)
+                {
```
benaadams: Same as response headers

src/Microsoft.AspNet.Http/SafeDictionary.cs
```diff
@@ -0,0 +1,219 @@
...
+            Store = store;
+        }
+
+        public TValue this[TKey key]
+        {
+            get
+            {
+                if (Store == null)
+                {
+                    new NullReferenceException();
```
justinvp: `throw`? (applies throughout)

davidfowl: throw NullReferenceException :\ what

src/Microsoft.AspNet.Http/Features/FormFeature.cs
```diff
@@ -71,23 +72,32 @@ public IFormCollection ReadForm()
             {
                 return Form;
             }
-
-            if (!HasFormContentType)
```
davidfowl: Bring this back.

src/Microsoft.AspNet.Http/SafeDictionary.cs
khellang: That's already take care of below anyway :wink:

src/Microsoft.AspNet.Http/Features/HttpResponseFeature.cs
```diff
@@ -6,6 +6,8 @@
 using System.IO;
 using System.Threading.Tasks;
 using Microsoft.Framework.Primitives;
+using Microsoft.AspNet.Http.Internal;
+using Microsoft.AspNet.WebUtilities;
```
davidfowl: Undo?

src/Microsoft.AspNet.Http/FormCollection.cs
```diff
@@ -10,29 +10,31 @@ namespace Microsoft.AspNet.Http.Internal
     /// <summary>
     /// Contains the parsed form values.
     /// </summary>
-    public class FormCollection : ReadableStringCollection, IFormCollection
+    public class FormCollection : SafeLazyDictionary<StringValues>, IFormCollection
```
davidfowl: SafeLazyDictionary shouldn't be exposed in public API

Tratcher:
@benaadams What you're getting at with much of this is that our multiple levels of abstraction are inefiicent and we should scale back. That's legit, but I think we need to choose which collection we use carefully. We used Dictionaries internally because they were convenient storage, but as fowler pointed out we hate that their index throws, and they also have a significantly larger surface area than we need.

Rather than pushing IDictionary up to the surface APIs, let's try pushing a reduced IHeaderCollection down into the features/servers. we only really need `IEnumerable<KeyValuePair<string, StringValues>>`, Count, indexer, and `ICollection<Keys>`. @davidfowl, @lodejard, what do you think?

Tratcher:
The same goes for query, form, etc..

davidfowl:
@Tratcher we can't push the form and query etc down to the server but that's not what you mean I assume.

Tratcher:
@davidfowl Not down to the server, but down to their respective storage/features.

benaadams:
@Tratcher yes. However, after a large amount of changes, and some iterations to maintain a close analogue to flow and constraints, I mostly ended up in a similar place to where the code started; including wrapping on get for headers! 

There are a few tweaks but they really end up independent to the abstraction changes (LazyDict and its inherited collections, KeyValueAccumulator). Which was't where I expected to be; and would have likely been the exact same place if I wasn't trying to avoid it. :crying_cat_face: 

However, I did ponder along the journey...

*Questions*

* **Is wrapping the collections for no "throw on get" necessary?**  I completely agree with the reasons; but could this be a documented constraint for writing a [custom server](http://docs.asp.net/en/latest/fundamentals/servers.html#custom-servers) or implementing ```IHttpRequestFeature``` / ```IHttpResponseFeature```? For example Kestrel's ```FrameHeaders``` collection will not throw on get and that will probably be the default for the majority of people. So is this overly defensive?

  Equally the default collections for forms and query are again provided so can be wrapped at that point (as per this PR with FormCollection : SafeLazyDictionary<StringValues> ignoring the inheritance exposure issue)

* **Is the request data being read-only desirable?** With the request data being read-only this means middleware providers are more likely to resolve down to ```IHttpRequestFeature``` and wholesale replace the collection as they can't otherwise alter the data; making the defensive approach in the above question more necessary.

I had a few more, but I think they are artefacts of the above two and what implementation the answers encourage you down. Not necessarily looking for answers.

Tratcher:
IDictionary's contract is that the indexer throws, that's not something we can/should change in our stack as it would case a lot of unpleasant surprises. If we want a different semantic then we need a different contract / interface. I don't think we even need IDictionary, it was just the closest match to what wanted. Now that we're cleaning things up I think we can come up with a slimmed down type that has the semantics we want and use that in both the surface API (HttpRequest) and internally (IHttpRequestFeature).

The request data should be mutable. There are common scenarios where middleware will alter the request to meet downstream requirements like applying x-forwarded headers to their respective fields.

benaadams:
Different Contract makes sense; also with IgnoreCase on keys?

If you did something like this (inheritance-wise) everyone will want some:
```csharp
interface IReadOnly<TKey, TValue>
{
    TValue Item { get; }
}

interface IReadWrite<TKey, TValue> : IReadOnly<TKey, TValue>
{
    new TValue Item { get; set; }
}
```

It seems to work, though you have to implement the get twice:

```csharp
class Thing<TKey, TValue> : IReadWrite<TKey, TValue>
{
    public TValue Item
    {
        get
        {
            throw new NotImplementedException();
        }

        set
        {
            throw new NotImplementedException();
        }
    }

    TValue IReadOnly<TKey, TValue>.Item
    {
        get
        {
            return Item;
        }
    }
}
```

src/Microsoft.AspNet.Http/SafeDictionary.cs
benaadams: Its a pass through wrapper that's only purpose it to massage ```Get```. If Store is not set that's the equivalent of not having the wrapper and not setting the original collection to anything.

You probably want to know this error up front, and the behaviour maintained? Rather than the disconnect disappearing into the void and just producing non-nonsensical results.

Ideally you wouldn't need class at all for this particular use.

Tratcher:
That looks like a nice trick.

benaadams:
Copied some of this PR to  #413 (has no api changes)

src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryExtensions.cs
```diff
@@ -23,7 +24,7 @@ public static void Append(this IHeaderDictionary headers, string key, StringValu
         /// </summary>
         /// <param name="key">The header name.</param>
         /// <param name="values">The header values.</param>
-        public static void AppendCommaSeparatedValues(this IHeaderDictionary headers, string key, params string[] values)
+        public static void AppendCommaSeparatedValues(this IDictionary<string, StringValues> headers, string key, params string[] values)
```
lodejard: These CommaSeparatedValues methods are actually very specific to request/response header notation... That is, for the header `x-example: "detect, commas", "inside", values, "like this"` and turn that into the string array: 
* `detect, commas`
* `inside`
* `values`
* `like this`

That is not a behavior which is appropriate for cookies, query, or form values - so the extension method on the common interface should be avoided...

What if `public interface IHeaderDictionary : IDictionary<string, StringValues> {}` was added to the same assembly as `IHttpRequest/ResponseFeature`? Doesn't add extra implementation burden to the server (just need to add that interface to the dictionary impl), you can put extension methods on it specifically, and you don't need to allocate a wrapper between the http abstraction class and the feature interface

Tratcher: @lodejard  That's more or less what we were discussing here: https://github.com/aspnet/HttpAbstractions/pull/411#issuecomment-142014003  Except that I don't think we even need to keep IDictionary.

lodejard:
I like this change - moving to IDictionary<string, StringValues> seems like the right thing to do.

Also add an interface `IHeaderCollection : IDictionary<string, StringValues>` along with request/response feature interface, so that the header extension methods can target that specifically.

Also consider adding `IQueryCollection: IDictionary...` for query/cookie if we think we need to pin down the semantics that those collections should not throw on indexing. At a minimum that states you cannot have a dictionary returned from that property which does not implement that interface, so you know you can't be given a `new Dictionary<string, StringValues>(...)` that has the default indexing behavior

#### TL;DR

There were a few big reasons to have the custom collections in the first place, and they may no longer apply.

The big one was that the value needed to be string-or-string-array in a relatively convenient way. That's where having a value type of StringValues makes that reason apply much, much less.

Another is that indexer of a key that isn't present for web programming has traditionally returned an empty value rather than throwing, which is different than the definition of IDictionary's [] getter. That said,  I have heard at least one person say that returning default(T) from the [] getter isn't the end of the world, especially because code that does care specifically about the non-existence will be calling TryGet instead of `try { [ ] } catch { ... }` anyway.

Another was that some of the collections aren't directly alterable. If that's the only reason remaining, you can always throw not supported, so that alone might not be enough to justify the collection interfaces.

src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryExtensions.cs
benaadams: @Tratcher raised a good point about the contract of ```IDictionary``` being throw on ```Get``` so a different interface would likely be better that returns default(TValue) on ```Get```

Wouldn't be an extra burden on Kestrel's HeaderDict as you don't actually wrap a dictionary and are implementing the interface in this manner. Would also allow a IReadOnly->IReadWrite interface hierarchy should it be desired which is not possible with IDictionary and IReadOnlyDictionary (see https://github.com/aspnet/HttpAbstractions/pull/411#issuecomment-142065105)

lodejard: Yep, I'm looking at the same PR, I just added a comment that being IDictionary is probably a good thing :)

benaadams:
I've done a 180° on this one; or maybe a 90° :smile: 

* A different interface than ```IDictionary``` (probably across the board) forces the implementer to at least pause when applying a collection; else the default method would probably just to use a ```Dictionary``` which to cover the throw safely you need to wrap on read. Whereas a new interface can have that it doesn't throw on get expressed in its Intellisense description; which means the wrapping can be dropped. Also the description can say that the keys should ignore case.

* A lower alloc for empty size can be offered as a default dictionary type implementing the new interfaces as shown by the [SafeLazyDictionary](https://github.com/benaadams/HttpAbstractions/blob/lazy-dict/src/Microsoft.AspNet.WebUtilities/LazyDictionary.cs) as well as embedding ```OrdinalIgnoreCase``` into it. Though it would want a change to expose a strongly typed enumerator, and probably a better name.

* A proper IReadOnly->IReadWrite interface hierarchy can be baked into the new interfaces which cannot be done with ```IDictionary``` and ```IReadOnlyDictionary``` as there is no inheritance between them and ```as``` casting from an [unrealted type is slow](https://github.com/dotnet/coreclr/issues/1592).

Although since you'd be wrapping anyway to match the interface you could go for straight inheritance rather than Interfaces; then always have a strongly typed enumerator at all levels (this seems to work):
```csharp
public abstract class ReadOnlyHttpDictionary
{
    abstract StringValues this[string key] { get; }
}

public abstract class ReadWriteHttpDictionary : ReadOnlyHttpDictionary
{
    abstract StringValues this[string key] { get; set; }

    ReadOnlyHttpDictionary ReadOnlyView()
    {
        return (ReadOnlyHttpDictionary)this;
    }
}

public class HttpDictionary : ReadWriteHttpDictionary
{
    StringValues this[string key] { get; set; }
}
```
Anyway... the choice is yours!

lodejard:
> * A different interface than ```IDictionary``` (probably across the board) forces the implementer to at least pause when applying a collection; else the default method would probably just to use a ```Dictionary``` which to cover the throw safely you need to wrap on read. Whereas a new interface can have that it doesn't throw on get expressed in its Intellisense description; which means the wrapping can be dropped. Also the description can say that the keys should ignore case.

True, but the same thing happens with `IHeaderCollection : IDictionary<string, StringValues>`. You can't have any-old-IDictionary returned from or assigned to that location.

> * A lower alloc for empty size can be offered as a default dictionary type implementing the new interfaces as shown by the [SafeLazyDictionary](https://github.com/benaadams/HttpAbstractions/blob/lazy-dict/src/Microsoft.AspNet.WebUtilities/LazyDictionary.cs) as well as embedding ```OrdinalIgnoreCase``` into it. Though it would want a change to expose a strongly typed enumerator, and probably a better name.

Yep, though the same thing also applies to a specific interface that inherits the standard interface. For convenience we would probably have the instantiable types exposed like `public class HeaderCollection : SafeLazyDictionary, IHeaderCollection`. Those would all get the benefits of correctly casing keys, and creating the actual dictionary store on-demand. Those would all be in the base, the actual class that adds the specific interface doesn't do more than implement any addtl interface members (which are mostly none).

Also in practice on this one you probably wouldn't bother with an allocation-free enumeration on the implementation type because you would be enumerating without down-casting in most (all) cases... So you wouldn't hit that code-path at all.

> * A proper IReadOnly->IReadWrite interface hierarchy can be baked into the new interfaces which cannot be done with ```IDictionary``` and ```IReadOnlyDictionary``` as there is no inheritance between them and ```as``` casting from an [unrealted type is slow](https://github.com/dotnet/coreclr/issues/1592).

That's true, but the benefit of having a readonly/readwrite declaration has to be weighed against the overhead of declaring and using a new interface... Before the **real** tipping point for declaring a different collection interfaces altogether was string-vs-string[], but with StringValues that goes away. 

TBH, IMO, the fact this change lets us remove the interfaces with their own definition of [], Count, Keys, etc is something that makes it pretty compelling.

davidfowl:
@lodejard @Tratcher feels like we want to take this change (I do too!). Let's close on this and merge it in tomorrow.

benaadams:
Removed wrapping
Incorporated improvements made to #413 
Issues:
* Don't expose```HttpDictionary``` in public interface? @davidfowl 
* New interfaces?  @lodejard ```IHeaderCollection```,  ```IQueryCollection```
* CI errors (```Duplicate member name 'System.Collections'``` caused by adding ```"System.Collections"``` to ```project.json``` inside ```frameworks->dnxcore50->dependencies``` in ```Microsoft.AspNet.WebUtilities```)

Contentious?:
* New function: [FormReader.ReadNullableFormAsync](https://github.com/aspnet/HttpAbstractions/pull/411/files#diff-22312b73697acf841af7cd8e8e4d1f3cR215) allowing for no alloc in ```FormRead``` [HasApplicationFormContentType(contentType)](https://github.com/aspnet/HttpAbstractions/pull/411/files#diff-25040e9a807f3ea89b0c6dc24fa8d50dR136) path
* Strong typed return from [KeyValueAccumulator.GetResults](https://github.com/aspnet/HttpAbstractions/pull/411/files#diff-1294cbe52c3abf8f3a474a86ff6f0146R33) though can be implicitly cast to into ```IDictionary<string,StringValues>``` by called.
* Strong typed [FormCollection constructor](https://github.com/aspnet/HttpAbstractions/pull/411/files#diff-0fa6841cdb34e605cd1192af8a5ecbc9R26)

Tratcher:
@davidfowl @lodejard Let's thrash out the details face-to-face so we can give @benaadams clear guidelines. Also, I don't recommend trying to push this through today as it will require non-trivial changes to both servers and we have higher priority things to resolve for beta8.

benaadams:
> Also add an interface IHeaderCollection : IDictionary<string, StringValues> along with request/response feature interface, so that the header extension methods can target that specifically.

@lodejard Could move the Header/FormFile etc extensions into a specific namespace for the use; then their visibility is shown by choice of using directive?

benaadams:
Also now resolves #368 ?

Tratcher:
Thanks for the investigation. Here's what we decided to do for the collections: https://github.com/aspnet/HttpAbstractions/issues/426. I'll take care of this since it requires updates to both the servers. Let's pick up on the rest of these changes after I'm done.

benaadams:
Rebasing and bringing in recent changes

Tratcher:
@benaadams This is going the opposite direction from our intended fix described here: https://github.com/aspnet/HttpAbstractions/issues/426

benaadams:
It will go the same direction; but am testing my new found abilities to reuse/remerge a PR; rather than create lots of new ones :)

benaadams:
Bringing in line with #426 next

benaadams:
Initial Interfaces, needs some clean-up and method comments

src/Microsoft.AspNet.WebUtilities/HttpDictionary.cs
```diff
@@ -0,0 +1,414 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.WebUtilities
+{
+    /// <summary>
+    /// The base case-insenstive HttpDictionary type used for <see cref="T:Microsoft.AspNet.Http.Internal.HeaderDictionary" /> and <see cref="T:ResponseHeaders" />.
+    /// </summary>
+    public class HttpDictionary : IDictionary<string, StringValues>
```
davidfowl: This isn't a thing we want to expose.

benaadams: :ok:

benaadams: Removed it

src/Microsoft.AspNet.Http.Abstractions/PathString.cs
```diff
@@ -66,7 +67,24 @@ public override string ToString()
         public string ToUriComponent()
         {
             // TODO: Measure the cost of this escaping and consider optimizing.
-            return HasValue ? string.Join("/", _value.Split('/').Select(UrlEncoder.Default.UrlEncode)) : string.Empty;
+            if (!HasValue)
+            {
+                return string.Empty;
+            }
+            var values = _value.Split(splitChar);
```
davidfowl: Nice! This makes me feel like we should make a non allocating string split using https://github.com/aspnet/Common/blob/dev/src/Microsoft.Extensions.Primitives/StringSegment.cs

src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryExtensions.cs
```diff
@@ -36,7 +36,7 @@ public static void AppendCommaSeparatedValues(this IHeaderDictionary headers, st
         /// <returns>the associated values from the collection separated into individual values, or StringValues.Empty if the key is not present.</returns>
         public static string[] GetCommaSeparatedValues(this IHeaderDictionary headers, string key)
         {
-            return ParsingHelpers.GetHeaderSplit(headers, key);
+            return ParsingHelpers.GetHeaderSplit(headers, key).ToArray();
```
davidfowl: Why?

src/Microsoft.AspNet.Http/Features/FormFeature.cs
```diff
@@ -74,41 +82,56 @@ public IFormCollection ReadForm()
...
             _request.EnableRewind();
 
-            IDictionary<string, StringValues> formFields = null;
-            var files = new FormFileCollection();
+            FormCollection formFields = null;
+            FormFileCollection files = null;
 
             // Some of these code paths use StreamReader which does not support cancellation tokens.
-            using (cancellationToken.Register(_request.HttpContext.Abort))
+            using (cancellationToken.Register((state) => ((FormFeature)state)._request.HttpContext.Abort(), this))
```
davidfowl: This would be cleaner if you just passed in the http context directly.

src/Microsoft.AspNet.Http/Features/FormFeature.cs
```diff
@@ -74,41 +82,56 @@ public IFormCollection ReadForm()
...
             {
                 var contentType = ContentType;
                 // Check the content-type
                 if (HasApplicationFormContentType(contentType))
                 {
                     var encoding = FilterEncoding(contentType.Encoding);
-                    formFields = await FormReader.ReadFormAsync(_request.Body, encoding, cancellationToken);
+                    formFields = new FormCollection(await FormReader.ReadNullableFormAsync(_request.Body, encoding, cancellationToken), null);
```
davidfowl: use optional arg syntax for the null

src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryExtensions.cs
benaadams: Its representing the ugliness of a potential array allocation if the StringValue only has one value

benaadams: Though where I do ```.ToString()``` is generally to avoid boxing the param.

Have taken the approach that implicit conversion is nice for users; but in perf sensitive code is better to be explicit about what's actually happening so the costs are more clearly understood when reviewing. Sensible approach?

src/Microsoft.AspNet.Http.Extensions/Internal/HeaderSegmentCollection.cs
```diff
@@ -16,7 +19,7 @@ public HeaderSegmentCollection(StringValues headers)
 
         public bool Equals(HeaderSegmentCollection other)
         {
-            return Equals(_headers, other._headers);
+            return StringValues.Equals(_headers, other._headers);
         }
```
benaadams: Currently both versions box equally; however change https://github.com/aspnet/Common/pull/55 is to resolve this and for a non-boxing path.

benaadams:
Updated summary

src/Microsoft.AspNet.Http.Abstractions/PathString.cs
Tratcher: And Join while you're at it.

src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryTypeExtensions.cs
```diff
@@ -80,7 +80,24 @@ internal static void SetList<T>(this IHeaderDictionary headers, string name, ILi
             }
             else
             {
-                headers[name] = values.Select(value => value.ToString()).ToArray();
+
+                switch (values.Count)
+                {
+                    case 0:
+                        headers[name] = StringValues.Empty;
```
Tratcher: or `headers.Remove("Name")`

benaadams:
Have struct enumerator change that will integrate more easily with non-dictionary based stuff (e.g. Kestrel's FrameRequestHeaders) and still be struct-based; though stuff is getting weirder.

src/Microsoft.AspNet.Http.Extensions/Internal/StringSegment.cs
```diff
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
 
 namespace Microsoft.AspNet.Http.Internal
```
Tratcher: Wasn't this moved to Common?

benaadams:
Need to check if this does work for Kestrel's dicts

benaadams:
Last commit (and slightly one before) - Ugly giant struct enumerator, but is all stack based with no GC involvement. Meaning even more unusual implementations like Kestrel's FrameRequestHeaders and FrameResponseHeaders can still have a struct based enumerator when enumerated via the IHttpDictionary interface.

Associated PR for Kestrel https://github.com/aspnet/KestrelHttpServer/pull/267

benaadams:
Think I'm done :stuck_out_tongue_closed_eyes: *Subject to feeback*

benaadams:
Merged to single commit as my rebasing skills seem to still need work :crying_cat_face: 

However kept the previous commits incase the Intreface Enumerator structs are going too far: https://github.com/benaadams/HttpAbstractions/tree/old-dict so I can back them out individually.

AppVeyor is reporting really weird errors...

> Unhandled Exception: System.InvalidCastException: Unable to cast object of type 'System.Threading.Thread' to type 'System.Net.Http.WinHttpRequestState'.
   at System.Net.Http.WinHttpRequestState.FromIntPtr(IntPtr gcHandle)
   at System.Net.Http.WinHttpRequestCallback.WinHttpCallback(IntPtr handle, IntPtr context, UInt32 internetStatus, IntPtr statusInformation, UInt32 statusInformationLength) 
When sending 'PUT http://localhost:1030/api/tests' with body '{"TestName":"Microsoft.AspNet.Builder.Extensions.MapPathMiddlewareTests.NullArguments_ArgumentNullException","FileName":"Microsoft.AspNet.Http.Abstractions.Tests.dll","TestFramework":"xUnit","Outcome":"Passed","DurationMilliseconds":1,"ErrorMessage":null,"ErrorStackTrace":null,"StdOut":"","StdErr":null}', exception was thrown: A task was canceled.

benaadams:
AppVeyor error may be this https://github.com/dotnet/corefx/pull/4001 was resolved in corecfx 15 hrs ago so probably not in current build. Will run again later.

benaadams:
Reverted the struct enum stuff as it was getting weird and the collections are unlikely to enumerated by interface too much.

src/Microsoft.AspNet.Http.Extensions/Internal/StringSegment.cs
benaadams: Was there twice? https://github.com/aspnet/HttpAbstractions/commit/221df714258668603319e75828e9546eaa13926d

benaadams: Removed

src/Microsoft.AspNet.Http.Abstractions/QueryString.cs
```diff
@@ -148,7 +148,7 @@ public static QueryString Create(IEnumerable<KeyValuePair<string, string>> param
         /// </summary>
         /// <param name="parameters"></param>
         /// <returns>The resulting QueryString</returns>
-        public static QueryString Create(IEnumerable<KeyValuePair<string, StringValues>> parameters)
+        public static QueryString Create(IQueryCollection parameters)
```
Tratcher: No, this should be a general purpose factory method.

src/Microsoft.AspNet.Http.Features/IQueryCollection.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// Represents the HttpRequest query string collection
+    /// </summary>
+    public interface IQueryCollection : IDictionary<string, StringValues>
+    {
+        /// <summary>
+        /// IQueryCollection has a different indexer contract than IDictionary, where it will return StringValues.Empty for missing entries.
+        /// </summary>
+        /// <param name="key"></param>
+        /// <returns>The stored value, or StringValues.Empty if the key is not present.</returns>
+        new StringValues this[string key] { get; set; }
```
Tratcher: I think this contract should only define `get` like IReadableStringCollection did. `set` raises a lot of consistency questions between request.Query and request.QueryString. The same goes for Form and Cookies.

I'm not positive we need the full IDictionary area either, but there's a lot more code to read here..

src/Microsoft.AspNet.Http.Features/IRequestCookies.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using System.Collections.Generic;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.Http
+{
+    /// <summary>
+    /// Represents the HttpRequest cookie collection
+    /// </summary>
+    public interface IRequestCookies : IDictionary<string, StringValues>
```
Tratcher: Review note: I'm trying to track down why these are in Features and not Abstractions.

src/Microsoft.AspNet.Http/Features/IRequestCookiesFeature.cs
```diff
@@ -1,10 +1,13 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
+using Microsoft.Extensions.Primitives;
```
Tratcher: nit: probably doesn't need these.

src/Microsoft.AspNet.Http/FormCollection.cs
```diff
@@ -10,29 +11,410 @@ namespace Microsoft.AspNet.Http.Internal
...
+#endif 
+        private static readonly Enumerator EmptyEnumerator = new Enumerator();
+        // Pre-box
+        private static readonly IEnumerator<KeyValuePair<string, StringValues>> EmptyIEnumeratorType = EmptyEnumerator;
+        private static readonly IEnumerator EmptyIEnumerator = EmptyEnumerator;
+
+        private static IFormFileCollection EmptyFiles = new FormFileCollection();
+
+        private IFormFileCollection _files;
+        private FormCollection()
```
Tratcher: nit: add an empty line

src/Microsoft.AspNet.Http/HeaderDictionary.cs
```diff
@@ -9,123 +9,273 @@
...
         /// </summary>
-        /// <returns>true if the <see cref="T:Microsoft.Owin.HeaderDictionary" /> is in read-only mode; otherwise, false.</returns>
-        public bool IsReadOnly
+        /// <param name="item">The item to add.</param>
+        public void Add(KeyValuePair<string, StringValues> item)
         {
-            get { return Store.IsReadOnly; }
+            if (item.Key == null)
+            {
+                throw new ArgumentNullException(nameof(item.Key));
```
Tratcher: ArgumentException with a message. Key is not one of Add's arguments.

src/Microsoft.AspNet.Http/QueryCollection.cs
```diff
@@ -0,0 +1,412 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.Http.Internal
+{
+    /// <summary>
+    /// The HttpRequest query string collection
+    /// </summary>
+    public sealed class QueryCollection : IQueryCollection
```
Tratcher: Yeah, I don't think we need most of this. The definition for ReadableStringCollection was adequate.

src/Microsoft.AspNet.Http/RequestCookies.cs
```diff
@@ -0,0 +1,467 @@
...
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Http.Internal
+{
+    public sealed class RequestCookies : IRequestCookies
```
Tratcher: Why aren't these named Collection like the others?

src/Microsoft.AspNet.WebUtilities/BufferedReadStream.cs
```diff
@@ -289,7 +289,7 @@ public bool EnsureBuffered(int minCount)
         {
             if (minCount > _buffer.Length)
             {
-                throw new ArgumentOutOfRangeException(nameof(minCount), minCount, "The value must be smaller than the buffer size: " + _buffer.Length);
+                throw new ArgumentOutOfRangeException(nameof(minCount), minCount, $"The value must be smaller than the buffer size: {_buffer.Length.ToString()}");
```
Tratcher: Note we do not use string interpolation for exception messages because they will eventually be moved to resource strings instead.

src/Microsoft.AspNet.WebUtilities/BufferedReadStream.cs
```diff
@@ -315,38 +315,42 @@ public bool EnsureBuffered(int minCount)
         public string ReadLine(int lengthLimit)
         {
             CheckDisposed();
-            var builder = new MemoryStream(200);
-            bool foundCR = false, foundCRLF = false;
-
-            while (!foundCRLF && EnsureBuffered())
+            using (var builder = new MemoryStream(200))
```
Tratcher: MemoryStream.Dispose doesn't really do anything. Are you calling it just as a best practice?

src/Microsoft.AspNet.WebUtilities/FormReader.cs
```diff
@@ -195,7 +195,24 @@ private void Buffer()
...
+
+            return result;
+        }
+
+        /// <summary>
+        /// Parses an HTTP form body - internal use, returns null if no entires.
+        /// </summary>
+        /// <param name="stream">The HTTP form body to parse.</param>
+        /// <returns>The collection containing the parsed HTTP form body, null if there are no entries.</returns>
+        public static async Task<Dictionary<string,StringValues>> ReadNullableFormAsync(Stream stream, Encoding encoding, CancellationToken cancellationToken = new CancellationToken())
```
Tratcher: This optimization seems impractical. Form POST's will almost always have at least one value.

src/Microsoft.AspNet.Http.Features/IRequestCookies.cs
benaadams: There was a build order problem; but I've moved `IFormCollection`/`IFormFile`/`IFormFileCollection`/`IQueryCollection`/`IRequestCookies` back to Abstractions and it doesn't seem to be an issue any more... Will update PR.

src/Microsoft.AspNet.WebUtilities/QueryHelpers.cs
```diff
@@ -107,13 +106,21 @@ public static string AddQueryString(string uri, IDictionary<string, string> quer
         /// </summary>
         /// <param name="text">The raw query string value, with or without the leading '?'.</param>
         /// <returns>A collection of parsed keys and values.</returns>
-        public static IDictionary<string, StringValues> ParseQuery(string queryString)
+        public static KeyValueAccumulator ParseQuery(string queryString)
```
Tratcher: `Dictionary<string, StringValues> ParseQuery` and `Dictionary<string, StringValues> ParseNullableQuery` like you did for Form?

src/Microsoft.AspNet.Http/RequestCookies.cs
benaadams: Was trying to match `ResponseCookies` and `IResponseCookies`?

Tratcher: They shouldn't match, request cookies are a collection, response cookies aren't.

benaadams: This statement confused me; then I had a look at request cookies and it really isn't a collection is it! Will change :)

src/Microsoft.AspNet.Http/QueryCollection.cs
benaadams: Path rewriting middleware; auth middleware (using query params; and removing them before moving up the pipeline)?

Tratcher: Manipulating the collection is problematic because it gets out of sync with request.QueryString. The recommended way to change something like Request.Query for downstream consumers is to update Request.QueryString, and then the QueryFeature will refresh. Similarly for cookies, you update the underlying headers. Path has no corresponding collection, just the raw PathString.

That said, I haven't seen any components need to re-write the query yet. Our auth middleware for example usually redirect if they've processed the request.

src/Microsoft.AspNet.Http.Abstractions/QueryString.cs
benaadams: Can you clarify? Or do you just mean the parameter type?

Tratcher: Yes, the parameter type here should be as generic as possible.

benaadams: Ok then to put this back and add 3rd overload which takes for `IQueryCollection` as it is non-boxing?

Tratcher: I don't think IQueryCollection->QueryString makes sense, it generally goes the other way. Or do you actually see it used that way somewhere?

benaadams: QueryFeature->Query->Set

src/Microsoft.AspNet.Http/QueryCollection.cs
benaadams: Forgot about redirects :+1:

src/Microsoft.AspNet.WebUtilities/BufferedReadStream.cs
benaadams: Best practice, disposing disposables, can remove?

src/Microsoft.AspNet.Http.Abstractions/QueryString.cs
justinvp: Does `IQueryCollection` implement `IEnumerable<KeyValuePair<string, StringValues>>`? If so, you can have just one overload that takes the `IEnumerable<...>` and internally have a fast path (non-boxing) for when `parameters` is `IQueryCollection`.

src/Microsoft.AspNet.WebUtilities/BufferedReadStream.cs
Tratcher: *shrug* do what you like, it won't have any affect here.

src/Microsoft.AspNet.Http.Abstractions/QueryString.cs
Tratcher: Right, IQueryCollection implements IEnumerable<KeyValuePair<string, StringValues>>. The Response.Query setter is primarily used in unit tests so there's no need to add an strange public API to optimize it.

benaadams: K, will drop all together

src/Microsoft.AspNet.Http/QueryCollection.cs
benaadams: Have changed the interfaces to be readonly, want me to remove the class direct functions also?

benaadams:
Summarising remaining (in case missed anything)
- [ ] ParseQuery -> match form parsing
- [ ] Remove writeable methods on classes (removed from interfaces, need confirm on this)
- [ ] Remove empty form post optimization (need confirm on this)

src/Microsoft.AspNet.Http.Features/IRequestCookies.cs
benaadams: Ahhh... remember now; wanted to return a strong type from the utility types e.g. `ParseQuery` and `ReadFormAsync` to avoid boxing on enumerator but `Microsoft.AspNet.Http` is after `Microsoft.AspNet.WebUtilities` in the build order; hence returning `KeyValueAccumulator` as a get around...

src/Microsoft.AspNet.Http/QueryCollection.cs
benaadams: Removed them; can do a `static` `Empty` type for each collection then.

benaadams:
Changes done; only outstanding is whether to remove the Form POST optimization for no values?

src/Microsoft.AspNet.WebUtilities/FormReader.cs
rynowak: There's definitely no scenario for this in MVC. You need an at least an xsrf token for a POST to any real site.

The 'ConfirmDelete' pattern that the old scaffolders support would be empty if not for xsrf :laughing:

benaadams: Not sure of code paths; but what would an action of unspecified method do? i.e. could be Get and/or Post (obv a non-modifying action); will it check the forms collection?

Tratcher: It has to check the content type first, and almost anything declared as a form would have at least one value.

src/Microsoft.AspNet.Http.Abstractions/IQueryCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+        ///     value.
+        /// </summary>
+        /// <param name="item">
+        ///     The object to locate in the <see cref="T:Microsoft.AspNet.Http.IQueryCollection" />.
+        /// </param>
+        /// <returns>
+        ///     true if item is found in the <see cref="T:Microsoft.AspNet.Http.IQueryCollection" />; otherwise,
+        ///     false.
+        /// </returns>
+        bool Contains(KeyValuePair<string, StringValues> item);
```
Tratcher: Remove

src/Microsoft.AspNet.Http.Abstractions/IQueryCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// </param>
+        /// <param name="arrayIndex">The zero-based index in array at which copying begins.</param>
+        /// <exception cref="T:System.ArgumentNullException">array is null.</exception>
+        /// <exception cref="T:System.ArgumentOutOfRangeException">arrayIndex is less than 0.</exception>
+        /// <exception cref="T:System.ArgumentException">
+        ///    The number of elements in the source <see cref="T:Microsoft.AspNet.Http.IQueryCollection" />
+        ///    is greater than the available space from arrayIndex to the end of the destination
+        ///    array.
+        /// </exception>
+        void CopyTo(KeyValuePair<string, StringValues>[] array, int arrayIndex);
```
Tratcher: Remove

src/Microsoft.AspNet.Http.Abstractions/IQueryCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        /// <summary>
+        ///     Gets an <see cref="T:System.Collections.Generic.ICollection`1" /> containing the values in the
+        ///     <see cref="T:Microsoft.AspNet.Http.IQueryCollection" />.
+        /// </summary>
+        /// <returns>
+        ///     An <see cref="T:System.Collections.Generic.ICollection`1" /> containing the values in the object
+        ///     that implements <see cref="T:Microsoft.AspNet.Http.IQueryCollection" />.
+        /// </returns>
+        ICollection<StringValues> Values { get; }
```
Tratcher: Remove

src/Microsoft.AspNet.WebUtilities/FormReader.cs
rynowak: @benaadams - https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormValueProviderFactory.cs#L27

We don't lift a finger unless it's really a formdata request.

benaadams:
Changes done

src/Microsoft.AspNet.Http.Abstractions/Extensions/MapWhenExtensions.cs
```diff
@@ -5,7 +5,6 @@
 using Microsoft.AspNet.Http;
```
davidfowl: Undo

src/Microsoft.AspNet.Http/Features/FormFeature.cs
```diff
@@ -77,17 +85,32 @@ public IFormCollection ReadForm()
                 throw new InvalidOperationException("Incorrect Content-Type: " + _request.ContentType);
             }
 
+            // TODO: Avoid Sync-over-Async http://blogs.msdn.com/b/pfxteam/archive/2012/04/13/10293638.aspx
```
davidfowl: Can you file a bug for this? Or put the bug number in the comment if you already filed one.

src/Microsoft.AspNet.Http/FormCollection.cs
```diff
@@ -10,29 +11,284 @@ namespace Microsoft.AspNet.Http.Internal
     /// <summary>
     /// Contains the parsed form values.
     /// </summary>
-    public class FormCollection : ReadableStringCollection, IFormCollection
+    public sealed class FormCollection : IFormCollection
```
davidfowl: Don't seal this

src/Microsoft.AspNet.Http/FormCollection.cs
```diff
@@ -10,29 +11,284 @@ namespace Microsoft.AspNet.Http.Internal
...
     {
-        public FormCollection(IDictionary<string, StringValues> store)
-            : this(store, new FormFileCollection())
+        public static readonly FormCollection Empty = new FormCollection();
+#if !DNXCORE50
```
davidfowl:
```C#
#if DNXCORE50
        private static readonly string[] EmptyKeys = Array.Empty<string>();
        private static readonly StringValues[] EmptyValues = Array.Empty<StringValues>();
#else
        private static readonly string[] EmptyKeys = new string[0];
        private static readonly StringValues[] EmptyValues = new StringValues[0];
#endif 
```

src/Microsoft.AspNet.Http/HeaderDictionary.cs
```diff
@@ -9,123 +9,273 @@
...
 {
     /// <summary>
-    /// Represents a wrapper for owin.RequestHeaders and owin.ResponseHeaders.
+    /// Represents a wrapper for RequestHeaders and ResponseHeaders.
     /// </summary>
-    public class HeaderDictionary : IHeaderDictionary
+    public sealed class HeaderDictionary : IHeaderDictionary
     {
-        public HeaderDictionary() : this(new Dictionary<string, StringValues>(StringComparer.OrdinalIgnoreCase))
+#if !DNXCORE50
```
davidfowl: Use the same pattern as above.

test/Microsoft.AspNet.Http.Extensions.Tests/HeaderDictionaryTypeExtensionsTest.cs
```diff
@@ -103,10 +103,11 @@ public void GetListT_KnownTypeWithValidValue_Success()
         public void GetListT_KnownTypeWithMissingValue_Null()
         {
             var context = new DefaultHttpContext();
-
-            var result = context.Request.GetTypedHeaders().GetList<MediaTypeHeaderValue>(HeaderNames.Accept);
-
-            Assert.Null(result);
+            var result1 = context.Request.GetTypedHeaders().GetList<MediaTypeHeaderValue>(HeaderNames.Accept);
```
davidfowl: ?

benaadams: Must have been something extra that was there at some point in the commit changes; will revert

benaadams:
Next change set done

/cc @davidfowl

src/Microsoft.AspNet.Http.Abstractions/IFormCollection.cs
```diff
@@ -1,13 +1,137 @@
...
+        ///     value.
+        /// </summary>
+        /// <param name="item">
+        ///     The object to locate in the <see cref="T:Microsoft.AspNet.Http.IFormCollection" />.
+        /// </param>
+        /// <returns>
+        ///     true if item is found in the <see cref="T:Microsoft.AspNet.Http.IFormCollection" />; otherwise,
+        ///     false.
+        /// </returns>
+        bool Contains(KeyValuePair<string, StringValues> item);
```
Tratcher: Same as query

src/Microsoft.AspNet.Http.Abstractions/IRequestCookieCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+        ///     value.
+        /// </summary>
+        /// <param name="item">
+        ///     The object to locate in the <see cref="T:Microsoft.AspNet.Http.IRequestCookieCollection" />.
+        /// </param>
+        /// <returns>
+        ///     true if item is found in the <see cref="T:Microsoft.AspNet.Http.IRequestCookieCollection" />; otherwise,
+        ///     false.
+        /// </returns>
+        bool Contains(KeyValuePair<string, StringValues> item);
```
Tratcher: Same as query

benaadams:
Done

/cc @Tratcher

benaadams:
Merge base

benaadams:
Rebased rather than merged

Tratcher:
@rynowak How badly will this break MVC?

src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryTypeExtensions.cs
```diff
@@ -80,7 +80,55 @@ internal static void SetList<T>(this IHeaderDictionary headers, string name, ILi
             }
             else
             {
-                headers[name] = values.Select(value => value.ToString()).ToArray();
+
+                switch (values.Count)
+                {
+                    case 0:
```
Tratcher: values.Count == 0 was already checked in the prior if block. Might as well make this `if..else if...else`.

rynowak:
@tratcher - I'd expect much more breakage in the tests which frequently new up collections for testing. We actually read these very few places in the product code.

src/Microsoft.AspNet.Http/RequestCookieCollection.cs
```diff
@@ -0,0 +1,269 @@
...
+                string value;
+                if (TryGetValue(key, out value))
+                {
+                    return value;
+                }
+                return string.Empty;
+            }
+        }
+
+        StringValues IRequestCookieCollection.this[string key]
```
Tratcher: Something is strange here, there shouldn't be two indexers.

benaadams: string and StringValues

src/Microsoft.AspNet.Http/RequestCookieCollection.cs
```diff
@@ -0,0 +1,269 @@
...
+        /// </summary>
+        /// <param name="key"></param>
+        /// <returns></returns>
+        public IList<string> GetValues(string key)
+        {
+            string value;
+            return TryGetValue(key, out value) ? new[] { value } : null;
+        }
+        
+        public static RequestCookieCollection Reparse(IList<string> values)
```
Tratcher: The point of Reparse was to re-fill the existing collection and avoid allocating a new one. That said, none of the other collections work that way and I don't expect much dynamic cookie updating during a request. Just rename this to Parse.

src/Microsoft.AspNet.Http/RequestCookieCollection.cs
```diff
@@ -0,0 +1,269 @@
...
+        {
+            if (Store == null)
+            {
+                value = default(string);
+                return false;
+            }
+            return Store.TryGetValue(key, out value);
+        }
+
+        bool IRequestCookieCollection.TryGetValue(string key, out StringValues value)
```
Tratcher: Again, there doesn't need to be two TryGetValue methods.

src/Microsoft.AspNet.Http/RequestCookieCollection.cs
Tratcher: Ah, right, this collection really is different from the others. It doesn't need to use StringValues anywhere, it only had that because of the old common IReadableStringCollection. We can just make this all `<string, string>` now.

benaadams: Ah ok, makes things simpiler :)

src/Microsoft.AspNet.WebUtilities/MultipartReader.cs
```diff
@@ -74,7 +74,7 @@ public MultipartReader(string boundary, Stream stream, int bufferSize)
             return new MultipartSection() { Headers = headers, Body = _currentStream, BaseStreamOffset = baseStreamOffset };
         }
 
-        private async Task<IDictionary<string, StringValues>> ReadHeadersAsync(CancellationToken cancellationToken)
+        private async Task<Dictionary<string, StringValues>> ReadHeadersAsync(CancellationToken cancellationToken)
```
Tratcher: What the the reason for using concrete Dictionary in public APIs?

Tratcher:
This is pretty much done. I'll rebase and merge it when I get a chance later today. Any of the minor remaining comments you don't get to by then I'll just take care of.  Thanks

src/Microsoft.AspNet.WebUtilities/MultipartReader.cs
benaadams: It gets added to a HeaderDictionary which uses Dictionary for struct enumerators; if you could enumerate using struct via an interface without boxing the world would be a simpler place :(

src/Microsoft.AspNet.Http/RequestCookieCollection.cs
benaadams: Hmm... how does that fit with RequestCookieFeature? https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http/Features/RequestCookiesFeature.cs#L62

Multiple entries for the same value as via headers? Concat?

benaadams: nvm; I think I've got it

benaadams:
Done, however quick check am returning `string.Empty` where a cookie doesn't exist which is `""` rather than `null`; is this the expected behavior?

Tratcher:
Should be fine. If they really care they can use TryGetValue

Tratcher:
Rebased, squashed, merged.

benaadams:
:clap: :)

src/Microsoft.AspNet.Http.Abstractions/PathString.cs
NickCraver: Yes, *please*. We have a `StringSplits` class with static arrays that's a copy/paste to every high perf project at Stack Overflow - it's silly IMO. This could also be adapted as a compiler optimization.

davidfowl: See https://github.com/dotnet/corefxlab/issues/245

src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
```diff
@@ -25,7 +25,7 @@ public static class UriHelper
             FragmentString fragment = new FragmentString())
         {
             string combinePath = (pathBase.HasValue || path.HasValue) ? (pathBase + path).ToString() : "/";
-            return combinePath + query + fragment;
+            return $"{combinePath}{query.ToString()}{fragment.ToString()}";
```
NickCraver: Since this has just 3 components, it would be more efficient reverting to the `String.Concat()` a+b+c format it was in before. While maintaining equal readability, it's ~3-4x faster on most machines (I've tested this on a variety locally and on the latest servers).

src/Microsoft.AspNet.WebUtilities/QueryHelpers.cs
```diff
@@ -107,21 +107,46 @@ public static string AddQueryString(string uri, IDictionary<string, string> quer
...
+        {
             var accumulator = new KeyValueAccumulator();
 
+            if (string.IsNullOrEmpty(queryString) || queryString == "?")
+            {
+                return null;
+            }
+
+            int scanIndex = 0;
+            if (!string.IsNullOrEmpty(queryString) && queryString[0] == '?')
```
NickCraver: Redundant check here on the IsNullOrEmpty()

src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
davidfowl: Send a PR :smile:

src/Microsoft.AspNet.Http.Abstractions/FragmentString.cs
```diff
@@ -100,7 +100,7 @@ public static FragmentString FromUriComponent(Uri uri)
             string fragmentValue = uri.GetComponents(UriComponents.Fragment, UriFormat.UriEscaped);
             if (!string.IsNullOrEmpty(fragmentValue))
             {
-                fragmentValue = "#" + fragmentValue;
+                fragmentValue = $"#{fragmentValue}";
```
yishaigalatzer: More of a question, how is this more efficient?

src/Microsoft.AspNet.Http.Extensions/Internal/ParsingHelpers.cs
```diff
@@ -61,35 +61,26 @@ public static void SetHeaderJoined(IDictionary<string, StringValues> headers, st
...
-            else if (value.Contains(','))
-            {
-                if (value[0] != '"' || value[value.Length - 1] != '"')
-                {
-                    value = '"' + value + '"';
-                }
+            if (!string.IsNullOrWhiteSpace(value) && 
+                value.Contains(',') && 
+                (value[0] != '"' || value[value.Length - 1] != '"'))
+            {
```
yishaigalatzer: this if does the same thing, but is a lot less readable or debuggable

src/Microsoft.AspNet.Http.Abstractions/FragmentString.cs
halter73: I don't think it is: https://github.com/aspnet/HttpAbstractions/pull/411#discussion_r43713247

src/Microsoft.AspNet.Http.Extensions/HeaderDictionaryExtensions.cs
Tratcher: This actually caused a subtle behavior change. The implicit converter would return null in some cases, where ToArray always returns at least an empty value. The same for ToString.
https://github.com/aspnet/Common/blob/dev/src/Microsoft.Extensions.Primitives/StringValues.cs#L104
https://github.com/aspnet/Common/blob/dev/src/Microsoft.Extensions.Primitives/StringValues.cs#L85

benaadams: :(

Needs tests?

benaadams: Much shame; corner cases and too many changes for a single PR - lesson learnt. Seems to have engaged people though :D

PR #413, opened 2015-09-22
Proposal: Form parsing improvements

benaadams:
Cached task pattern for avoiding state machine and task allocation for repeated ReadFormAsync
struct KeyValueAccumulator
Delayed allocations; reduced allocs for empty paths


src/Microsoft.AspNet.Http/ReadableStringCollection.cs
```diff
@@ -13,7 +13,15 @@ namespace Microsoft.AspNet.Http.Internal
     /// </summary>
     public class ReadableStringCollection : IReadableStringCollection
     {
-        public static readonly IReadableStringCollection Empty = new ReadableStringCollection(new Dictionary<string, StringValues>(0));
+        public static readonly IReadableStringCollection Empty = new ReadableStringCollection();
+
+        private static readonly string[] EmptyKeys = new string[0];
+        private static readonly IEnumerable<KeyValuePair<string, StringValues>> EmptyEnumerable = (IEnumerable<KeyValuePair<string, StringValues>>)(new KeyValuePair<string, StringValues>[0]);
```
davidfowl: `Enumerable.Empty<T>`

src/Microsoft.AspNet.WebUtilities/KeyValueAccumulator.cs
```diff
@@ -7,17 +7,16 @@
 
 namespace Microsoft.AspNet.WebUtilities
 {
-    public class KeyValueAccumulator
+    public struct KeyValueAccumulator
```
davidfowl: Does this even need an inner dictionary? It could be just have a callback OnKeyValuePair(string key, string value) or something.

src/Microsoft.AspNet.Http/ReadableStringCollection.cs
benaadams: Not sure where that is? Can't seem to grab a ref.

Can do the ```#if !DNXCORE50``` stuff on ```EmptyKeys``` with ```Array.Empty<string>()``` though?
Also ```#if DNXCORE50``` for ```EmptyEnumerable = Array.Empty<KeyValuePair<string, StringValues>>())```

Didn't know if it was too messy

src/Microsoft.AspNet.WebUtilities/KeyValueAccumulator.cs
benaadams: Could do but would need the allocation of an outer dictionary, a closure; or some code rearranging? For example its used in the extension method [```QueryHelpers.ParseQuery```](https://github.com/aspnet/HttpAbstractions/blob/ceded805f32e3aca0066182fd96bbf171bd40574/src/Microsoft.AspNet.WebUtilities/QueryHelpers.cs#L116)

You can avoid alloc with these changes by checking ```HasValues``` before calling ```GetResults()``` as is done in the [```FormFeature.InnerReadFormAsync```](https://github.com/aspnet/HttpAbstractions/pull/413/files#diff-25040e9a807f3ea89b0c6dc24fa8d50dR189)

src/Microsoft.AspNet.Http/Features/FormFeature.cs
```diff
@@ -119,7 +144,7 @@ public IFormCollection ReadForm()
                     var section = await multipartReader.ReadNextSectionAsync(cancellationToken);
                     while (section != null)
                     {
-                        var headers = new HeaderDictionary(section.Headers);
+                        var headers = section.Headers;
                         ContentDispositionHeaderValue contentDisposition;
                         ContentDispositionHeaderValue.TryParse(headers[HeaderNames.ContentDisposition], out contentDisposition);
```
Tratcher: regression: this will throw if Content-Disposition is missing. You could just use section.ContentDisposition instead.

benaadams: Which would wrap; so delay wrapping down a few lines for ```FormFile``` wouldn't help. Will revert.

Tratcher:
Can you share your metrics for these changes?

src/Microsoft.AspNet.Http/Features/FormFeature.cs
Tratcher: That wouldn't wrap: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.WebUtilities/MultipartSection.cs#L25-L38

benaadams: Changed as suggested

benaadams:
> Can you share your metrics for these changes?

Can't actually run it atm; the unstable feeds are a few versions back so need to wait for them to be updated

benaadams:
Resolves #368 ?

src/Microsoft.AspNet.Http/Features/FormFeature.cs
```diff
@@ -98,10 +123,10 @@ public IFormCollection ReadForm()
             _request.EnableRewind();
 
             IDictionary<string, StringValues> formFields = null;
-            var files = new FormFileCollection();
+            FormFileCollection files = null;
 
             // Some of these code paths use StreamReader which does not support cancellation tokens.
-            using (cancellationToken.Register(_request.HttpContext.Abort))
+            using (cancellationToken.Register((Action)_request.HttpContext.Abort))
```
davidfowl: Why are you casting here? You can actually avoid the delegate allocation per request here as well.

```
cancellationToken.Register(state => ((HttpContext)state).Abort, _request.HttpContext)
```

benaadams: Couldn't work out how to avoid alloc; that will do nicely :)

src/Microsoft.AspNet.Http/Features/FormFeature.cs
```diff
@@ -63,7 +65,15 @@ public bool HasFormContentType
             }
         }
 
-        public IFormCollection Form { get; set; }
+        public IFormCollection Form
+        {
+            get { return _form; }
+            set
+            {
+                _parsedFormTask = null;
```
davidfowl: Why not set the cached task here?

src/Microsoft.AspNet.Http/Features/FormFeature.cs
```diff
@@ -129,8 +153,12 @@ public IFormCollection ReadForm()
 
                             var file = new FormFile(_request.Body, section.BaseStreamOffset.Value, section.Body.Length)
                             {
-                                Headers = headers,
+                                Headers = new HeaderDictionary(section.Headers),
```
davidfowl: Another optimization might be to use a different collection type that implement IHeaderDictionary. I'm assuming this sub header dictionary normally has a tiny number of items in the most common cases. Something to consider in the future.

/cc @Tratcher

src/Microsoft.AspNet.Http/Features/FormFeature.cs
benaadams: You've got the non-async method; so didn't want to set the task if that's the only path used

benaadams:
Will be covered by rebased #411

PR #420, opened 2015-09-27
Remove FeatureReference indirection

benaadams:
~~Is more code~~ Is greatly improved in recent incarnation.

Doesn't cause any extra casting; shouldn't be extra instructions executed; cuts ```CreateHttpContext``` allocation by 17.5%

From 1824 kB per 4k requests

logging-scope

To 1504 kB per 4k requests

logging-scope-prev

So down by 320 kB per 4k requests; as similar saving to the ```AsyncLocal``` [change in coreclr](https://github.com/dotnet/coreclr/pull/1629)

### DIFF
--- a/src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
+++ b/src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
@@ -12,25 +12,44 @@ using Microsoft.AspNet.Http.Features.Authentication.Internal;
 
 namespace Microsoft.AspNet.Http.Authentication.Internal
 {
-    public class DefaultAuthenticationManager : AuthenticationManager
+    public class DefaultAuthenticationManager : AuthenticationManager, IFeatureCache
     {
         private readonly IFeatureCollection _features;
-        private FeatureReference<IHttpAuthenticationFeature> _authentication = FeatureReference<IHttpAuthenticationFeature>.Default;
-        private FeatureReference<IHttpResponseFeature> _response = FeatureReference<IHttpResponseFeature>.Default;
+        private int _cachedFeaturesRevision = -1;
+
+        private IHttpAuthenticationFeature _authentication;
+        private IHttpResponseFeature _response;
 
         public DefaultAuthenticationManager(IFeatureCollection features)
         {
             _features = features;
         }
 
+        void IFeatureCache.CheckFeaturesRevision()
+        {
+            if (_cachedFeaturesRevision != _features.Revision)
+            {
+                _authentication = null;
+                _response = null;
+                _cachedFeaturesRevision = _features.Revision;
+            }
+        }
+
         private IHttpAuthenticationFeature HttpAuthenticationFeature
         {
-            get { return _authentication.Fetch(_features) ?? _authentication.Update(_features, new HttpAuthenticationFeature()); }
+            get
+            {
+                return FeatureHelpers.GetOrCreateAndCache(
+                    this, 
+                    _features, 
+                    () => new HttpAuthenticationFeature(), 
+                    ref _authentication);
+            }
         }
 
         private IHttpResponseFeature HttpResponseFeature
         {
-            get { return _response.Fetch(_features); }
+            get { return FeatureHelpers.GetAndCache(this, _features, ref _response); }
         }
 
         public override IEnumerable<AuthenticationDescription> GetAuthenticationSchemes()
--- a/src/Microsoft.AspNet.Http/DefaultConnectionInfo.cs
+++ b/src/Microsoft.AspNet.Http/DefaultConnectionInfo.cs
@@ -10,26 +10,51 @@ using Microsoft.AspNet.Http.Features.Internal;
 
 namespace Microsoft.AspNet.Http.Internal
 {
-    public class DefaultConnectionInfo : ConnectionInfo
+    public class DefaultConnectionInfo : ConnectionInfo, IFeatureCache
     {
         private readonly IFeatureCollection _features;
+        private int _cachedFeaturesRevision = -1;
 
-        private FeatureReference<IHttpConnectionFeature> _connection = FeatureReference<IHttpConnectionFeature>.Default;
-        private FeatureReference<ITlsConnectionFeature> _tlsConnection = FeatureReference<ITlsConnectionFeature>.Default;
+        private IHttpConnectionFeature _connection;
+        private ITlsConnectionFeature _tlsConnection;
 
         public DefaultConnectionInfo(IFeatureCollection features)
         {
             _features = features;
         }
 
+        void IFeatureCache.CheckFeaturesRevision()
+        {
+            if (_cachedFeaturesRevision != _features.Revision)
+            {
+                _connection = null;
+                _tlsConnection = null;
+                _cachedFeaturesRevision = _features.Revision;
+            }
+        }
+
         private IHttpConnectionFeature HttpConnectionFeature
         {
-            get { return _connection.Fetch(_features) ?? _connection.Update(_features, new HttpConnectionFeature()); }
+            get
+            {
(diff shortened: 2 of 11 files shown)


davidfowl:
@lodejard please review.

lodejard:
This is removing a few ints, doesn't reduce the number of objects allocated, and has much more manual code in several objects that reference features. I don't think it's worth the change...

benaadams:
Will come back with some more analysis; but agreed its a lot more messy

benaadams:
Its down by 80 bytes a request which if it were just ints (4 bytes) would be 20 ints; which its clearly not; so not entirely sure what's happening.

Perhaps on 64bit the structs (IntPtr 8 bytes + Int 4 bytes) 8 byte align with a 4 byte padding; doubling the effect? Still seems high though.

benaadams:
Ok, is a memory layout thing, exacerbated on x64, with the 4 memory layouts for DefaultHttpContext looking like the following

memory layout

lodejard:
Hrmmm - yeah, that does seem like a really unfortunate bloating and a reasonable way to avoid it. Plus it's living in per-request memory pressure. Thanks for digging into it deeper!

And if we want to keep up the manually-maintained programming pattern under control down the road we can use the Xxx.Generated.cs partial class trick for the cached-feature code paths. Don't need to do that now as part of this PR, but it's good to keep in mind if the extra code is considered unfortunate.

:shipit: 

Last thought... We *might* want to keep the FeatureReference class around, if middleware ever have per-request contexts that query for particular features at multiple points in time. The FR might still be a good halfway point for those cases.

Tratcher:
Is there at least some way to share this bookkeeping code? This looks like it will cause maintenance problems.

benaadams:
> Is there at least some way to share this bookkeeping code?

Could probably do something with extensions and ```ref``` params; don't know what the feel on ```ref``` is though? Haven't seem much of it about.

Tratcher:
Sample?

benaadams:
Would have to be static methods, rather than extensions as there are nulls involved; and also a couple variants as some have different paths, but something like:
````csharp
internal interface IFeatureCache
{
    void CheckFeaturesRevision();
}
````
````csharp
internal sealed class FeatureCacheHelpers
{
    public static T GetAndCacheFeature<T>(
        IFeatureCache cache, 
        IFeatureCollection features, 
        ref T cachedObject)
    {
        cache.CheckFeaturesRevision();
        if (cachedObject == null)
        {
            cachedObject = features.Get<T>();
        }
        return cachedObject;
    }
    public static T GetOrCreateAndCacheFeature<T, F>(
        IFeatureCache cache, 
        IFeatureCollection features,
        ref T cachedObject) where F : T, new()
    {
        cache.CheckFeaturesRevision();
        if (cachedObject == null)
        {
            cachedObject = features.Get<T>();
            if (cachedObject == null)
            {
                cachedObject = new F();
                features.Set(cachedObject);
            }
        }
        return cachedObject;
    }
}
````
Then for example ```DefaultAuthenticationManager``` would become
````csharp
public class DefaultAuthenticationManager : AuthenticationManager, IFeatureCache
{
    private readonly IFeatureCollection _features;
    private int _cachedFeaturesRevision = -1;

    private IHttpAuthenticationFeature _authentication;
    private IHttpResponseFeature _response;

    public DefaultAuthenticationManager(IFeatureCollection features)
    {
        _features = features;
    }

    void IFeatureCache.CheckFeaturesRevision()
    {
        if (_cachedFeaturesRevision != _features.Revision)
        {
            _authentication = null;
            _response = null;
            _cachedFeaturesRevision = _features.Revision;
        }
    }

    private IHttpAuthenticationFeature HttpAuthenticationFeature
    {
        get
        {
            return FeatureCacheHelpers
                .GetOrCreateAndCacheFeature<IHttpAuthenticationFeature, HttpAuthenticationFeature>
                (this, _features, ref _authentication);
        }
    }

    private IHttpResponseFeature HttpResponseFeature
    {
        get { return FeatureCacheHelpers.GetAndCacheFeature(this, _features, ref _response); }
    }

...
````

**Comment by benaadams** [comment]
@Tratcher Something more like the PR now?

**Comment by Tratcher** [comment]
Better. Really what you've done is turn FeatureReference from a struct to a static helper class. One more comment inline.

I think Lou still wants to save FeatureReference for use elsewhere.

File: src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs [hunk 1]
```diff
@@ -12,25 +12,44 @@
...
+
         private IHttpAuthenticationFeature HttpAuthenticationFeature
         {
-            get { return _authentication.Fetch(_features) ?? _authentication.Update(_features, new HttpAuthenticationFeature()); }
+            get
+            {
+                return FeatureHelpers.GetOrCreateAndCache(
+                    this, 
+                    _features, 
+                    () => new HttpAuthenticationFeature(),
```
Tratcher: Does this lamda bother you? Or are you relying on it getting optimized into a static field?

benaadams: It should be a compile time const I believe? Was being careful not to create any closures which would then cause runtime pre-call creation.

Tratcher: It's worth verifying.

benaadams:
> Really what you've done is turn FeatureReference from a struct to a static helper class. 

Pretty much, took a few iterators to get there though :)

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
davidfowl: It will be optimized. If we want to be more anal, we can declare static delegates.

davidfowl:
This looks like a good change and is much simpler than previous incantations :smile:

benaadams:
Pushing 	FastRequestIdentifierFeature -> DefaultRequestIdentifierFeature into Abstractions as a default type from Hosting; which also requires new static; then allows lazy instantiation back in Hosting.

src/Microsoft.AspNet.Http.Abstractions/HttpContext.cs
```diff
@@ -36,6 +36,8 @@ public abstract class HttpContext : IDisposable
 
         public abstract ISession Session { get; set; }
 
+        public abstract string TraceIdentifier { get; set; }
```
davidfowl: API change. I actually think this should be exposed but this change just went from being an optimization to an API change.

benaadams: Happy to back peddle; but want to put a PR on a PR and my git fu is not strong enough...

davidfowl: Revert the last commit and force push

benaadams: Will await instruction.

benaadams: Reverted; have branched other changes to different repo

src/Microsoft.AspNet.Http.Features/FeatureReference.cs
```diff
@@ -1,38 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-namespace Microsoft.AspNet.Http.Features
```
davidfowl: @benaadams Bring this type back then I'll be ready to merge this PR. Also squash the commits.

benaadams:
Added FeatureReference back and squashed commits; passed tests

davidfowl:
LGTM

PR #421, opened 2015-09-29
Auth changes: error handling + make error txt less sucky

HaoK:


davidfowl:
Context?

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
```diff
@@ -62,75 +61,65 @@ public override IEnumerable<AuthenticationDescription> GetAuthenticationSchemes(
...
             {
-                await handler.ChallengeAsync(challengeContext);
+                await handler.ChallengeAsync(context);
             }
 
-            if (!challengeContext.Accepted)
+            if (!context.Accepted)
             {
-                throw new InvalidOperationException($"The following authentication scheme was not accepted: {authenticationScheme}");
+                throw new InvalidOperationException($"No authentication handler is configured to handle a challenge for the scheme: {context.AuthenticationScheme}");
```
Tratcher: this still reads strangely if the challenge was for string.Empty.

HaoK:
There's currently no way to shortcut that a redirect has happened and no further processing should be done for the rest of the request in auth. This is an attempt to put that in all in the contexts consistently. The security pr is more interesting.

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
HaoK: I guess we can special case that and replace empty or null with something like automatic authentication

src/Microsoft.AspNet.Http.Features/Authentication/AuthenticateContext.cs
```diff
@@ -41,5 +41,12 @@ public virtual void NotAuthenticated()
         {
             Accepted = true;
         }
+
+        public bool IsRequestCompleted { get; private set; }
```
Tratcher: I think IsRequestCompleted is redundant with Accepted (and possibly Response.HasStarted, depending on what you're trying to do).

HaoK: Yeah I guess authenticate is special, maybe we don't need this on the authenticate context since most of the handlers have specialized implementations anyways, and its mostly creating the ticket/error, as opposed to actually finishing the response/request

HaoK:
Reverted a bunch of changes we don't need.  Added an Exception and Failed helper to AuthenticateContext.  Will tweak error msgs a bit more

HaoK:
@blowdart do you want to wordsmith these errors a bit?  We also need alternate text for when no/empty auth scheme is specified for all of these

blowdart:
The words as they stand are fine. For the no handlers at all I'd just do "No Authentication handlers are registered". Up to you if you want to disambiguate the operation, you'd get that from the call stack though.

HaoK:
Ok i'll standardize the text for all of them then

HaoK:
Okay went with: 

```
"No authentication handler is configured to handle the scheme: {scheme|<automatic>}
```

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
```diff
@@ -109,7 +110,8 @@ public override IEnumerable<AuthenticationDescription> GetAuthenticationSchemes(
 
             if (!signInContext.Accepted)
             {
-                throw new InvalidOperationException($"The following authentication scheme was not accepted: {authenticationScheme}");
+                var scheme = string.IsNullOrEmpty(authenticationScheme) ? "<automatic>" : authenticationScheme;
```
Tratcher: SignIn can't be automatic.

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
```diff
@@ -130,7 +132,8 @@ public override IEnumerable<AuthenticationDescription> GetAuthenticationSchemes(
 
             if (!signOutContext.Accepted)
             {
-                throw new InvalidOperationException($"The following authentication scheme was not accepted: {authenticationScheme}");
+                var scheme = string.IsNullOrEmpty(authenticationScheme) ? "<automatic>" : authenticationScheme;
```
Tratcher: SignOut can't be automatic.

HaoK:
Updated removing automatic from SignIn/SignOut

Tratcher:
:shipit:

HaoK:
@Tratcher Updated PR added AutomaticScheme constant to AuthenticationManager since Challenge overloads use this value (formerly string.empty)

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
```diff
@@ -11,6 +11,11 @@ namespace Microsoft.AspNet.Http.Authentication
 {
     public abstract class AuthenticationManager
     {
+        /// <summary>
+        /// Constant used to represent the automatic scheme
+        /// </summary>
+        public const string AutomaticScheme = "(_automatic_)";
```
Tratcher: That will look terrible in error messages. `Automatic` should be fine, I'm not too worried about collisions.

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
```diff
@@ -34,7 +39,7 @@ public virtual Task ChallengeAsync()
 
         public virtual Task ChallengeAsync(AuthenticationProperties properties)
         {
-            return ChallengeAsync(authenticationScheme: string.Empty, properties: properties);
+            return ChallengeAsync(authenticationScheme: AutomaticScheme, properties: properties);
```
Tratcher: If we're going to do this should we also change the null checks to IsNullOrEmpty?

src/Microsoft.AspNet.Http.Features/Authentication/ChallengeContext.cs
```diff
@@ -15,11 +15,6 @@ public ChallengeContext(string authenticationScheme)
 
         public ChallengeContext(string authenticationScheme, IDictionary<string, string> properties, ChallengeBehavior behavior)
         {
-            if (authenticationScheme == null)
```
Tratcher: ?

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
```diff
@@ -83,7 +83,8 @@ public override IEnumerable<AuthenticationDescription> GetAuthenticationSchemes(
 
             if (!challengeContext.Accepted)
             {
-                throw new InvalidOperationException($"The following authentication scheme was not accepted: {authenticationScheme}");
+                var scheme = string.IsNullOrEmpty(authenticationScheme) ? "<automatic>" : authenticationScheme;
```
Tratcher: Don't need this anymore.

src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
HaoK: Sure I guess that's fine

HaoK:
Updated

Tratcher:
:shipit:

HaoK:
d28c6e1dbbd6156a6168e331ec8137044cd05ee9

PR #428, opened 2015-10-03
Move BufferedHtmlContent to HttpAbstractions

pranavkm:
Current layering is incorrect


Tratcher:
I'm amused that we keep moving HTML stuff into HTTP Abstractions....

rynowak:
We should move all the HTTP stuff into it's own repo... maybe HtmlAbstractions?

davidfowl:
I told you we should move it :smile: /cc @Eilon

PR #434, opened 2015-10-08
Incorporate StringSegment from primitives

rynowak:
/cc @Tratcher


rynowak:
The code here: https://github.com/aspnet/HttpAbstractions/pull/434/files#diff-39dff6421a269d5f55533106f769fb40L188 will throw now if you try to access `Current` while in an invalid state. Is this a concern? This code is internal, so I expect that callers are calling `MoveNext()` properly.

Tratcher:
Not too concerned until it breaks :-). :shipit:

davidfowl:
:shipit:

rynowak:
Feel free to merge this if you want, I'm in the desert :cactus: :camel: :cactus:

davidfowl:
@rynowak plzmergekthxbye

PR #438, opened 2015-10-13
Update `MediaTypeHeaderValue.IsSubsetOf()` to perform consistent checks

dougbu:
- aspnet/Mvc#3138 part 1/2
  - check parameters with same polarity as type and subtype
    - ignore quality factors
  - bug was obscured because MVC has no formatters supporting wildcard media types

nits:
- add doc comments
- spelling
- correct typo in a `project.json` file


src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -241,22 +257,29 @@ public bool IsSubsetOf(MediaTypeHeaderValue otherMediaType)
                 }
             }
 
-            if (Parameters != null)
+            // "text/plain; charset=utf-8; level=1" is a subset of "text/plain; charset=utf-8". In turn
+            // "text/plain; charset=utf-8" is a subset of "text/plain".
+            if (otherMediaType._parameters != null && otherMediaType.Parameters.Count != 0)
```
Tratcher: It's weird to mix fields and properties here. Pick one.

Tratcher:
:shipit:

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
dougbu: :ok: will use the field because first reference to the property will allocate.

dougbu:
0581bcf

PR #435, opened 2015-10-10
[Proposal] Move RequestIdentifierFeature to HttpContext

benaadams:
Allow lazier instantiation
Expose TraceIdentifier on Httpcontext
Also resolves #412

Add tests

PR #440, opened 2015-10-15
Move RequestIdentifierFeature to HttpContext

### LINKED ISSUE #412: Expose TraceIdentifier as a property on HttpContext
It's pretty convenient to make available.

benaadams:
Rebased #435
Allow lazier instantiation
Expose TraceIdentifier on Httpcontext
Also resolves #412

Add tests

### DIFF
--- a/src/Microsoft.AspNet.Http/DefaultHttpContext.cs
+++ b/src/Microsoft.AspNet.Http/DefaultHttpContext.cs
@@ -118,6 +118,15 @@ namespace Microsoft.AspNet.Http.Internal
             }
         }
 
+        private IHttpRequestIdentifierFeature RequestIdentifierFeature
+        {
+            get {
+                return FeatureHelpers.GetOrCreate<IHttpRequestIdentifierFeature>(
+                  _features,
+                  () => new HttpRequestIdentifierFeature());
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Http/DefaultHttpContext.cs
```diff
@@ -118,6 +118,19 @@ private ISessionFeature SessionFeature
...
         }
 
+        private IHttpRequestIdentifierFeature RequestIdentifierFeature
+        {
+            get {
+                return FeatureHelpers.GetOrCreate<IHttpRequestIdentifierFeature>(
+                  _features,
+                  () => new HttpRequestIdentifierFeature());
+            }
+            set
```
Tratcher: This setter is never used, correct? Remove it.

Tratcher:
:shipit: after my minor comment.

src/Microsoft.AspNet.Http/DefaultHttpContext.cs
benaadams: Removed, squished

benaadams:
Applied feedback, squished

Issue #442, opened 2015-10-19
Unit test Microsoft.AspNet.Html.Abstractions.Test.HtmlContentBuilderExtensionsTest.Builder_AppendFormat_WithDifferentCurrentCulture fails on CoreCLR on Linux


cesarbs:
Assert.Equal() Failure
                                ↓ (pos 12)
          Expected: HtmlEncode[[01 February 2015]]
          Actual:   HtmlEncode[[Sunday, 1 February 2015]]
                                ↑ (pos 12)
          Stack Trace:
               at Microsoft.AspNet.Html.Abstractions.Test.HtmlContentBuilderExtensionsTest.Builder_AppendFormat_WithDifferentCurrentCulture()
               at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
               at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
               at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)

cesarbs:
This affects https://github.com/aspnet/Universe/issues/304.

cesarbs:
Opened CoreCLR issue https://github.com/dotnet/corefx/issues/3974

PR #443, opened 2015-10-19
Fix CoreCLR test pass on Linux

cesarbs:
Use a locale that does not show date formatting discrepancies between platorms.

#442


cesarbs:
@Tratcher

Tratcher:
:shipit:
Please include the bug numbers in your commit messages.

rynowak:
:shipit:

PR #448, opened 2015-10-20
Url Path Decoder

troydai:
Addressing: https://github.com/aspnet/KestrelHttpServer/issues/124
/cc @Tratcher @blowdart @lodejard 

Note:
- Unescaping every thing except `/`. `%2F` will be leave in the string.
- If a sequence can't be correctly decoded under UTF8, just leave them as is.
- Supposed to be O(n). Need more performance testing to prove.

Follow up:
- [x] Still working on more tests.
- [ ] More performance tuning will happen as result of performance test.
- [x] Also, I'm looking at a change to potentially decode the string in place without allocate a new char array.
- [ ] Also, need security review.


src/Microsoft.Extensions.WebEncoders.Core/PathDecoder.cs
```diff
@@ -0,0 +1,259 @@
...
+                }
+            }
+
+            return length;
+        }
+
+        /// <summary>
+        /// Find the next % in the sequence. If % is not found, return the sequence length.
+        /// </summary>
+        private static int NextPercentage(int start, string source)
```
davidfowl: GetNextEncoded?

src/Microsoft.Extensions.WebEncoders.Core/PathDecoder.cs
```diff
@@ -0,0 +1,259 @@
...
+        {
+            for (var i = start; i < source.Length; ++i)
```
davidfowl: IndexOf?

src/Microsoft.Extensions.WebEncoders.Core/PathDecoder.cs
```diff
@@ -0,0 +1,259 @@
...
+            return (expectBytes + 1) * 3;
+        }
+
+        /// <summary>
+        /// Copy characters in source string to target char array. After copied the indices points at the beginning
+        /// offset of both source and destination are updated to point to the position after the last copied character
+        /// </summary>
+        private static int Copy(string source, ref int sourceStart, char[] destination, ref int destinatonStart, int length)
+        {
+            for (var i = 0; i < length; ++i)
```
davidfowl: Array.Copy?

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
```diff
@@ -0,0 +1,326 @@
...
+            {
+                return source.Length;
+            }
+            else
+            {
+                return next;
+            }
+        }
+
+        private static bool TryGetUnescapedAscii(char first, char second, out char result)
```
Tratcher: This name and purpose are out of sync. You don't fail if it's > 0x7F and you keep doing bitwise operation on the resulting char. How about changing it to TryGetUnescapedByte with an out byte?

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
```diff
@@ -0,0 +1,326 @@
...
+    /// </summary>
+    public class UrlPathDecoder
+    {
+        private static Lazy<Encoding> _utf8 = new Lazy<Encoding>(() =>
+        {
+            return Encoding.GetEncoding(
+                "utf-8",
+                new EncoderExceptionFallback(),
+                new DecoderExceptionFallback());
+        }, isThreadSafe: true);
```
halter73: Never seen the `isThreadSafe` parameter used before. Apparently `true` is already the default so this isn't necessary: https://msdn.microsoft.com/en-us/library/dd997286(v=vs.110).aspx

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
```diff
@@ -0,0 +1,326 @@
...
+        /// - It doesn't check if the string contains query
+        /// </summary>
+        public static string Decode(string original)
+        {
+            if (original == null)
+            {
+                throw new ArgumentNullException(nameof(original));
+            }
+
+            if (string.IsNullOrEmpty(original))
```
halter73: Seems like an unnecessary optimization.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
rynowak: You can also just not use Lazy at all. static field initialization is always done under a lock.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
```diff
@@ -0,0 +1,326 @@
...
+                if (remainingBytes == 2 && currentDecodeBits >= 0x110)
+                {
+                    // This is going to be out of the upper Unicode bound 0x10FFFF.
+                    return 0;
+                }
+            }
+
+            if (currentDecodeBits < expectValueMin)
+            {
+                // Overlong encoding (e.g. using 2 bytes to encode something that only needed 1).
```
Tratcher: I think this a place where normalization would be performed.
https://tools.ietf.org/html/rfc3987#section-5.3.2.2

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
```diff
@@ -0,0 +1,326 @@
...
+                return 0;
+            }
+
+            char first;
+            if (!TryGetUnescapedAscii(source[start + 1], source[start + 2], out first))
+            {
+                return 0;
+            }
+
+            if ((first & 0x80) == 0)
```
halter73: `first <= 0x7F` might be a little clearer if that's what you want since a `char` is 16 bits.

halter73: Never mind. I guess this is more consistent with what you have below.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
troydai: I'll change to a `static`. Looks like a over-optimization. The chance that over the lifespan of a server nothing escaped in URL is rare.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
troydai: Will update
- Name
- Out a `byte`

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
troydai: Will go away.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
troydai: Will update.

troydai:
Ok, I make one significant update. I added a function to do inplace decoding to avoid allocating additional char array.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
```diff
@@ -0,0 +1,341 @@
...
+    /// </summary>
+    public class UrlPathDecoder
+    {
+        private static Encoding _utf8 = Encoding.GetEncoding(
+            "utf-8",
+            new EncoderExceptionFallback(),
+            new DecoderExceptionFallback());
+
+        /// <summary>
+        /// Unescape a url char array in place. Returns the length of the result.
```
Tratcher: use standard doc comment tags like `<returns>`

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
```diff
@@ -0,0 +1,341 @@
...
+        /// 
+        /// - Everything is unescaped except %2F ('\')
```
Tratcher: `/`, not `\`

troydai: yup ...

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
```diff
@@ -0,0 +1,341 @@
...
+            var len = DecodeCore(buffer, buffer);
+            return new string(buffer, 0, len);
+        }
+
+        /// <summary>
+        /// Unescape a url path string.
+        /// 
+        /// - The result is saved in place
+        /// </summary>
+        private static int DecodeCore(char[] source, char[] destination)
```
Tratcher: This is only ever called with source == destination. Do you have a reason to maintain the difference?

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
```diff
@@ -0,0 +1,341 @@
...
+        }
+
+        private static bool IsHex(char value)
+        {
+            return (((value >= '0') && (value <= '9')) ||
+                    ((value >= 'A') && (value <= 'F')) ||
+                    ((value >= 'a') && (value <= 'f')));
+        }
+
+        private static int HexToDec(char value)
```
Tratcher: return byte?

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
```diff
@@ -0,0 +1,341 @@
...
+        /// </summary>
+        private static int Unescape(char[] source, int start, byte[] bytesBuffer, ref char[] output, ref int count)
+        {
+            if (start + 2 >= source.Length)
+            {
+                return 0;
+            }
+
+            byte first;
+            if (!TryGetUnescapedByte(source[start + 1], source[start + 2], out first))
```
Tratcher: This is always called with `source[start + 1], source[start + 2]`. Pass in the array and offset instead and do the Length check internally. I see a few places where you may have missed the length check.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
troydai: will do.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
troydai: good point.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
halter73: Do we not expect the path to already be normalized?

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
troydai: This formation make it easier to expand the feature (in the future) when we do need to store the decoded string in a separate place. To achieve so we just need to add an additional `public` function call this method in different form.

A function implemented in this form also helps to improve the readability.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
troydai: On another look I'll not change it. It is consumed by `TryGetUnescapedByte` only and return `int`  avoid casting in the body.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
troydai: I'll leave normalization out of this PR for now.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
halter73: Maybe update the summary: "- The result is saved in place" could be "- source and destination can be the same" instead.

troydai:
Ping @blowdart for security review.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
troydai: The difficult there is although `String.Normalize` can easily normalize a unicode string, it does that by allocating additional memory. I'm seeking approaches to do it in place. Url decode will be on the hot path of the server, I'll try to avoid any unnecessary memory allocate.

Eilon:
Yes, why is this here? This needs to be in CoreFx.

Eilon:
In fact, @Tratcher is in the process of gutting this entire project: https://github.com/aspnet/HttpAbstractions/pull/449

troydai:
Sure, this can be moved to corefx eventually. I'm wondering what's the process of adding this to corefx and how long does it take since I have four days to fix the corresponding Kestrel issue 124. 

May be we can do this parallel. I'll open an PR on corefx begin adding this there in the meantime let's add this here first to get by the rc1.

Eilon:
Make this more internal-ish in this repo then, as opposed to making it public-ish. (I don't care about actual accessibility, but I want it to look like it's going away, e.g. in a `.Internal` namespace.

troydai:
Will do.

src/Microsoft.Extensions.WebEncoders.Core/UrlPathDecoder.cs
lodejard: Path should be decoded, not normalized, by the server so this should be fine.

troydai:
Ping every one: @muratg @Eilon @Tratcher @halter73 @lodejard

src/Microsoft.Extensions.WebEncoders.Core/Internals/UrlPathDecoder.cs
```diff
@@ -0,0 +1,431 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Text;
+
+namespace Microsoft.Extensions.WebEncoders.Internals
```
Eilon: `.Internal`

troydai:
Close this one for now. I'm aiming at handle this entirely in Kestrel by itself.

troydai:
https://github.com/aspnet/KestrelHttpServer/pull/274

Issue #391, opened 2015-09-08
Start using System.Text.Encoding.Web from CLR


muratg:

Eilon:
This is now available on the internal share that we use, so this is unblocked.

muratg:
@Tratcher

Tratcher:
@Eilon @muratg 4.0.0-beta-23326 is not compatible with .NET 4.5.1, it requires 4.6. It depends on System.Runtime 4.0.20, where we need 4.0.10 for 4.5.1.

Tratcher:
@ericstj ?

ericstj:
It's compiling against quite a few contracts that are only supported in .NET 4.6.  

I gave a try at moving it to contract versions that would support .NET 4.5.1 and it is not possible because the implementation [depends on Buffer.MemoryCopy](https://github.com/dotnet/corefx/blob/9ce81a8b75df175987c3dbb00cfca2239ce1178e/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/TextEncoder.cs#L89) which was added in .NET 4.6. This appears to be the only dependency that is preventing portability to even to .NET 4.5.  @KrzysztofCwalina

muratg:
@ericstj thanks, @KrzysztofCwalina Should we file a bug on your side to track this?  Otherwise we'll need to maintain our implementation at least for Desktop CLR.

KrzysztofCwalina:
I will fix this tomorrow. No need to file a new bug.

muratg:
@KrzysztofCwalina Sounds good, thanks!

davidfowl:
Ifdefs will set you free

KrzysztofCwalina:
@ericstj, most of our projects in corefx depend on System.Runtime 4.0.20.

ericstj:
That's because most of these don't intend to run downlevel.  There are a few that have downlevel scenarios (like Immutable & Reflection.Metadata) and intentionally restrict their dependencies to versions that will work on those downlevel platforms.

KrzysztofCwalina:
@ericstj, almost all (if not all) components in corefx depend on 4.0.20. And it seems like ASP.NET needs to and is successfully using them on 4.5.1 (i.e. downlevel)

Eilon:
Right, all of DNX, ASP.NET, and EF (with maybe some very small exceptions, where it's usually runtime light-up) need to run on .NET 4.5.1.

[...]

PR #449, opened 2015-10-20
Migrate to System.Text.Encoding.Web

Tratcher:
#391 @davidfowl @muratg @Eilon 
David asked me to keep the interfaces for DI, so I replaced the encoders with wrappers. Components that were not using the interfaces I change to directly use the BCL types.


src/Microsoft.Extensions.WebEncoders.Core/HtmlEncoderWrapper.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        public static HtmlEncoderWrapper Default { get; } = new HtmlEncoderWrapper(HtmlEncoder.Default);
+
+        public string HtmlEncode(string value)
+        {
+            return _encoder.Encode(value);
+        }
+
+        public void HtmlEncode(string value, int startIndex, int charCount, TextWriter output)
```
Tratcher: Should we change these method names and parameter ordering to match the new BCL APIs? It will make it easier for people to switch between our DI wrapper and the underlying BCL types, but it's more of a break for us.

muratg: Having parity between the wrapper and BCL types makes sense to me. How much work would it be to do it?

Tratcher:
@muratg Updated with the new names and signatures. It will make the down stream breaks much worse, but is probably better in the long run.

Tratcher:
+@rynowak who does not like wrappers.

muratg:
Looks good to me.

Tratcher:
@davidfowl Removing the wrappers mostly works, except that it's much harder to mock and test because the needed methods aren't virtual.
https://github.com/aspnet/HttpAbstractions/commit/90c13adc5d90108a4826af3b5149a3e923b79436
Compare the old CommonTestEncoder (https://github.com/aspnet/HttpAbstractions/commit/90c13adc5d90108a4826af3b5149a3e923b79436#diff-5093aee64e24c262ac1165902a6ceef9) with the new HtmlTestEncoder (https://github.com/aspnet/HttpAbstractions/commit/90c13adc5d90108a4826af3b5149a3e923b79436#diff-387f3dc26a94ebcb7c37dd754bab84c5)

Tratcher:
https://github.com/dotnet/corefx/pull/4038

src/Microsoft.AspNet.Http.Abstractions/project.json
```diff
@@ -11,13 +11,19 @@
...
+    "dnx451": {
+      "frameworkAssemblies": {
+        "System.IO": "",
+        "System.Runtime": ""
+      }
+    },
     "dnxcore50": {
       "dependencies": {
         "System.Collections": "4.0.11-beta-*",
+        "System.ComponentModel": "4.0.1-beta-*",
```
Eilon: Hmm, is this actually used? I'm curious for what...

Eilon:
:shipit:

src/Microsoft.AspNet.Http.Abstractions/QueryString.cs
```diff
@@ -154,13 +163,23 @@ public static QueryString Create(IEnumerable<KeyValuePair<string, StringValues>>
             bool first = true;
             foreach (var pair in parameters)
             {
+                if (pair.Key == null)
+                {
+                    throw new ArgumentNullException(nameof(pair.Key));
```
muratg: `InvalidOperationException` as this is not an arg.

src/Microsoft.AspNet.Http.Abstractions/QueryString.cs
```diff
@@ -154,13 +163,23 @@ public static QueryString Create(IEnumerable<KeyValuePair<string, StringValues>>
...
+                }
+
                 foreach (var value in pair.Value)
                 {
+                    if (value == null)
+                    {
+                        throw new ArgumentNullException(nameof(value));
```
muratg: Ditto.

src/Microsoft.AspNet.Http.Abstractions/project.json
Tratcher: System.Text.Encodings.Web seems to require it.

Eilon:
Clarification: My :shipit: was contingent on an offline discussion regarding keeping the interfaces. I don't think it's bad either way, but I prefer the nice interfaces.

src/Microsoft.Extensions.WebEncoders.Core/project.json
```diff
@@ -1,27 +1,33 @@
...
+    "type": "git",
+    "url": "git://github.com/aspnet/httpabstractions"
+  },
+  "compilationOptions": {
+    "warningsAsErrors": true
+  },
+  "dependencies": {
+    "System.Text.Encodings.Web": "4.0.0-beta-*"
+  },
+  "frameworks": {
```
davidfowl: These need to be updated to use generations now

src/Microsoft.Extensions.WebEncoders.Core/UrlEncoderWrapper.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            _encoder = encoder;
+        }
+
+        public UrlEncoderWrapper(TextEncoderSettings settings)
+        {
+            _encoder = UrlEncoder.Create(settings);
+        }
+
+        public static IUrlEncoder Default { get; } = new UrlEncoderWrapper(UrlEncoder.Default);
```
davidfowl: Do we still need Default? People using this should just use the BCL type right?

Tratcher: This default is for plugging into places that take the interface, see https://github.com/aspnet/HttpAbstractions/pull/449/files#diff-af79926afbc0631c7212bff6ed282bc0R208

src/Microsoft.AspNet.Http.Abstractions/project.json
Eilon: Won't the transitive dependency solve that?

davidfowl: Should...

Tratcher:
Rebased and updated. Now that we have the right virtuals we can remove the interfaces and the entire WebEncoders.Core package.

CommonTestEncoder is also replaced by HtmlTestEncoder, UrlTestEncoder, and JavaScriptTestEncoder, provided directly in the WebEncoders package.

Eilon:
I miss the interfaces, but :shipit:

PR #453, opened 2015-10-29
Add ForbidAsync overload for automatic authentication scheme.

pranavkm:
cc @HaoK


src/Microsoft.AspNet.Http.Abstractions/Authentication/AuthenticationManager.cs
```diff
@@ -99,6 +99,14 @@ public virtual Task ForbidAsync(string authenticationScheme, AuthenticationPrope
             return ChallengeAsync(authenticationScheme, properties, ChallengeBehavior.Forbidden);
         }
 
+        /// <summary>
+        /// Creates a challenge for the authentication manager with a <see cref="ChallengeBehavior.Forbidden"/>.
+        /// </summary>
+        /// <param name="properties">Additional arbitrary values which may be used by particular authentication types.</param>
+        /// <returns>A <see cref="Task"/> that represents the asynchronous challenge operation.</returns>
+        public virtual Task ForbidAsync(AuthenticationProperties properties)
```
Tratcher: Should there also be ForbidAsync with no params?

HaoK: Yeah if there is one with just properties. The empty one is likely to be more useful.

pranavkm: Updated

HaoK:
:shipit:

PR #458, opened 2015-10-31
Restore `null` and `string.Empty` handling from `EncoderExtensions`

dougbu:
- prior test encoders were never invoked for `null` or empty `string`s e.g.
``` c#
    public static void HtmlEncode(this IHtmlEncoder htmlEncoder, string value, TextWriter output)
    {
...
        if (!String.IsNullOrEmpty(value))
        {
            htmlEncoder.HtmlEncode(value, 0, value.Length, output);
        }
    }
```


src/Microsoft.Extensions.WebEncoders/Testing/HtmlTestEncoder.cs
```diff
@@ -18,6 +18,11 @@ public override int MaxOutputCharactersPerInputCharacter
 
         public override string Encode(string value)
         {
+            if (string.IsNullOrEmpty(value))
```
Tratcher: Careful, you've moved this behavior from production code to test code and you're now mis-representing the actual encoder's behaviors for null. https://github.com/dotnet/corefx/blob/c02d33b18398199f6acc17d375dab154e9a1df66/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/TextEncoder.cs#L44

dougbu: Good point and I will certainly add the missing `null` check. Same for the other overloads in this class.

Turns out I need to check for `string.Empty` in the `Encode(TextWriter output, string value, int startIndex, int characterCount)` overload as well. [`TextEncoder.Encode(TextWriter output, string value)`](https://github.com/dotnet/corefx/blob/master/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/TextEncoder.cs#L161) calls that one.

dougbu:
:up::date:

Tratcher:
:shipit:

dougbu:
Thanks @Tratcher Just waiting for aspnet/Mvc#3453 opinions. Also slightly prefer to get this in after CI looks greener.

src/Microsoft.Extensions.WebEncoders/Testing/HtmlTestEncoder.cs
```diff
@@ -18,11 +19,31 @@ public override int MaxOutputCharactersPerInputCharacter
 
         public override string Encode(string value)
         {
+            if (value == null)
+            {
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            if (string.IsNullOrEmpty(value))
```
Eilon: Why checking for null twice?

Eilon: (Applies to many methods here)

dougbu: `IsNullOrEmpty()` is heavily optimized, `value == string.Empty` often gets the opposite code review comments, and `IsNullOrEmpty()` is what we use everywhere even for known non-`null` values.

So, still prefer `value == string.Empty`?

Eilon: We never compare strings to `string.Empty`. Why not check `.Length == 0`?

dougbu: :ok:

dougbu:
:up::date: @Eilon

test/Microsoft.Extensions.WebEncoders.Tests/HtmlTestEncoderTest.cs
```diff
@@ -0,0 +1,27 @@
...
+        [InlineData("abcd", "HtmlEncode[[abcd]]")]
+        [InlineData("<<''\"\">>", "HtmlEncode[[<<''\"\">>]]")]
+        public void StringEncode_EncodesAsExpected(string input, string expectedOutput)
+        {
+            // Arrange
+            var encoder = new HtmlTestEncoder();
+
+            // Act
+            var output = encoder.Encode(input);
+            encoder.Encode(new System.IO.StringWriter(), "");
```
Eilon: What's this line for?

dougbu: Garbage that I was playing with. Won't be in when I squish the commits.

dougbu:
Slight :up::date: to remove garbage.

test/Microsoft.Extensions.WebEncoders.Tests/HtmlTestEncoderTest.cs
Eilon: Oh nice :smile:

Eilon:
:shipit:

src/Microsoft.Extensions.WebEncoders/Testing/HtmlTestEncoder.cs
```diff
@@ -30,6 +51,21 @@ public override void Encode(TextWriter output, char[] value, int startIndex, int
...
+            {
+                throw new ArgumentNullException(nameof(output));
+            }
+
+            if (value == null)
+            {
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            if (value.Length == 0)
```
Tratcher: `|| chacterCount == 0`? Same in the other overloads.

dougbu: That raises another issue: We're not currently validating the `startIndex` and `characterCount` values as the product code would. And this check could short-circuit those validations. One sec while I fix it...

Eilon: Wait - we don't want to duplicate like a million lines of validation logic. The obscure bad parameters should just fail at the inner method.

dougbu: So just `characterCount == 0` and call it a day i.e. let `Substring()` (the only "inner method" here) yell as it will?

Eilon: These are just the test encoders. As long as they roughly behave like the real ones, except for not actually encoding anything, I don't think it matters that much.

dougbu:
:up::date: FYI. Will push shortly.

src/Microsoft.Extensions.WebEncoders/Testing/HtmlTestEncoder.cs
dougbu: Done. Note this particular comment wasn't hidden because there's no `characterCount` parameter here.

dougbu:
037196d

PR #460, opened 2015-11-03
String performance improvements on top of #411

NickCraver:
This adds additional performance improvements (namely string.Concat overloads) for a few cases. It also eliminates an extra check inside QueryHelper.

### DIFF
--- a/src/Microsoft.AspNet.Http.Abstractions/FragmentString.cs
+++ b/src/Microsoft.AspNet.Http.Abstractions/FragmentString.cs
@@ -100,7 +100,7 @@ namespace Microsoft.AspNet.Http
             string fragmentValue = uri.GetComponents(UriComponents.Fragment, UriFormat.UriEscaped);
             if (!string.IsNullOrEmpty(fragmentValue))
             {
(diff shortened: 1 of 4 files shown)


davidfowl:
W00p!

Tratcher:
As this is performance related, please share your relevant measurements.

NickCraver:
@Tratcher sorry! I took for granted this was a more widely known thing - `string.Concat()` has overloads for up to 4 strings and it's substantially faster than `string.Format()` for these cases. I've put up some quick benchmarks here as a sample: https://gist.github.com/NickCraver/fce9754c50aa7a4c9593

halter73:
:shipit:

benaadams:
> string.Concat() has overloads for up to 4 strings and it's substantially faster than string.Format() for these cases.

Can't disagree; but feels like coreclr or rosyln is missing a trick here; or maybe its that they aren't missing the trick with `.Concat` :D

khellang:
Is there an existing issue for optimizing these types of interpolated strings (basically concatenating strings) in the Roslyn repo?

PR #461, opened 2015-11-03
C#6 Cleanup & Optimizations

NickCraver:
The main intent is cleanup using C# 6 operators and normalization of type aliases. While there are no intended functional changes here, it does eliminate a few tight race conditions as a bonus (not a real-win since this isn't thread-safe all over, simply noting).

### DIFF
--- a/src/Microsoft.AspNet.Http.Abstractions/QueryString.cs
+++ b/src/Microsoft.AspNet.Http.Abstractions/QueryString.cs
@@ -17,7 +17,7 @@ namespace Microsoft.AspNet.Http
         /// <summary>
         /// Represents the empty query string. This field is read-only.
         /// </summary>
-        public static readonly QueryString Empty = new QueryString(String.Empty);
+        public static readonly QueryString Empty = new QueryString(string.Empty);
 
         private readonly string _value;
 
@@ -84,7 +84,7 @@ namespace Microsoft.AspNet.Http
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1057:StringUriOverloadsCallSystemUriOverloads", Justification = "Delimiter characters ? and # must be escaped by this method instead of truncating the value")]
         public static QueryString FromUriComponent(string uriComponent)
         {
-            if (String.IsNullOrEmpty(uriComponent))
+            if (string.IsNullOrEmpty(uriComponent))
             {
                 return new QueryString(string.Empty);
             }
@@ -231,7 +231,7 @@ namespace Microsoft.AspNet.Http
 
         public override int GetHashCode()
         {
-            return (_value != null ? _value.GetHashCode() : 0);
+            return _value?.GetHashCode() ?? 0;
         }
 
         public static bool operator ==(QueryString left, QueryString right)
--- a/src/Microsoft.AspNet.Http/HttpContextAccessor.cs
+++ b/src/Microsoft.AspNet.Http/HttpContextAccessor.cs
@@ -21,7 +21,7 @@ namespace Microsoft.AspNet.Http.Internal
             get
             {
                 var handle = CallContext.LogicalGetData(LogicalDataKey) as ObjectHandle;
-                return handle != null ? handle.Unwrap() as HttpContext : null;
+                return handle?.Unwrap() as HttpContext;
             }
             set
             {
--- a/src/Microsoft.Net.Http.Headers/CacheControlHeaderValue.cs
+++ b/src/Microsoft.Net.Http.Headers/CacheControlHeaderValue.cs
@@ -184,7 +184,7 @@ namespace Microsoft.Net.Http.Headers
             if (_noCache)
             {
                 AppendValueWithSeparatorIfRequired(sb, NoCacheString);
-                if (_noCacheHeaders?.Count > 0)
+                if ((_noCacheHeaders != null) && (_noCacheHeaders.Count > 0))
                 {
                     sb.Append("=\"");
                     AppendValues(sb, _noCacheHeaders);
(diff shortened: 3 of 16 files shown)


NickCraver:
There's also ample usage to migrate properties to expression bodies (`=>`) (or `{get;}`-only autos) in the codebase, but I wasn't sure if this would be desired - it's a matter of taste. It'd shave of quite a bit of code on the simpler accessors or overloads for returning value types. If that's a welcome change, I can do another PR.

khellang:
Nice :shipit:

src/Microsoft.Net.Http.Headers/CacheControlHeaderValue.cs
```diff
@@ -184,7 +184,7 @@ public override string ToString()
             if (_noCache)
             {
                 AppendValueWithSeparatorIfRequired(sb, NoCacheString);
-                if ((_noCacheHeaders != null) && (_noCacheHeaders.Count > 0))
+                if (_noCacheHeaders?.Count > 0)
```
pranavkm: This would box the int. Could we leave this as is?

test/Microsoft.Net.Http.Headers.Tests/CookieHeaderValueTest.cs
```diff
@@ -154,7 +154,7 @@ public void CookieHeaderValue_Ctor2InitializesCorrectly(string name, string valu
         public void CookieHeaderValue_Value()
         {
             var cookie = new CookieHeaderValue("name");
-            Assert.Equal(String.Empty, cookie.Value);
+            Assert.Equal(string.Empty, cookie.Value);
```
pranavkm: `Assert.Empty(cookie.Value);`

src/Microsoft.AspNet.Http/HttpContextAccessor.cs
```diff
@@ -21,7 +21,7 @@ public HttpContext HttpContext
             get
             {
                 var handle = CallContext.LogicalGetData(LogicalDataKey) as ObjectHandle;
-                return handle != null ? handle.Unwrap() as HttpContext : null;
+                return handle?.Unwrap() as HttpContext;
```
pranavkm: Should this just cast?

src/Microsoft.Net.Http.Headers/CacheControlHeaderValue.cs
NickCraver: Doh, fair point. Mental lapse on these - reverting the CacheControlHeaders changes. There is a chance for better Roslyn behavior here with the syntax sugar to prevent the boxing in many cases such as this though...

test/Microsoft.Net.Http.Headers.Tests/CookieHeaderValueTest.cs
NickCraver: Is that what's intended here? IMO using the IEnumerable overload when checking strings isn't the most intuitive.

src/Microsoft.AspNet.Http/HttpContextAccessor.cs
NickCraver: In theory, someone could put whatever in there - the `as` is a bit safer still.

test/Microsoft.Net.Http.Headers.Tests/CookieHeaderValueTest.cs
pranavkm: We use it elsewhere in our tests. I'd leave it up to you since it doesn't change the intent.

src/Microsoft.AspNet.Http.Abstractions/QueryString.cs
```diff
@@ -231,7 +231,7 @@ public override bool Equals(object obj)
 
         public override int GetHashCode()
         {
-            return (_value != null ? _value.GetHashCode() : 0);
+            return _value?.GetHashCode() ?? 0;
```
pranavkm: Boxing again.

Tratcher:
@benaadams are you following this one?

src/Microsoft.AspNet.Http.Abstractions/QueryString.cs
benaadams: Does this box? Nullable<int> is struct? Will do more work though

benaadams: Hmm might do; unless it doesn't converts to:
```csharp
var nullable = _value?.GetHashCode();
return nullable.HasValue ? nullable.Value : 0;
```
will check the cil

benaadams: Pre:
```
.method public hidebysig virtual instance int32 
        GetHashCode() cil managed
{
  // Code size       22 (0x16)
  .maxstack  8
  IL_0000:  ldarg.0
  IL_0001:  ldfld      string Microsoft.AspNet.Http.QueryString::_value
  IL_0006:  brtrue.s   IL_000a
  IL_0008:  ldc.i4.0
  IL_0009:  ret
  IL_000a:  ldarg.0
  IL_000b:  ldfld      string Microsoft.AspNet.Http.QueryString::_value
  IL_0010:  callvirt   instance int32 [System.Runtime]System.Object::GetHashCode()
  IL_0015:  ret
} // end of method QueryString::GetHashCode
```
Post
```
.method public hidebysig virtual instance int32 
        GetHashCode() cil managed
{
  // Code size       18 (0x12)
  .maxstack  8
  IL_0000:  ldarg.0
  IL_0001:  ldfld      string Microsoft.AspNet.Http.QueryString::_value
  IL_0006:  dup
  IL_0007:  brtrue.s   IL_000c
  IL_0009:  pop
  IL_000a:  ldc.i4.0
  IL_000b:  ret
  IL_000c:  callvirt   instance int32 [System.Runtime]System.Object::GetHashCode()
  IL_0011:  ret
} // end of method QueryString::GetHashCode
```

pranavkm: It does look like it's doing what you listed. Nevermind, I guess I was just confused about how null coalescing works for nullable types.

benaadams: Was just saying how it could work; never actually looked at what it did do in depth before

pranavkm: :+1:

benaadams:
LGTM

Tratcher:
:shipit:

Tratcher:
Rebased, squashed, and merged.

PR #463, opened 2015-11-04
Revert use of explicit converters that prevent APIs from returning null.

Tratcher:
Reverting the use of explicit type converters (ToString and ToArray) that were preventing some APIs from returning null. The implicit converters can return null.

See https://github.com/aspnet/Common/blob/dev/src/Microsoft.Extensions.Primitives/StringValues.cs#L85
https://github.com/aspnet/Common/blob/dev/src/Microsoft.Extensions.Primitives/StringValues.cs#L104

@rynowak 
/cc @benaadams


Tratcher:
Note this is breaking several MVC tests due to ContentType not returning null.

Tratcher:
Updated with tests.

Tratcher:
@BrennanConroy

Kagamine:
LGTM

JunTaoLuo:
:shipit:

NickCraver:
+1 - Thanks for changing these up.

PR #465, opened 2015-11-06
Reduce allocations on Conneg hotpath

rynowak:
**Before**
image
image

**After**
image

(enumerator allocations totally gone)


src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -13,7 +13,7 @@ public static class HeaderUtilities
         private const string QualityName = "q";
         internal const string BytesUnit = "bytes";
 
-        internal static void SetQuality(ICollection<NameValueHeaderValue> parameters, double? value)
+        internal static void SetQuality(ObjectCollection<NameValueHeaderValue> parameters, double? value)
```
rynowak: The main pivot here is around what to do in this PR with `ObjectCollection<T>`. The main issue is that `ICollection<T>` is a **loser interface** and has no friends.

We could do something else simple here like **inherit** from `Collection<T>` and pass this around as `IList<T>`. Then we could use the indexer to avoid allocations. That would remove an object allocation per `ObjectCollection<T>` that we created.

I really just kind of #YOLOed this change because `ObjectCollection<T>` is internal. Let's discuss more tomorrow if needed.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -457,17 +485,17 @@ private static int GetMediaTypeExpressionLength(string input, int startIndex, ou
...
+            var mediaTypeLength = current + subtypeLength - startIndex;
+            if (typeLength + subtypeLength + 1 == mediaTypeLength)
             {
-                mediaType = input.Substring(startIndex, mediatTypeLength);
+                mediaType = input.Substring(startIndex, mediaTypeLength);
             }
             else
             {
                 mediaType = input.Substring(startIndex, typeLength) + "/" + input.Substring(current, subtypeLength);
             }
```
rynowak: I didn't try to remove these substrings as part of this change. If we wanted to re-engineer all of this code around `StringSegment` we could get rid of them too.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -320,6 +320,25 @@ private static int GetNameValueLength(string input, int startIndex, out NameValu
...
+
+            foreach (var value in values)
+            {
+                if (string.Compare(value.Name, name, StringComparison.OrdinalIgnoreCase) == 0)
+                {
+                    return value;
+                }
+            }
+            return null;
+        }
```
rynowak: Had to dupe this because the existing thing is public. If we want to base this on `IList<T>` then this isn't needed.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -210,7 +222,7 @@ public bool MatchesAllSubTypes
         {
             get
             {
-                return SubType.Equals("*", StringComparison.Ordinal);
+                return string.Compare(_mediaType, _mediaType.IndexOf('/') + 1, "*", 0, 1, StringComparison.Ordinal) == 0;
```
davidfowl: ?

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -53,13 +53,13 @@ public string Charset
                 HeaderUtilities.ThrowIfReadOnly(IsReadOnly);
                 // We don't prevent a user from setting whitespace-only charsets. Like we can't prevent a user from
                 // setting a non-existing charset.
-                var charsetParameter = NameValueHeaderValue.Find(_parameters, CharsetString);
+                var charsetParameter = NameValueHeaderValue.Find(ParametersInternal, CharsetString);
```
dougbu: `Find()` returns `null` if the first parameter is `null`. So this could now unnecessarily allocate an `ObjectCollection` instance.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -53,13 +53,13 @@ public string Charset
...
                 if (string.IsNullOrEmpty(value))
```
dougbu: Put another way, setting `Charset` (or `Boundary`) to `null` should just leave `ParametersInternal` `null` if possible.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -320,6 +320,25 @@ private static int GetNameValueLength(string input, int startIndex, out NameValu
             }
         }
 
+        internal static NameValueHeaderValue Find(ObjectCollection<NameValueHeaderValue> values, string name)
+        {
+            Contract.Requires((name != null) && (name.Length > 0));
```
dougbu: Thought we were avoiding `Contract` assertions?

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -320,6 +320,25 @@ private static int GetNameValueLength(string input, int startIndex, out NameValu
...
+
+            if ((values == null) || (values.Count == 0))
+            {
+                return null;
+            }
+
+            foreach (var value in values)
+            {
+                if (string.Compare(value.Name, name, StringComparison.OrdinalIgnoreCase) == 0)
```
dougbu: Curious: Is this not `string.Equals()` primarily for consistency with some of the comparisons that pass more arguments to `Compare()`?

src/Microsoft.Net.Http.Headers/ObjectCollection.cs
```diff
@@ -88,5 +90,49 @@ private static void CheckNotNull(T item)
...
+        {
+            private readonly ObjectCollection<T> _collection;
+            private int _index;
+
+            public Enumerator(ObjectCollection<T> collection)
+            {
+                _collection = collection;
+
+                _index = -1;
+                Current = default(T);
```
dougbu: Suggest calling `Reset()` here.

dougbu:
:watch:

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
benaadams: Is this for Accept/Language type headers? With an ordering precedent? If so `IList<T>` would make a lot of sense; but then you'd also want it in order of the `;q=` bit if specified; otherwise parse order

rynowak: Yeah, looking at the impact of these changes as a whole, `IList` seems like a better fit.

benaadams: Non-ordered stuff though `IList` would be bad; as its harder to get an access by indexer out of an api that it would be to change the jit to understand `foreach` enumerators via interface without boxing; and since that hasn't happened yet its probably quite hard ;-)

benaadams: Lol, nvm its a strongly typed collection, so an `IList` would be more flexible that current :+1:

rynowak:
Updated this to inherit from `Collection<T>` and use `IList<T>` @Tratcher thoughts?

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -59,7 +59,7 @@ public string Charset
                     // Remove charset parameter
                     if (charsetParameter != null)
                     {
-                        _parameters.Remove(charsetParameter);
+                        Parameters.Remove(charsetParameter);
```
Tratcher: More inconsistency.

dougbu: Slight inconsistency doesn't matter that much. Impossible for this line to be reached if `_parameters` is `null` and the potential allocation at [line 73](https://github.com/aspnet/HttpAbstractions/pull/465/files#diff-afd3378d76454b61ea0cdd22e56685fdR73) is required. I somewhat prefer @rynowak's reduction in `_parameters` use.

Tratcher:
:shipit:

davidfowl:
Why is the ci red?

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -218,9 +218,9 @@ internal static int GetHashCode(ICollection<NameValueHeaderValue> values)
             }
 
             var result = 0;
```
benaadams: Microsoft.Extensions.HashCodeCombiner

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
rynowak: It's intended to avoid allocating the `Parameters` collection unless we need to mutate it.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
rynowak: This code already has like 5 other hash functions. I don't want to change one without changing the others, and I don't want to puzzle out all of the knots in the others with no real clear reason to change them. This was ported from an exiting vOld codebase.

Issue #398, opened 2015-09-14
UseBuilder should take optional Action<IServiceCollection> 


davidfowl:
To make it easier to configure services in the ASP.NET 5 pipeline when exposed via Owin

muratg:
Won't the overload with `Action<IApplicationBuilder>` be enough for that? It has a `ServiceProvider` property.

Tratcher:
Except ServiceProvider is null: https://github.com/aspnet/HttpAbstractions/blob/f475e53ad235478b5ac35bab31a72e6cafb52156/src/Microsoft.AspNet.Owin/OwinExtensions.cs#L72

JunTaoLuo:
merged to dev

PR #468, opened 2015-11-09
Adding option to configure services when exposing the ASP.NET 5 pipel…

JunTaoLuo:
…ine via OWIN #398

### DIFF
--- a/src/Microsoft.AspNet.Owin/OwinExtensions.cs
+++ b/src/Microsoft.AspNet.Owin/OwinExtensions.cs
@@ -67,21 +67,26 @@ namespace Microsoft.AspNet.Builder
 
         public static IApplicationBuilder UseBuilder(this AddMiddleware app)
         {
+            return app.UseBuilder(serviceProvider: null);
+        }
+
+        public static IApplicationBuilder UseBuilder(this AddMiddleware app, IServiceProvider serviceProvider)
+        {
             // Adapt WebSockets by default.
             app(OwinWebSocketAcceptAdapter.AdaptWebSockets);
-            var builder = new ApplicationBuilder(serviceProvider: null);
+            var builder = new ApplicationBuilder(serviceProvider: serviceProvider);
 
-            CreateMiddleware middleware = CreateMiddlewareFactory(exit =>
+            var middleware = CreateMiddlewareFactory(exit =>
             {
                 builder.Use(ignored => exit);
                 return builder.Build();
-            });
+            }, builder.ApplicationServices);
 
             app(middleware);
             return builder;
         }
 
-        private static CreateMiddleware CreateMiddlewareFactory(Func<RequestDelegate, RequestDelegate> middleware)
+        private static CreateMiddleware CreateMiddlewareFactory(Func<RequestDelegate, RequestDelegate> middleware, IServiceProvider applicationServices)
         {
             return next =>
             {
@@ -105,6 +110,7 @@ namespace Microsoft.AspNet.Builder
                         context = new DefaultHttpContext(
                                     new FeatureCollection(
                                         new OwinFeatureCollection(env)));
+                        context.ApplicationServices = applicationServices;
                     }
 
                     return app.Invoke(context);
@@ -114,7 +120,12 @@ namespace Microsoft.AspNet.Builder
 
         public static AddMiddleware UseBuilder(this AddMiddleware app, Action<IApplicationBuilder> pipeline)
         {
-            var builder = app.UseBuilder();
+            return app.UseBuilder(pipeline, serviceProvider: null);
+        }
+
+        public static AddMiddleware UseBuilder(this AddMiddleware app, Action<IApplicationBuilder> pipeline, IServiceProvider serviceProvider)
+        {
+            var builder = app.UseBuilder(serviceProvider);
             pipeline(builder);
             return app;
         }
--- a/test/Microsoft.AspNet.Owin.Tests/OwinExtensionTests.cs
+++ b/test/Microsoft.AspNet.Owin.Tests/OwinExtensionTests.cs
@@ -1,0 +1,87 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Builder;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Microsoft.AspNet.Owin
+{
+    using AppFunc = Func<IDictionary<string, object>, Task>;
+    using CreateMiddleware = Func<
+          Func<IDictionary<string, object>, Task>,
+          Func<IDictionary<string, object>, Task>
+        >;
+    using AddMiddleware = Action<Func<
+          Func<IDictionary<string, object>, Task>,
+          Func<IDictionary<string, object>, Task>
+        >>;
+
+    public class OwinExtensionTests
+    {
+        static AppFunc notFound = env => new Task(() => { env["owin.ResponseStatusCode"] = 404; });
+
+        [Fact]
+        public void OwinConfigureServiceProviderAddsServices()
+        {
+            var list = new List<CreateMiddleware>();
+            AddMiddleware build = list.Add;
+            IServiceProvider serviceProvider = null;
+            FakeService fakeService = null;
+
+            var builder = build.UseBuilder(applicationBuilder =>
+            {
+                serviceProvider = applicationBuilder.ApplicationServices;
+                applicationBuilder.Run(async context =>
+                {
+                    fakeService = context.ApplicationServices.GetService<FakeService>();
+                });
+            }, new ServiceCollection().AddSingleton(new FakeService()).BuildServiceProvider());
+
+            list.Reverse();
+            list.Aggregate(notFound, (next, middleware) => middleware(next)).Invoke(new Dictionary<string, object>());
+
+            Assert.NotNull(fakeService);
+            Assert.NotNull(serviceProvider.GetService<FakeService>());
+        }
+
+        [Fact]
+        public void OwinDefaultNoServices()
+        {
+            var list = new List<CreateMiddleware>();
+            AddMiddleware build = list.Add;
+            IServiceProvider serviceProvider = null;
+            FakeService fakeService = null;
+            bool builderExecuted = false;
+            bool applicationExecuted = false;
+
+            var builder = build.UseBuilder(applicationBuilder =>
+            {
+                builderExecuted = true;
+                serviceProvider = applicationBuilder.ApplicationServices;
+                applicationBuilder.Run(async context =>
+                {
+                    applicationExecuted = true;
+                    fakeService = context.ApplicationServices.GetService<FakeService>();
+                });
+            });
+
+            list.Reverse();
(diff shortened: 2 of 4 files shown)


JunTaoLuo:
@Tratcher @davidfowl

test/Microsoft.AspNet.Owin.Tests/OwinExtensionTests.cs
```diff
@@ -0,0 +1,48 @@
...
+            var builder = build.UseBuilder(applicationBuilder =>
+            {
+                serviceProvider = applicationBuilder.ApplicationServices;
+            }
+            , serviceCollection =>
+            {
+                serviceCollection.AddInstance(new FakeService());
+            });
+
+            Assert.NotNull(serviceProvider);
```
Tratcher: You need to assert that you can actually resolve the FakeService

src/Microsoft.AspNet.Owin/OwinExtensions.cs
```diff
@@ -105,6 +112,7 @@ private static CreateMiddleware CreateMiddlewareFactory(Func<RequestDelegate, Re
                         context = new DefaultHttpContext(
                                     new FeatureCollection(
                                         new OwinFeatureCollection(env)));
+                        context.ApplicationServices = applicationServices;
```
Tratcher: Needs a test for this scenario too.

JunTaoLuo:
:up::date:

Tratcher:
:shipit:

src/Microsoft.AspNet.Owin/OwinExtensions.cs
```diff
@@ -65,23 +66,29 @@ public static IApplicationBuilder UseOwin(this IApplicationBuilder builder, Acti
             return builder;
         }
 
-        public static IApplicationBuilder UseBuilder(this AddMiddleware app)
+        public static IApplicationBuilder UseBuilder(this AddMiddleware app, Action<IServiceCollection> configureServices = null)
```
davidfowl: No optional params. Add 2 overloads

JunTaoLuo:
:up::date:

src/Microsoft.AspNet.Owin/project.json
```diff
@@ -10,7 +10,8 @@
     "keyFile": "../../tools/Key.snk"
   },
   "dependencies": {
-    "Microsoft.AspNet.Http": "1.0.0-*"
+    "Microsoft.AspNet.Http": "1.0.0-*",
+    "Microsoft.Extensions.DependencyInjection": "1.0.0-*"
```
davidfowl: Microsoft.Extensions.DependencyInjection.Abstractions

JunTaoLuo: We need the implementation `ServiceCollection` from `DependencyInjection`. That is not available in `DependencyInject.Abstractions`.

src/Microsoft.AspNet.Owin/OwinExtensions.cs
```diff
@@ -67,21 +68,32 @@ public static IApplicationBuilder UseOwin(this IApplicationBuilder builder, Acti
 
         public static IApplicationBuilder UseBuilder(this AddMiddleware app)
         {
+            return app.UseBuilder(configureServices: null);
+        }
+
+        public static IApplicationBuilder UseBuilder(this AddMiddleware app, Action<IServiceCollection> configureServices)
```
pranavkm: Shouldn't this return a `IServiceProvider` i.e. `Func<IServiceProvider>`? How would you configure a different IoC container?

JunTaoLuo: I'm not understanding your concern, what do you want to configure exactly? There is the `ApplicationServices` property on the `IApplicationBuilder` from which you can access the application's service provider. The intended use of these extensions are illustrated here https://github.com/aspnet/Entropy/pull/72

pranavkm: The current approach forces us to depend on our implementation of DI. If we let the user return a `IServiceProvider`, they can configure and return a container of their choice. It would also remove the need to depend on `Microsoft.Extensions.DependencyInjection` and instead depend on `Microsoft.Extensions.DependencyInjection.Abstractions`.

davidfowl: Ah yes we should support both overloads.

pranavkm: Doesn't having the `Action<IServiceCollection>` result in bad layering (by forcing `HttpAbstractions` to rely on DI implementation)?

Tratcher: Just for the Owin package, it doesn't affect the rest of HttpAbstractions.

pranavkm: Sure, it's still bad layering at the expense of one extra line of user code. Doesn't seem necessary.

Tratcher: That dependency already exists. HttpAbstractions -> Options -> DI.

pranavkm: Isn't that DI.Abstractions? https://github.com/aspnet/Options/blob/dev/src/Microsoft.Extensions.OptionsModel/project.json#L11

src/Microsoft.AspNet.Owin/OwinExtensions.cs
```diff
@@ -67,21 +68,36 @@ public static IApplicationBuilder UseOwin(this IApplicationBuilder builder, Acti
 
         public static IApplicationBuilder UseBuilder(this AddMiddleware app)
         {
+            return app.UseBuilder(configureServiceProvider: null);
+        }
+
+        public static IApplicationBuilder UseBuilder(this AddMiddleware app, Action<IServiceCollection> configureServices)
```
Tratcher: `Func<IServiceCollection, IServiceProvider>` isn't as useful here because this one isn't providing any default services.

khellang: Might be best to just kill this overload and remove the dependency altogether, then...

src/Microsoft.AspNet.Owin/OwinExtensions.cs
```diff
@@ -67,21 +67,26 @@ public static IApplicationBuilder UseOwin(this IApplicationBuilder builder, Acti
...
         {
+            return app.UseBuilder(serviceProvider: null);
+        }
+
+        public static IApplicationBuilder UseBuilder(this AddMiddleware app, IServiceProvider serviceProvider)
+        {
             // Adapt WebSockets by default.
             app(OwinWebSocketAcceptAdapter.AdaptWebSockets);
-            var builder = new ApplicationBuilder(serviceProvider: null);
+            var builder = new ApplicationBuilder(serviceProvider: serviceProvider ?? null);
```
davidfowl: just pass the serviceProvider. `serviceProvider ?? null` does nothing

davidfowl:
Squash and :shipit:

test/Microsoft.AspNet.Owin.Tests/OwinExtensionTests.cs
```diff
@@ -0,0 +1,80 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Builder;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+
```
pranavkm: Extra newline.

test/Microsoft.AspNet.Owin.Tests/OwinExtensionTests.cs
```diff
@@ -0,0 +1,80 @@
...
+          Func<IDictionary<string, object>, Task>
+        >;
+    using AddMiddleware = Action<Func<
+          Func<IDictionary<string, object>, Task>,
+          Func<IDictionary<string, object>, Task>
+        >>;
+
+    public class OwinExtensionTests
+    {
+        static AppFunc notFound = async env => env["owin.ResponseStatusCode"] = 404;
```
pranavkm: Think this would complain if you have warning as errors on. Could you have it return a `Task` instead?

test/Microsoft.AspNet.Owin.Tests/OwinExtensionTests.cs
```diff
@@ -0,0 +1,80 @@
...
+                applicationBuilder.Run(async context =>
+                {
+                    fakeService = context.ApplicationServices.GetService<FakeService>();
+                });
+            }, new ServiceCollection().AddSingleton(new FakeService()).BuildServiceProvider());
+
+            list.Reverse().Aggregate(notFound, (next, middleware) => middleware(next)).Invoke(new Dictionary<string, object>());
+
+            Assert.NotNull(fakeService);
+            Assert.NotNull(serviceProvider?.GetService<FakeService>());
```
pranavkm: Shouldn't need to do `?.` here - `serviceProvider` shouldn't be null.

test/Microsoft.AspNet.Owin.Tests/OwinExtensionTests.cs
```diff
@@ -0,0 +1,80 @@
...
+        >>;
+
+    public class OwinExtensionTests
+    {
+        static AppFunc notFound = async env => env["owin.ResponseStatusCode"] = 404;
+
+        [Fact]
+        public void OwinConfigureServiceProviderAddsServices()
+        {
+            IList<CreateMiddleware> list = new List<CreateMiddleware>();
```
pranavkm: `var` all day everyday.

test/Microsoft.AspNet.Owin.Tests/OwinExtensionTests.cs
```diff
@@ -0,0 +1,80 @@
...
+                });
+            });
+
+            list.Reverse().Aggregate(notFound, (next, middleware) => middleware(next)).Invoke(new Dictionary<string, object>());
+
+            Assert.Null(fakeService);
+            Assert.Null(serviceProvider);
+        }
+
+        public class FakeService
```
pranavkm: `private`

test/Microsoft.AspNet.Owin.Tests/OwinExtensionTests.cs
```diff
@@ -0,0 +1,80 @@
...
+                serviceProvider = applicationBuilder.ApplicationServices;
+                applicationBuilder.Run(async context =>
+                {
+                    fakeService = context.ApplicationServices.GetService<FakeService>();
+                });
+            });
+
+            list.Reverse().Aggregate(notFound, (next, middleware) => middleware(next)).Invoke(new Dictionary<string, object>());
+
+            Assert.Null(fakeService);
```
pranavkm: Could you additionally set a bool inside the builder func and verify if it's true here? These values are just as likely to be null if the builder func never gets executed so we'd validate it actually was executed.

pranavkm:
:shipit: once comments are addressed.

test/Microsoft.AspNet.Owin.Tests/OwinExtensionTests.cs
JunTaoLuo: I had a reason for this hmm...

JunTaoLuo: Ahh, I wanted to use the Linq's version of `Reverse` which allows me to do the whole thing in one line but I guess I'll separate it and use System.Collection.Generic's version instead.

PR #469, opened 2015-11-10
Changed string.Format to StringBuilder

khellang:
// @davidfowl @benaadams

### DIFF
--- a/src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
+++ b/src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
@@ -50,15 +50,22 @@ namespace Microsoft.AspNet.Http.Extensions
             QueryString query = new QueryString(),
             FragmentString fragment = new FragmentString())
         {
-            string combinePath = (pathBase.HasValue || path.HasValue) ? (pathBase + path).ToString() : "/";
+            var combinedPath = (pathBase.HasValue || path.HasValue) ? (pathBase + path).ToString() : "/";
 
-            return new StringBuilder()
+            var encodedHost = host.ToString();
+            var encodedQuery = query.ToString();
+            var encodedFragment = fragment.ToString();
+
+            var length = scheme.Length + SchemeDelimiter.Length + encodedHost.Length
+                + combinedPath.Length + encodedQuery.Length + encodedFragment.Length;
+
+            return new StringBuilder(length)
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
```diff
@@ -48,7 +51,15 @@ public static class UriHelper
             FragmentString fragment = new FragmentString())
         {
             string combinePath = (pathBase.HasValue || path.HasValue) ? (pathBase + path).ToString() : "/";
-            return $"{scheme}://{host.ToString()}{combinePath}{query.ToString()}{fragment.ToString()}";
+
+            return new StringBuilder()
```
benaadams: Add up lengths first and initialize with capacity?

khellang: Is it worth passing the length to these?

khellang: LOL. Just refreshed this and noticed your comment :stuck_out_tongue_closed_eyes:

benaadams: Yes otherwise is allocs at size 16; then then starts doing interesting things when it overflows

src/Microsoft.AspNet.Http.Extensions/UriHelper.cs
```diff
@@ -47,8 +50,23 @@ public static class UriHelper
...
         {
-            string combinePath = (pathBase.HasValue || path.HasValue) ? (pathBase + path).ToString() : "/";
-            return $"{scheme}://{host.ToString()}{combinePath}{query.ToString()}{fragment.ToString()}";
+            var combinedPath = (pathBase.HasValue || path.HasValue) ? (pathBase + path).ToString() : "/";
+
+            var encodedHost = host.ToString();
+            var encodedQuery = query.ToString();
+            var encodedFragment = fragment.ToString();
+
+            var length = scheme.Length + SchemeDelimiter.Length + encodedHost.Length
```
davidfowl: Add a comment saying this is for performance in the form of:

```
// PERF: Reason
```

khellang: Aight :+1:

Issue #473, opened 2015-11-10
Don't set IHttpContextAccessor by default


DamianEdwards:
Setting up `IHttpContextAccessor` on every request is expensive and thus should probably be opt-in. Frameworks that require it (like MVC) can add it to the container as needed. If it's in the container, then we should set it up per request, otherwise, don't.

rynowak:
We can drop our reliance on this and similar things in MVC. See https://github.com/aspnet/Mvc/issues/3523

PR #475, opened 2015-11-11
Set IHttpContextAccessor only if DI provides it

pakrym:
https://github.com/aspnet/HttpAbstractions/issues/473

/cc @DamianEdwards @davidfowl  @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Http/HttpContextFactory.cs
+++ b/src/Microsoft.AspNet.Http/HttpContextFactory.cs
@@ -9,6 +9,10 @@ namespace Microsoft.AspNet.Http.Internal
     {
         private IHttpContextAccessor _httpContextAccessor;
 
+        public HttpContextFactory() : this(httpContextAccessor: null)
+        {
+        }
+
         public HttpContextFactory(IHttpContextAccessor httpContextAccessor)
         {
             _httpContextAccessor = httpContextAccessor;
@@ -17,13 +21,19 @@ namespace Microsoft.AspNet.Http.Internal
         public HttpContext Create(IFeatureCollection featureCollection)
         {
             var httpContext = new DefaultHttpContext(featureCollection);
-            _httpContextAccessor.HttpContext = httpContext;
+            if (_httpContextAccessor != null)
+            {
+                _httpContextAccessor.HttpContext = httpContext;
+            }
             return httpContext;
         }
 
         public void Dispose(HttpContext httpContext)
         {
-            _httpContextAccessor.HttpContext = null;
+            if (_httpContextAccessor != null)
+            {
+                _httpContextAccessor.HttpContext = null;
+            }
         }
     }
 }
--- a/test/Microsoft.AspNet.Http.Tests/HttpContextFactoryTests.cs
+++ b/test/Microsoft.AspNet.Http.Tests/HttpContextFactoryTests.cs
@@ -21,5 +21,16 @@ namespace Microsoft.AspNet.Http.Internal
             // Assert
             Assert.True(ReferenceEquals(context, accessor.HttpContext));
         }
+
+        [Fact]
+        public void AllowsCreatingContextWithoutSettingAccessor()
+        {
+            // Arrange
+            var contextFactory = new HttpContextFactory();
+
+            // Act & Assert
+            var context = contextFactory.Create(new FeatureCollection());
+            contextFactory.Dispose(context);
+        }
     }
 }


test/Microsoft.AspNet.Http.Tests/HttpContextFactoryTests.cs
```diff
@@ -21,5 +21,15 @@ public void CreateHttpContextSetsHttpContextAccessor()
             // Assert
             Assert.True(ReferenceEquals(context, accessor.HttpContext));
         }
+
+        [Fact]
+        public void AllowsCreatingContextWithoutSettingAccessor()
```
pakrym: This test is strange

davidfowl:
Downs this even work?

pakrym:
@davidfowl our DI supports this scenario, i checked

avanderhoorn:
Thanks for the heads up!

src/Microsoft.AspNet.Http/HttpContextFactory.cs
```diff
@@ -9,6 +9,10 @@ public class HttpContextFactory : IHttpContextFactory
     {
         private IHttpContextAccessor _httpContextAccessor;
 
+        public HttpContextFactory() : this(null)
```
Tratcher: @halter73 Wasn't the default DI container really bad at choosing between multiple constructors?

pranavkm: This was addressed as part of https://github.com/aspnet/DependencyInjection/issues/69

src/Microsoft.AspNet.Http/HttpContextFactory.cs
```diff
@@ -17,13 +21,19 @@ public HttpContextFactory(IHttpContextAccessor httpContextAccessor)
         public HttpContext Create(IFeatureCollection featureCollection)
         {
             var httpContext = new DefaultHttpContext(featureCollection);
-            _httpContextAccessor.HttpContext = httpContext;
+            if (_httpContextAccessor != null)
+            {
+                _httpContextAccessor.HttpContext = httpContext;
```
khellang: Does the null-conditional operator work on property setters? I.e. don't set the property if the instance is null. If not, this seems like a good example for a suggestion.

khellang:
Even though *your* DI supports it, doesn't make it clear to other containers that they have to conform to this convention (pick the most greedy ctor you can satisfy), unless the test suite reflects it. Are there tests for this in the test package/project? Just making sure so people get what they expect :smile:

src/Microsoft.AspNet.Http/HttpContextFactory.cs
davidfowl: I think so:

```C#
_httpContextAccessor?.HttpContext  = httpContext;
```

davidfowl:
@khellang good point. @pakrym and @pranavkm Can you guys write some tests in the DI spec to see what the existing behavior is?

pranavkm:
@davidfowl we have tests for it - however it's not part of the DI specification tests. https://github.com/aspnet/DependencyInjection/issues/328 tracks adding additional tests in the test specification.

pakrym:
:up: :date:

Tratcher:
:shipit:

src/Microsoft.AspNet.Http/HttpContextFactory.cs
khellang: Looks like there's a proposal (added to the C# 7 and VB 15 milestone) for this: https://github.com/dotnet/roslyn/issues/1276

davidfowl: Damn

davidfowl:
:shipit:

avanderhoorn:
What milestone is this?

Tratcher:
RC2

avanderhoorn:
:+1:

dneimke:
I've been bitten by this today.  We are using Autofac and had been injecting IHttpContextAccessor into some of our services.  

As of build rc2-16557 I now get an error stating that the IHttpContextAccessor cannot be resolved by the container.

I tried manually registering it with the following code but still not resolving:

```
builder.RegisterType<HttpContextAccessor>().As<IHttpContextAccessor>().SingleInstance();
```

davidfowl:
@dneimke Can you share a sample

dneimke:
It's a pretty big codebase I can try to create a repro but it will take time.  

But essentially, we are resolving injected interfaces via an Autofac Container.  This all ran fine until rc2-16557 (e.g. it ran fine last night).

The exception that Autofac is throwing includes the following:

```
Cannot resolve parameter 'Microsoft.AspNet.Http.IHttpContextAccessor httpContextAccessor' of constructor 'Void .ctor
````

**Comment by davidfowl** [comment]
@dneimke A minimal repro would be great

**Comment by dneimke** [comment]
OK, give me 30 mins.  

Do you have some test projects around this so I can see how you register it?

**Comment by rynowak** [comment]
@dneimke https://github.com/aspnet/Mvc/blob/dd774366f6fb18af9382ca84c276e6eb73c66b3a/test/WebSites/BasicWebSite/Startup.cs

**Comment by davidfowl** [comment]
What you're doing looks fine. I just want to see the end to end sample. Empty project with this single dependency registration and an attempt to resolve it.

**Comment by dneimke** [comment]
@davidfowl I'm on it!  :-)

**Comment by dneimke** [comment]
My tail is between my legs :-)  I have a minimal repro working fine using both OOTB DI and Autofac latest bits.

Thx for the prompt reply.

**Comment by dneimke** [comment]
I'll track it down and let you know if it is anything else significant that impacted on the framework side of the fence.   It's an odd one because nothing much changed in our code base and, as I said, it ran through the build just fine last night.

**Comment by davidfowl** [comment]
I'm sure it's something but narrowing it down helps! This was one of the big changes we made for performance reasons.

**Comment by dneimke** [comment]
I managed to fix it this morning by changing the registration from:

```
builder.RegisterInstance(Configuration).As<IConfiguration>().SingleInstance();    
```

to

```
builder.RegisterInstance(Configuration).As<IConfiguration>();    
```

I'll see if I can provide a canonical repro later today

Issue #437, opened 2015-10-13
ResponseCookies.Delete does not use the same default path as ResponseCookies.Append


tuespetre:
`ResponseCookies.Append` and the `CookieOptions` class [both use a default path of `"/"`](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http/ResponseCookies.cs#L44) but [the single-parameter version of `ResponseCookies.Delete` does not.](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http/ResponseCookies.cs#L81)

This can be a confusing road bump when you have set a cookie with a simple call like `Response.Cookies.Append("EditSuccess", string.Empty)` and then later try to delete it with `Response.Cookies.Delete("EditSuccess")`, because some browsers (namely Chrome) will not remove the cookie since it does not have a matching path specified.

PR #476, opened 2015-11-13
Set default path=/ when removing cookie

pakrym:
https://github.com/aspnet/HttpAbstractions/issues/437

/cc @Tratcher @JunTaoLuo @halter73


halter73:
:shipit:

davidfowl:
Weird we had that much logic that could be readily deleted :smile: /cc @Tratcher

pakrym:
:up: :date:

JunTaoLuo:
:shipit:

davidfowl:
:shipit:

PR #478, opened 2015-11-14
Add HtmlContentBuilder and HtmlTextWriter - Remove BufferedHtmlContent

rynowak:


src/Microsoft.AspNet.Html.Abstractions/HtmlContentBuilder.cs
```diff
@@ -1,62 +1,97 @@
...
-namespace Microsoft.Extensions.Internal
+namespace Microsoft.AspNet.Html.Abstractions
 {
     /// <summary>
-    /// Enumerable object collection which knows how to write itself.
+    /// An <see cref="IHtmlContentBuilder"/> implementation using an in memory list.
     /// </summary>
-    [DebuggerDisplay("{DebuggerToString()}")]
-    internal class BufferedHtmlContent : IHtmlContentBuilder
+    public class HtmlContentBuilder : IHtmlContentBuilder
```
rynowak: This is mostly copy-paste. Also some general improvements here because this is a public API now.

src/Microsoft.AspNet.Html.Abstractions/HtmlEncodedString.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.IO;
+using System.Text.Encodings.Web;
+
+namespace Microsoft.AspNet.Html.Abstractions
+{
+    /// <summary>
+    /// An <see cref="IHtmlContent"/> impelementation that wraps an HTML encoded <see cref="string"/>.
+    /// </summary>
+    [DebuggerDisplay("{DebuggerToString()}")]
+    public class HtmlEncodedString : IHtmlContent
```
rynowak: This will de-dupe some code in MVC, and was already duped here.

src/Microsoft.AspNet.Html.Abstractions/HtmlTextWriter.cs
```diff
@@ -0,0 +1,49 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+
+namespace Microsoft.AspNet.Html.Abstractions
+{
+    /// <summary>
+    /// A <see cref="TextWriter"/> which supports special processing of <see cref="IHtmlContent"/>.
+    /// </summary>
+    public abstract class HtmlTextWriter : TextWriter
```
rynowak: This will help resolve some layering issues in ViewFeatures/Razor

src/Microsoft.AspNet.Html.Abstractions/HtmlContentBuilder.cs
```diff
@@ -1,62 +1,97 @@
...
 
-        /// <summary>
-        /// Appends the <see cref="string"/> to the collection.
-        /// </summary>
-        /// <param name="unencoded">The <c>string</c> to be appended.</param>
-        /// <returns>A reference to this instance after the Append operation has completed.</returns>
+        /// <inheritdoc />
         public IHtmlContentBuilder Append(string unencoded)
         {
+            if (unencoded == null)
```
pranavkm: Maybe invert this:

```
if (unencoded != null)
{
   Entries.Add(unencoded);
}

return this;
```

No preference either way

src/Microsoft.AspNet.Html.Abstractions/HtmlEncodedString.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        /// <summary>
+        /// An <see cref="IHtmlContent"/> instance for <see cref="Environment.NewLine"/>.
+        /// </summary>
+        public static readonly IHtmlContent NewLine = new HtmlEncodedString(Environment.NewLine);
```
pranavkm: Should this be a property?

pranavkm:
:shipit:

src/Microsoft.AspNet.Html.Abstractions/HtmlContentBuilder.cs
```diff
@@ -1,62 +1,97 @@
...
 
-        /// <summary>
-        /// Appends the HTML encoded <see cref="string"/> to the collection.
-        /// </summary>
-        /// <param name="encoded">The HTML encoded <c>string</c> to be appended.</param>
-        /// <returns>A reference to this instance after the Append operation has completed.</returns>
+        /// <inheritdoc />
         public IHtmlContentBuilder AppendHtml(string encoded)
         {
+            if (encoded == null)
```
dougbu: `string.IsNullOrEmpty(encoded)`?

src/Microsoft.AspNet.Html.Abstractions/HtmlEncodedString.cs
davidfowl: Why?

src/Microsoft.AspNet.Html.Abstractions/HtmlContentBuilder.cs
rynowak: Sure.

src/Microsoft.AspNet.Html.Abstractions/HtmlContentBuilder.cs
rynowak: Sure

src/Microsoft.AspNet.Html.Abstractions/HtmlEncodedString.cs
rynowak: I don't see any reason to change this.

src/Microsoft.AspNet.Html.Abstractions/HtmlContentBuilder.cs
```diff
@@ -1,62 +1,97 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="HtmlContentBuilder"/> with the given list of entries.
+        /// </summary>
+        /// <param name="entries">
+        /// The list of entries. The <see cref="HtmlContentBuilder"/> will use this list without making a copy.
+        /// </param>
+        public HtmlContentBuilder(List<object> entries)
```
Eilon: Taking in `List<T>` is odd - is it basically for perf to avoid the copy?

rynowak: The idea here is that if you want to use this as a wrapper around pre-allocated concept. It could just as easily be `IList<T>` if that makes you happier.

Eilon: It would :smile:

PR #482, opened 2015-11-16
Change List<T> to IList<T>

rynowak:


rynowak:
:shipit: in person from @Eilon

Issue #496, opened 2015-12-02
SendFile extension method should fallback to StreamCopy if ISendFileFeature isn't available


davidfowl:

muratg:
Isn't the fallback mechanism in StaticFiles good enough?

davidfowl:
@muratg I'm moving that into here. That requires a static file dependency

muratg:
Sounds good :+1:

PR #497, opened 2015-12-03
Change SendFileAsync to use a fallback implementation instead of throwing

davidfowl:
- If the user wants to use the SendFile API directly then they can access the feature
explicitly.

#496


src/Microsoft.AspNet.Http.Extensions/SendFileResponseExtensions.cs
```diff
@@ -74,10 +74,56 @@ public static Task SendFileAsync(this HttpResponse response, string fileName, lo
...
+        {
+            cancel.ThrowIfCancellationRequested();
+
+            if (string.IsNullOrWhiteSpace(fileName))
+            {
+                throw new ArgumentNullException(nameof(fileName));
+            }
+            if (!File.Exists(fileName))
+            {
+                throw new FileNotFoundException(string.Empty, fileName);
```
muratg: Why not just put `fileName` instead?

src/Microsoft.AspNet.Http.Extensions/SendFileResponseExtensions.cs
```diff
@@ -33,7 +33,7 @@ public static bool SupportsSendFile(this HttpResponse response)
         /// Sends the given file using the SendFile extension.
         /// </summary>
         /// <param name="response"></param>
-        /// <param name="fileName"></param>
+        /// <param name="fileName">The full or relative path to the file.</param>
```
Tratcher: Callers should use the full path. Relative paths are unpredictable.

src/Microsoft.AspNet.Http.Extensions/SendFileResponseExtensions.cs
```diff
@@ -74,10 +74,56 @@ public static Task SendFileAsync(this HttpResponse response, string fileName, lo
...
+                FileShare.ReadWrite,
+                bufferSize: 1024 * 64,
+                options: FileOptions.Asynchronous | FileOptions.SequentialScan);
+
+            try
+            {
+                fileStream.Seek(offset, SeekOrigin.Begin);
+
+                await StreamCopyOperation.CopyToAsync(fileStream, outputStream, length, cancel);
+            }
```
muratg: Should you catch instead and log?

davidfowl: Log with what?

src/Microsoft.AspNet.Http.Extensions/SendFileResponseExtensions.cs
```diff
@@ -74,10 +74,56 @@ public static Task SendFileAsync(this HttpResponse response, string fileName, lo
...
         }
+
+        // Not safe for overlapped writes.
+        private static async Task SendFileAsync(Stream outputStream, string fileName, long offset, long? length, CancellationToken cancel)
+        {
+            cancel.ThrowIfCancellationRequested();
+
+            if (string.IsNullOrWhiteSpace(fileName))
+            {
+                throw new ArgumentNullException(nameof(fileName));
```
Eilon: Wrong exception type. Can this throw earlier, where the parameter is provided?

src/Microsoft.AspNet.Http.Extensions/StreamCopyOperation.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    // FYI: In most cases the source will be a FileStream and the destination will be to the network.
+    internal static class StreamCopyOperation
+    {
+        private const int DefaultBufferSize = 1024 * 16;
+
+        internal static async Task CopyToAsync(Stream source, Stream destination, long? length, CancellationToken cancel)
+        {
+            long? bytesRemaining = length;
+            byte[] buffer = new byte[DefaultBufferSize];
```
muratg: nit: var

src/Microsoft.AspNet.Http.Extensions/StreamCopyOperation.cs
```diff
@@ -0,0 +1,61 @@
...
+            {
+                // The natural end of the range.
+                if (bytesRemaining.HasValue && bytesRemaining.Value <= 0)
+                {
+                    return;
+                }
+
+                cancel.ThrowIfCancellationRequested();
+
+                int readLength = buffer.Length;
```
muratg: nit: var

src/Microsoft.AspNet.Http.Extensions/SendFileResponseExtensions.cs
```diff
@@ -74,10 +74,56 @@ public static Task SendFileAsync(this HttpResponse response, string fileName, lo
             var sendFile = response.HttpContext.Features.Get<IHttpSendFileFeature>();
             if (sendFile == null)
             {
-                throw new NotSupportedException(Resources.Exception_SendFileNotSupported);
+                return SendFileAsync(response.Body, fileName, offset, count, cancellationToken);
```
Tratcher: This fallback makes SupportsSendFile unnecessary, remove it.

davidfowl: It's not unnecessary. It's just convenient. But sure.

src/Microsoft.AspNet.Http.Extensions/SendFileResponseExtensions.cs
```diff
@@ -74,10 +74,56 @@ public static Task SendFileAsync(this HttpResponse response, string fileName, lo
...
+            {
+                throw new ArgumentOutOfRangeException(nameof(length), length, string.Empty);
+            }
+
+            var fileStream = new FileStream(
+                fileName,
+                FileMode.Open,
+                FileAccess.Read,
+                FileShare.ReadWrite,
+                bufferSize: 1024 * 64,
```
Tratcher: This is a 64kb buffer, but StreamCopyOperation uses a 16kb buffer. Would it help if they were aligned? Do we know if this is a managed or native buffer?

davidfowl: Copied your codez bro :smile: . This should totally use the memory pool.

src/Microsoft.AspNet.Http.Extensions/StreamCopyOperation.cs
Tratcher: https://github.com/aspnet/StaticFiles/issues/91 needs to move to this repo as well.

src/Microsoft.AspNet.Http.Extensions/SendFileResponseExtensions.cs
muratg: Ugh just thought there would be a logger here. Apparently not.

src/Microsoft.AspNet.Http.Extensions/SendFileResponseExtensions.cs
davidfowl: Copy pasta

src/Microsoft.AspNet.Http.Extensions/SendFileResponseExtensions.cs
davidfowl: @Tratcher Maybe they should be the same.

src/Microsoft.AspNet.Http.Extensions/StreamCopyOperation.cs
davidfowl: Use the fancy issue mover

Issue #489, opened 2015-11-27
HeaderDictionary can't add existing key


tenor:
The header dictionary won't allow a header that already exists in the dictionary to be added.

```csharp
            Response.Headers.Add("Hello", new string[] { "One" });
            Response.Headers.Add("Hello", new string[] { "Two" });
```

The code above will throw an exception on the second addition.
This was permissible in Web API 2.

```csharp
        //Web API 2 Example
        // GET api/values
        public IHttpActionResult Get()
        {
            var response = Request.CreateResponse(HttpStatusCode.OK);
            response.Headers.Add("Hello", new string[] { "One" });
            response.Headers.Add("Hello", new string[] { "Two", "Three" });
            return ResponseMessage(response);
        }
```

In this scenario, Web API 2 adds the header values to the underlying values collection. I think this is the right behavior because that was the developer's intention even though they broke the contract.

davidfowl:
Why not just call 

```C#
Response.Headers.Append("Hello", "One");
Response.Headers.Append("Hello", "Two");
```

Tratcher:
FYI: That's the required contract for IDictionary. WebApi didn't implement IDictionary.

Yes, you should use the Append extension for this behavior.

tenor:
I don't feel strongly either way about this. `Append` is an extension that is not visible unless I'm already aware of it.
I'm just pointing out that one way breaks a working pattern in older apps, and the other way properly adheres to the IDictionary contract.

davidfowl:
This is one of those things that I think needs to change in application code. I think Append could be closer to the actual type since it's a fundamental operation (or is it?).

davidfowl:
I'm moving those extension methods into Http.Abstractions.

PR #498, opened 2015-12-03
Move the header extensions to Http.Abstractions

davidfowl:
#489

/cc @Tratcher @muratg


khellang:
No csproj changes! Oh, how I :heart: this :grin:

davidfowl:
:golf: :clap:

Tratcher:
Why these and not any of the other extensions? They aren't abstractions, they're just convenience methods.

Tratcher:
If you think these are core functionality then put them directly on the interface.

davidfowl:
Nah. The interface should have whatever is minimally required to implement it. These are extensions that work on top of *any*  implementation.

The bug is specifically about header extensions, also these don't bring in any other dependencies so easy to move.

Issue #352, opened 2015-07-19
Possible FormFileCollection optimizations


davidfowl:
- Avoid closure by looping over files instead of using Find and FindAll.
- Cache parsing of ContentDisposition header

khellang:
> - Cache parsing of ContentDisposition header

Should be fixed by #500

PR #500, opened 2015-12-04
Add Name and FileName to IFormFile

### LINKED ISSUE #499: No easy way to get file names in  IFormFileCollection / IFormFile
When using `IFormCollection`, it is easy to get the form keys.
But when working with the `IFormFileCollection` interface, there is not way to get all form names and file names (somehing like `IFormCollection.Keys` / `IFormCollection.FileNames`) or at least two properties `IFormFile.FileName` and `IFormFile.Name`.
I have found a workaround (using `ContentDispositionHeaderValue` like in the `FormFileCollection` class, but it is not very fun to use.

khellang:
Fixes #499. These names should probably be cached somewhere. Anyone opposed to exposing a `Name` property on `IFormFile` backed by a `Lazy<string>`?

### DIFF
--- a/src/Microsoft.AspNet.Http/Features/FormFeature.cs
+++ b/src/Microsoft.AspNet.Http/Features/FormFeature.cs
@@ -149,7 +149,10 @@ namespace Microsoft.AspNet.Http.Features.Internal
                             // Find the end
                             await section.Body.DrainAsync(cancellationToken);
 
-                            var file = new FormFile(_request.Body, section.BaseStreamOffset.Value, section.Body.Length)
+                            var name = HeaderUtilities.RemoveQuotes(contentDisposition.Name) ?? string.Empty;
+                            var fileName = HeaderUtilities.RemoveQuotes(contentDisposition.FileName) ?? string.Empty;
+
+                            var file = new FormFile(_request.Body, section.BaseStreamOffset.Value, section.Body.Length, name, fileName)
                             {
                                 Headers = new HeaderDictionary(section.Headers),
                             };
--- a/src/Microsoft.AspNet.Http/Features/FormFile.cs
+++ b/src/Microsoft.AspNet.Http/Features/FormFile.cs
@@ -8,15 +8,16 @@ namespace Microsoft.AspNet.Http.Features.Internal
 {
     public class FormFile : IFormFile
     {
-        private Stream _baseStream;
-        private long _baseStreamOffset;
-        private long _length;
+        private readonly Stream _baseStream;
+        private readonly long _baseStreamOffset;
 
-        public FormFile(Stream baseStream, long baseStreamOffset, long length)
+        public FormFile(Stream baseStream, long baseStreamOffset, long length, string name, string fileName)
         {
             _baseStream = baseStream;
             _baseStreamOffset = baseStreamOffset;
-            _length = length;
+            Length = length;
+            Name = name;
+            FileName = fileName;
         }
 
         public string ContentDisposition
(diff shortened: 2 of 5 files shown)


davidfowl:
This isn't how I think it should be done. We should update FormFile to have a name property on it. Make it lazy there.

khellang:
@davidfowl Hence my "These names should probably be cached somewhere. Anyone opposed to exposing a `Name` property on `IFormFile` backed by a `Lazy<string>`?" :wink:

davidfowl:
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http/Features/FormFeature.cs#L150-L160

davidfowl:
@khellang Lazy<string> is overkill, it doesn't need to be thread safe. Also add a test for this once you're done.

khellang:
> Lazy is overkill, it doesn't need to be thread safe.

So are you saying parse the name eagerly? Or just `_name ?? (_name = GetName())`?

davidfowl:
Lazy is fine but using `Lazy<T>` is just overhead since we don't care about thread safety.

khellang:
@davidfowl How does it look now?

src/Microsoft.AspNet.Http/Features/FormFeature.cs
```diff
@@ -149,7 +149,9 @@ public Task<IFormCollection> ReadFormAsync(CancellationToken cancellationToken)
                             // Find the end
                             await section.Body.DrainAsync(cancellationToken);
 
-                            var file = new FormFile(_request.Body, section.BaseStreamOffset.Value, section.Body.Length)
+                            var name = HeaderUtilities.RemoveQuotes(contentDisposition.Name);
```
davidfowl: Can name ever be null here?

davidfowl: Seems not

davidfowl:
Much better!

khellang:
What about `FileName` as well?

davidfowl:
Go for it

davidfowl:
I think I change my mind on the lazy. Not worth it. Parse it out and just assign eagerly.

khellang:
Have you looked at the getters for Name and FileName in CDHV? Granted, it's not often you send files, and not very many of them, so it probably won't matter much.

khellang:
Added 5d1cd10 to close #352 as well.

src/Microsoft.AspNet.Http/Features/FormFile.cs
```diff
@@ -3,20 +3,23 @@
...
     public class FormFile : IFormFile
     {
-        private Stream _baseStream;
-        private long _baseStreamOffset;
-        private long _length;
+        private readonly Stream _baseStream;
+        private readonly long _baseStreamOffset;
 
-        public FormFile(Stream baseStream, long baseStreamOffset, long length)
+        public FormFile(Stream baseStream, long baseStreamOffset, long length, ContentDispositionHeaderValue contentDisposition)
```
davidfowl: Can we move the quote removal on to the outside and just pass the name and file name. It's weird to have the header dependency in here (always was)

khellang: Done!

src/Microsoft.AspNet.Http/FormFileCollection.cs
```diff
@@ -2,33 +2,39 @@
...
 
-        private static string GetName(string contentDisposition)
-        {
-            // Content-Disposition: form-data; name="myfile1"; filename="Misc 002.jpg"
-            ContentDispositionHeaderValue cd;
-            ContentDispositionHeaderValue.TryParse(contentDisposition, out cd);
-            return HeaderUtilities.RemoveQuotes(cd?.Name);
+            foreach (var file in this)
+            {
+                if (string.Equals(name, file.Name))
```
Tratcher: Ordinal or ignore case?

khellang: Good question. I just kept what was already there. What would you expect?

davidfowl: OrdinalIgnoreCase

khellang: Done!

davidfowl:
Ok this PR is good to go once the CI passes. Then we'll have to see what repos break. If I had to guess. MVC would be broken

PR #501, opened 2015-12-06
Allow feature cache to be updated

### LINKED ISSUE #402: [Perf] High Allocator DefaultHttpContext

### LINKED ISSUE #502: Allow feature cache to be updated+invalidated
No point in pooling if you can't actually update stuff

See: https://github.com/aspnet/benchmarks/pull/34/files#diff-a48f9d66e0bc9b9c9ec15dae230d8e50R167

in aspnet/benchmarks#34 for specific usage

benaadams:
No point in pooling if you can't actually update stuff; allows for a saving of 959KB per 4k requests

See: https://github.com/aspnet/benchmarks/pull/34/files#diff-a48f9d66e0bc9b9c9ec15dae230d8e50R167
 in https://github.com/aspnet/benchmarks/pull/34 for specific usage

Resolves #502
Resolves #402

Allocation details when pooling (in benchmarks repo), usual stuff, 4k requests, 16 pipelined, using FF

Pre-allocations for DefaultContext 1.12MB

before-allocs

Post-allocations for DefaultContext 160KB
 
after-allocs

Almost all of which is logging scope

before-allocs


benaadams:
Restore on is having issues

src/Microsoft.AspNet.Http/Features/IFeatureCache.cs
```diff
@@ -3,8 +3,9 @@
 
 namespace Microsoft.AspNet.Http.Features
 {
-    internal interface IFeatureCache
+    public interface IFeatureCache
     {
         void CheckFeaturesRevision();
+        void ReplaceFeatures(IFeatureCollection features);
```
davidfowl: UpdateFeatures

benaadams: Done

src/Microsoft.AspNet.Http/DefaultHttpContext.cs
```diff
@@ -49,15 +50,33 @@ void IFeatureCache.CheckFeaturesRevision()
...
+        {
+            _features = features;
+            ResetFeatures();
+
+            _request.UpdateFeatures(features);
+            _response.UpdateFeatures(features);
+
+            _authenticationManager?.UpdateFeatures(features);
+            _connection?.UpdateFeatures(features);
+            _websockets?.UpdateFeatures(features);
```
Tratcher: Should these lazy sub-objects be cleared out per request? Only some requests will need them. If they don't get cleared then eventually every pooled HttpContext will have all of them.

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
```diff
@@ -29,12 +29,23 @@ void IFeatureCache.CheckFeaturesRevision()
...
+        public void UpdateFeatures(IFeatureCollection features)
+        {
+            _features = features;
+            ResetFeatures();
+        }
+
+        private void ResetFeatures()
+        {
+            _authentication = null;
+            _response = null;
```
Tratcher: It looks like _response is no longer used and can be removed.

src/Microsoft.AspNet.Http/Features/QueryFeature.cs
```diff
@@ -41,11 +41,23 @@ void IFeatureCache.CheckFeaturesRevision()
         {
             if (_cachedFeaturesRevision != _features.Revision)
             {
-                _request = null;
-                _cachedFeaturesRevision = _features.Revision;
+                ResetFeatures();
             }
         }
 
+        public void UpdateFeatures(IFeatureCollection features)
```
Tratcher: This doesn't make sense, UpdateFeatures is never called on any of these Feature objects, they just get nulled out.

benaadams: Added it to interface so they need to be implemented; could remove it from interface and just have them as methods? Though exposing the ability to invalidate the cache seems good; why you'd want to do it for these cases - I don't know...

src/Microsoft.AspNet.Http/DefaultHttpContext.cs
benaadams: Paid the price; fairly small, are a defined size (rather than an unknown user code type) - might as well? (Happy either way)

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
benaadams: Removed

src/Microsoft.AspNet.Http/Features/QueryFeature.cs
Tratcher: I think you're conflating two types of objects, Managers and Features. The Managers have UpdateFeatures, where the Features : IFeatureCache with CheckFeaturesRevision. CheckFeaturesRevision never appears to be called on the Manager objects.

benaadams: Gets called by the `FeatureHelpers.GetOrCreateAndCache` methods

Tratcher: Ah, gotcha. You still have two classes of objects though, IFeatureCache { CheckFeaturesRevision } and IUpdatableFeatureCache : IFeatureCache { UpdateFeatures }. Only the managers will be updated.

benaadams: K, will split the types.

benaadams: Added IFeatureManager and removed the method from the features

src/Microsoft.AspNet.Http/Features/IFeatureManager.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Http.Features
+{
+    public interface IFeatureManager
```
davidfowl: Why do we need this new public interface?

davidfowl:
@benaadams We need some tests for this to make sure all of our features properly update and nothing is cached (unexpectedly).

src/Microsoft.AspNet.Http/Features/IFeatureManager.cs
benaadams: No, but would be implementing without contract; which is probably fine :)
Removed.

benaadams:
> ... and nothing is cached (unexpectedly).

Thinking future proofing... reflection to discover interfaces and make sure they are null?

src/Microsoft.AspNet.Http/DefaultHttpContext.cs
Tratcher: *shrug* It's a probability question. We can figure it out based on profiles later.

src/Microsoft.AspNet.Http/Features/QueryFeature.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.AspNet.Http.Features.Internal
 {
     public class QueryFeature : IQueryFeature, IFeatureCache
     {
-        private readonly IFeatureCollection _features;
+        private IFeatureCollection _features;
```
Tratcher: The changes to these Feature classes can be reverted, especially _features. ResetFeatures doesn't matter much one way or the other.

benaadams: Done

Tratcher:
Looking good, but yes it still needs some tests.

benaadams:
Added test; with it, found some areas where it wasn't actually caching and some where it was over clearing, so fixed those too.

test/Microsoft.AspNet.Http.Tests/DefaultHttpContextTests.cs
```diff
@@ -147,6 +150,140 @@ public void SetItems_NewCollectionUsed()
             Assert.Same(item, context.Items["foo"]);
         }
 
+#if DNX451
```
davidfowl: Why is this ifdefed?

benaadams: coreclr doesn't have `IsInterface` in reflection

benaadams: Ahh, are some extension methods on `.GetTypeInfo()`

benaadams: Enabled for all types now

src/Microsoft.AspNet.Http/Authentication/DefaultAuthenticationManager.cs
```diff
@@ -29,12 +28,28 @@ void IFeatureCache.CheckFeaturesRevision()
...
             if (_cachedFeaturesRevision != _features.Revision)
             {
-                _authentication = null;
-                _response = null;
-                _cachedFeaturesRevision = _features.Revision;
+                ResetFeatures();
             }
         }
 
+        void IFeatureCache.SetFeaturesRevision()
```
davidfowl: Why the explicit implementation?

benaadams: Don't want someone to call it; as it says "mark cache as up to date; but don't check it"

src/Microsoft.AspNet.Http/Features/FeatureHelpers.cs
```diff
@@ -20,6 +20,7 @@ internal static class FeatureHelpers
             {
                 obj = features.Get<T>();
                 cachedObject = obj;
+                cache.SetFeaturesRevision();
```
Tratcher: I don't think you need SetFeaturesRevision, it looks like it's always called after CheckFeatureRevision which already updates _cachedFeaturesRevision.

src/Microsoft.AspNet.Http/DefaultWebSocketManager.cs
```diff
@@ -27,12 +27,29 @@ void IFeatureCache.CheckFeaturesRevision()
...
-                _request = null;
-                _webSockets = null;
-                _cachedFeaturesRevision = _features.Revision;
+                ResetFeatures();
             }
         }
 
+        void IFeatureCache.SetFeaturesRevision()
+        {
+            _cachedFeaturesRevision = _features.Revision;
```
Tratcher: Would it make sense to also set _cachedFeaturesRevision in the constructor? It would save you work on your first call to CheckFeaturesRevision.

src/Microsoft.AspNet.Http/Features/FeatureHelpers.cs
benaadams: Yeah, this call doesn't alter the features collection.

src/Microsoft.AspNet.Http/DefaultWebSocketManager.cs
benaadams: Yep; done both, just running tests locally.

benaadams:
Changes made

Tratcher:
Rebased and merged.

benaadams:
w00p! :+1:

PR #503, opened 2015-12-08
* Removing unused dependencies from code

pranavkm:
* Cleaning up usage of CodeAnalysis.


src/Microsoft.AspNet.Http/ParsingHelpers.cs
```diff
@@ -73,7 +73,6 @@ public override int GetHashCode()
         #endregion
     }
 
-    [System.CodeDom.Compiler.GeneratedCode("App_Packages", "")]
```
pranavkm: This was an odd one. Was this attribute intentionally placed here?

davidfowl:
:shipit:

Issue #386, opened 2015-09-04
Use DateTimeOffset for CookieOptions.Expires


PinpointTownes:
https://github.com/aspnet/Security/pull/432#discussion_r38775905

BrennanConroy:
59b32934e322660983322ceccf37dfc48da56a7f

PinpointTownes:
:clap: :clap:

PR #509, opened 2015-12-14
Change Expires to DateTimeOffset

BrennanConroy:
#386


Tratcher:
:shipit: Then fix Security and whoever else breaks.

PR #516, opened 2015-12-17
Streamlining feature cache and object re-initialiation code paths

lodejard:
Shouldn't increase object size, but does improve maintainability


lodejard:
/cc @DamianEdwards also

Outside of pooling this also has the benefit of making DefaultHttpContext more useful as a base class. By overriding the Initialize methods an app author can subclass the request/response/etc companion objects, otherwise you have no control over those associated objects.

samples/SampleApp/PooledHttpContext.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        DefaultHttpRequest _pooledHttpRequest;
+        DefaultHttpResponse _pooledHttpResponse;
+
+        public PooledHttpContext(IFeatureCollection featureCollection) : 
+            base(featureCollection)
+        {
+        }
+
+        protected override HttpRequest InitializeHttpRequest()
```
davidfowl: You need to override *all* of methods in order to properly pool right?

src/Microsoft.AspNet.Http.Features/FeatureReferences.cs
```diff
@@ -0,0 +1,57 @@
...
+    public struct FeatureReferences<TCache>
+    {
+        public FeatureReferences(IFeatureCollection collection)
+        {
+            Collection = collection;
+            Cache = default(TCache);
+            Revision = collection.Revision;
+        }
+
+        public readonly IFeatureCollection Collection;
```
davidfowl: If these are fields instead of properties for a specific reason you should put that in a comment.

samples/SampleApp/PooledHttpContext.cs
lodejard: That's a policy choice, really - the ConnectionInfo/WebSocket objects might be used infrequently enough it could be could be more efficient to let them be garbage collected. But yes, you would override all child object management methods to take full control

src/Microsoft.AspNet.Http/DefaultHttpRequest.cs
```diff
@@ -11,96 +11,41 @@
...
-            {
-                return FeatureHelpers.GetOrCreateAndCache(
-                    this, 
-                    _features, 
-                    (f) => new RequestCookiesFeature(f), 
-                    ref _cookies);
-            }
-        }
+        private IFormFeature FormFeature =>
+            _features.Fetch(ref _features.Cache.Form, this, self => new FormFeature(self));
```
davidfowl: `self => new FormFeature(self)`

`f` => new RequestCookiesFeature(f)`

Use f in both places?

test/Microsoft.AspNet.Http.Tests/DefaultHttpContextTests.cs
```diff
@@ -274,6 +271,7 @@ private static void TestFeatureProperties(object value, IFeatureCollection featu
                 {
                     if (property.Name.Contains("Feature"))
                     {
+                        Console.WriteLine($"{value.GetType().Name} {property.Name}");
```
davidfowl: Remove this code.

lodejard: :bangbang:

davidfowl:
Looks good to me :shipit: . @Tratcher might want to comment

test/Microsoft.AspNet.Http.Tests/DefaultHttpContextTests.cs
```diff
@@ -154,68 +154,63 @@ public void SetItems_NewCollectionUsed()
...
-            Assert.NotEqual(0, newFeatures.Count());
+            Assert.NotEqual(3, newFeatures.Count());
         }
 
-        void TestCachedFeaturesAreNull(Type type, object value, IFeatureCollection features)
+        void TestCachedFeaturesAreNull(object value, IFeatureCollection features)
         {
+            var type = value.GetType();
 
             var fields = type
```
Tratcher: This doesn't verify anything anymore because the object layout has changed. The only field it will find is `FeatureReferences<FeatureInterfaces> _features;`.

Tratcher: Unless it manages to unroll the nested structs?

Tratcher:
*shrug*

lodejard:
What's the emoji for *shrug*? :neutral_face:? :no_mouth:? :see_no_evil:?

:)

test/Microsoft.AspNet.Http.Tests/DefaultHttpContextTests.cs
lodejard: Yeah, a unit test that reflects over the private implementation details seemed a little :suspect: to me in the first place, but comparing the nested struct to default(T) could certainly be done...

pranavkm:
¯\_(ツ)_/¯

test/Microsoft.AspNet.Http.Tests/DefaultHttpContextTests.cs
lodejard: Test updated to assert _features field state is reset to default(T). Though I'm still highly suspicious of the practice.

test/Microsoft.AspNet.Http.Tests/DefaultHttpContextTests.cs
```diff
@@ -149,93 +149,88 @@ public void SetItems_NewCollectionUsed()
...
         {
+            TestCachedFeaturesAreSet(context, features);
+            TestCachedFeaturesAreSet(context.Request, features);
+            TestCachedFeaturesAreSet(context.Response, features);
+            TestCachedFeaturesAreSet(context.Authentication, features);
+            TestCachedFeaturesAreSet(context.Connection, features);
+            TestCachedFeaturesAreSet(context.WebSockets, features);
+        }
+
+        void TestCachedFeaturesAreSet(object value, IFeatureCollection features)
```
Tratcher: TestCachedFeaturesAreSet needs to be updated as well.

lodejard: That one is already getting the property, and getting the interface type, and asserting they are the same. Isn't using reflection on fields to verify a private has been changed highly unusual? I can add unit test over FeatureReferences is that verify it's calling through only once, but doing the testing of that class indirectly seems wrong.

benaadams: I was trying to test if the cached (private var) was correct for the current and any future additions so act as a regression safety test; even found a few instances where it wasn't actually being saved to the private var with the test.

benaadams:
LGTM, and does look a lot better

davidfowl:
:shipit:

benaadams:
Hmm, related  https://github.com/aspnet/HttpAbstractions/issues/522 though not a blocker; thought I could work round this issue for Frame but the call stack runs through here.

Current flow
`Middleware` -> `httpContext.Request.Path` -> `DefaultHttpRequest.get_Path` -> `GetAndCache<T>` -> `FeatureCollectionExtenstions.Get<T>` -> `(TFeature) features[typeof (TFeature)]` -> `Frame.FastFeaturesGet(Type)`

src/Microsoft.AspNet.Http.Features/FeatureReferences.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                feature = Collection.Get<TFeature>();
+                if (feature == null)
+                {
+                    feature = factory(state);
+
+                    Collection.Set(feature);
+                    if (!cleared)
+                    {
+                        Cache = default(TCache);
```
benaadams: Bit late; but not sure about this if statement? What's it doing?

benaadams: Should it be https://github.com/aspnet/HttpAbstractions/pull/524 ?

Issue #393, opened 2015-09-08
Dispose the FileBufferingReadStream created by ReadFormAsync


Tratcher:
BufferingHelper.EnableRewind creates a temp file to store the request buffer in if it gets too large. Currently this file is only deleted when closed by GC, where it could be deleted at the end of the given request. Register for Response.RegisterForDispose when the file is created.

https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http/BufferingHelper.cs#L47-L49

BrennanConroy:
fixed via #520

PR #520, opened 2015-12-23
Register FileBufferingReadStream for disposal

BrennanConroy:
#393


src/Microsoft.AspNet.Http/BufferingHelper.cs
```diff
@@ -50,7 +50,9 @@ public static HttpRequest EnableRewind(this HttpRequest request, int bufferThres
             {
                 // TODO: Register this buffer for disposal at the end of the request to ensure the temp file is deleted.
```
Tratcher: Remove the TODO

Tratcher:
:shipit:

PR #521, opened 2015-12-25
Faster StartsWithSegments Ordinal check

benaadams:
Framework's call to native comes with a cost

Before
string-startswith-before

After
string-startswith-after


benaadams:
As an aside; the sets and whatnot have high costs as they are also doing dictionary lookups; typecasts and initalizing the feature cache

halter73:
:shipit:

Do you have any thoughts @Tratcher?

davidfowl:
Not a fan of this. Looks pretty unmaintainable. Is it really worth it? Can we make a change like this in corefx instead?

benaadams:
Would be change to coreclr https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/String.cs#L2562

Will propose.

halter73:
@davidfowl No less maintainable than `GetAsciiString` :stuck_out_tongue:

I see your point though. This is something probably done once per request at most, and I doubt most apps are going to specify `StringComparison.Ordinal`.

benaadams:
@halter73 The crazy is currently localized to one project :)

benaadams:
Raised PR in coreclr https://github.com/dotnet/coreclr/pull/2667

PR #525, opened 2015-12-31
[Proposal] Generic accessor into feature collection

### LINKED ISSUE #522: Make IFeatureCollection have generic<T> accessor
Rather than [have it as an Extension](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http.Features/FeatureCollectionExtensions.cs#L14); so the [type of checks](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Generated.cs#L64) in `Frame.FastFeatureGet(Type key)` becomes `typeof` checks of `T` in `Frame.FastFeatureGet<T>()` 

At that point they also become compile time constants meaning 90% of the time the entire method will be Jitted away - as per the `typeof` checks in [System.Numerics.Vectors](https://github.com/dotnet/corefx/blob/master/src/System.Numerics.Vectors/src/System/Numerics/Vector.cs#L18)

benaadams:
Also making the FeatureInterfaces structs public and checking for those means the `IGetConcreteStructWrapped` fast path can be done in `Frame.FastFeatureGet<T>` (half speed of having no branches)

Resolves #522


src/Microsoft.AspNet.Http.Features/FeatureCollection.cs
```diff
@@ -93,6 +94,21 @@ IEnumerator IEnumerable.GetEnumerator()
             }
         }
 
+        public TFeature Get<TFeature>()
+        {
+            var type = typeof(TFeature);
+            if (type.GetTypeInfo().IsValueType)
```
davidfowl: Why this check?

benaadams: `(TFeature)this[type]` blows up if its a struct and not in the collection with a NRE

PR #526, opened 2015-12-31
Move generic Get/Set into Interface

### LINKED ISSUE #522: Make IFeatureCollection have generic<T> accessor

benaadams:
Resolves #522

Paired Kestrel change https://github.com/aspnet/KestrelHttpServer/pull/549


pranavkm:
@benaadams could you send a follow PR to fix this - https://github.com/aspnet/Hosting/blob/dev/test/Microsoft.AspNet.Hosting.Tests/WebApplicationTests.cs#L539?

Issue #530, opened 2016-01-07
Bad performance when invoking using UseMiddleware with ctor args


davidfowl:
https://github.com/aspnet/HttpAbstractions/blob/05c476d1b5e818a25d62e3b14c13f75278ed8d05/src/Microsoft.AspNet.Http.Abstractions/Extensions/UseMiddlewareExtensions.cs#L83-L94

We need to do a similar trick to the DI system (Expression tree compilation to speed this up)

davidfowl:
https://github.com/aspnet/HttpAbstractions/commit/ca8136b73c1d2c840a56708616a9b377d35e6580

PR #531, opened 2016-01-07
Initial change to compile middleware invoke method when extra args 

davidfowl:
#530 

We could maybe move some of this compilation logic to Common (not sure if we do it in more than one place).

/cc @DamianEdwards @pranavkm

### DIFF
--- a/src/Microsoft.AspNet.Http.Abstractions/Extensions/UseMiddlewareExtensions.cs
+++ b/src/Microsoft.AspNet.Http.Abstractions/Extensions/UseMiddlewareExtensions.cs
@@ -95,7 +95,6 @@ namespace Microsoft.AspNet.Builder
 
         private static Func<T, HttpContext, IServiceProvider, Task> Compile<T>(MethodInfo methodinfo, ParameterInfo[] parameters)
         {
-
             // If we call something like
             // 
             // public class Middleware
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNet.Http.Abstractions.Tests/UseMiddlewareTest.cs
```diff
@@ -6,6 +6,7 @@
 using Microsoft.AspNet.Builder;
```
davidfowl: @mikaelm12 didn't format this file properly :disappointed:

JamesNK:
Do you need to worry about methods that could have ref or out parameters? You should probably either add support for them or throw an exception with an error message when building the expression tree.

davidfowl:
@JamesNK seems unrelated to this change in particular. What happens when you do methodInfo.Invoke and you have ref and out params? Does it just work?

JamesNK:
Yeah it does so this would be a regression.

This is a method I wrote to handle calling pretty much any method info via a compiled expression: https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Utilities/ExpressionReflectionDelegateFactory.cs#L86-L190

Looking at what I wrote, you might be ok with ref parameters because you don't need to capture the return value. You probably shouldn't try to resolve out params from DI. Either leave them alone or tell the user it is an error to have out parameter here.

src/Microsoft.AspNet.Http.Abstractions/Extensions/UseMiddlewareExtensions.cs
```diff
@@ -80,20 +88,97 @@ public static IApplicationBuilder UseMiddleware(this IApplicationBuilder app, Ty
...
+            methodArguments[0] = httpContextArg;
+            for (int i = 1; i < parameters.Length; i++)
+            {
+                var parameterType = parameters[i].ParameterType;
+                if (parameterType.IsByRef)
+                {
+                    throw new NotSupportedException(Resources.FormatException_InvokeDoesNotSupportRefOrOutParams(InvokeMethodName));
+                }
+
+                var parameterTypeExpression = new Expression[] {
```
pranavkm: `c#
new Expression[]
{

};
```
(brace on separate line)

pranavkm:
:shipit:

Issue #446, opened 2015-10-20
ContentDispositionHeaderValue doesnt parse FileName and Name correctly


Tratcher:
_From @Gillardo on October 14, 2015 9:17_

In order to upload a file, i have seen the example code here
https://github.com/aspnet/Mvc/commit/437eb93bdec0d9238d672711ebd7bd3097b6537d

To get the name of the file that is uploaded, i am using the ContentDispositionHeaderValue class, using the TryParse method, as used in the example, which is like so

```
    public async Task<IActionResult> PostAsync(IFormFile file)
    {
        ContentDispositionHeaderValue parsedContentDisposition;
        ContentDispositionHeaderValue.TryParse(file.ContentDisposition, out parsedContentDisposition);
        ....
    }
```

But it seems to include \" around the FileName and Name properties.

My Content-Disposition is the same as it was in MVC5, like so
    Content-Disposition: form-data; name="file"; filename="Key.png"

Since these properties are strings anyway, it doesnt need to wrap the values in quotes.

Fix is very simple

```
parsedContentDisposition.FileName = parsedContentDisposition.FileName.TrimStart('\"').TrimEnd('\"');
parsedContentDisposition.Name = parsedContentDisposition.Name.TrimStart('\"').TrimEnd('\"');
```

_Copied from original issue: aspnet/Home#987_

Tratcher:
_From @Eilon on October 15, 2015 4:55_

I believe the value is supposed to be a quoted string, per the RFCs. See Section 4.1 of RFC 6266: http://tools.ietf.org/html/rfc6266#section-4.1

And `quoted-string` is defined here: http://tools.ietf.org/html/rfc2616#section-2.2

>        quoted-string  = ( <"> *(qdtext | quoted-pair ) <"> )
>        qdtext         = <any TEXT except <">>

So, while it is perhaps allowed to omit the quotes, it is certainly permitted.

The examples section in RFC 6266 makes this clear: http://tools.ietf.org/html/rfc6266#section-5

> 5.  Examples
> 
> 
>    Direct the UA to show "save as" dialog, with a filename of
>    "example.html":
> 
>      Content-Disposition: Attachment; filename=example.html
> 
>    Direct the UA to behave as if the Content-Disposition header field
>    wasn't present, but to remember the filename "an example.html" for a
>    subsequent save operation:
> 
>      Content-Disposition: INLINE; FILENAME= "an example.html"
> 
>    Note: This uses the quoted-string form so that the space character
[...]

PR #533, opened 2016-01-13
Feature/#446 content disposition

EvAlex:
This resolves issue #446. 

It turned out that `IFormFile` now has `FileName` property, so there's no need to use `ContentDispositionHeaderValue.Parse()` to parse `IFormFile.ContentDisposition` in MVC controller action.

But nevertheless `ContentDispositionHeaderValue.Parse()` is still publicly available and can be used to parse `IFormFile.ContentDisposition`. That method returns an instance of `ContentDispositionHeaderValue`, that has `FileName` property.

Here's an example of `IFormFile.ContentDisposition` value: 
```
form-data; name="avatar"; filename="my-face.png"
``` 
This is the case when user posted a file named "my-face.png" from html page with form with `enctype="multipart/form-data"` and input with `name="avatar"`.

So when we parse this ContentDisposition with `ContentDispositionHeaderValue.Parse()`, we get an instance of `ContentDispositionHeaderValue` with FileName containing **quoted** "my-face.png" and Name containing **quoted** "avatar". This is not convenient, so this MR unquotes those values.

### DIFF
--- a/src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
+++ b/src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
@@ -64,13 +64,13 @@ namespace Microsoft.Net.Http.Headers
 
         public string Name
         {
-            get { return GetName(NameString); }
+            get { return Unquote(GetName(NameString)); }
             set { SetName(NameString, value); }
         }
 
         public string FileName
         {
-            get { return GetName(FileNameString); }
+            get { return Unquote(GetName(FileNameString)); }
             set { SetName(FileNameString, value); }
         }
 
@@ -460,6 +460,16 @@ namespace Microsoft.Net.Http.Headers
 
             return value.Length > 1 && value.StartsWith("\"", StringComparison.Ordinal)
                 && value.EndsWith("\"", StringComparison.Ordinal);
+        }
+
+        /// <summary>
+        /// Unquotes string if it is not null. Does nothing if the string is null.
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        private string Unquote(string value)
+        {
+            return value == null ? value : value.Trim('"');
         }
 
         // tspecials are required to be in a quoted string.  Only non-ascii needs to be encoded.
--- a/test/Microsoft.Net.Http.Headers.Tests/ContentDispositionHeaderValueTest.cs
+++ b/test/Microsoft.Net.Http.Headers.Tests/ContentDispositionHeaderValueTest.cs
@@ -152,7 +152,7 @@ namespace Microsoft.Net.Http.Headers
             Assert.Equal(1, contentDisposition.Parameters.Count);
             Assert.Equal("FILENAME", contentDisposition.Parameters.First().Name);
             Assert.Equal("\"=?utf-99?Q?R=mlsZcODTmFtZS5iYXQ=?=\"", contentDisposition.Parameters.First().Value);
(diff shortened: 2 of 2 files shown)


src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
```diff
@@ -462,6 +462,16 @@ private bool IsQuoted(string value)
                 && value.EndsWith("\"", StringComparison.Ordinal);
         }
 
+        /// <summary>
+        /// Unquotes string if it is not null. Does nothing if the string is null.
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        private string Unquote(string value)
```
khellang: Why isn't [`HeaderUtilities.RemoveQuotes`](https://github.com/aspnet/HttpAbstractions/blob/fa8c5cbb3a8eca81223c34eb90fa3a81107b093e/src/Microsoft.Net.Http.Headers/HeaderUtilities.cs#L226-L233) used here?

EvAlex: You are right. Should I create new MR to fix it?

khellang: Just push a new commit to the same branch and it'll be included here :smile:

khellang: Or maybe even just make the changed, and commit usint `git commit --amend` to amend the last commit. Then you can force push to this PR.

test/Microsoft.Net.Http.Headers.Tests/ContentDispositionHeaderValueTest.cs
```diff
@@ -418,6 +418,34 @@ public void Parse_SetOfValidValueStrings_ParsedCorrectly()
         }
 
         [Fact]
+        public void Parse_FileNameWithoutQuotes()
```
Tratcher: These test names are misleading. It sounds like it's describing the input, and you're not actually testing the behavior of parse here, just FileName. Try FileName_RemovesQuotes.

Tratcher:
@EvAlex Can you please sign the CLA and address the given feedback?

src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
```diff
@@ -64,13 +64,13 @@ public IList<NameValueHeaderValue> Parameters
 
         public string Name
         {
-            get { return GetName(NameString); }
+            get { return Unquote(GetName(NameString)); }
```
davidfowl: Why unquote every time?

Tratcher: The final representation is supposed to be quoted. Removing them is just for convenience. It could be cached, but I don't think that's an immediate concern.

davidfowl: What final representation? Just assign it unquoted in the ctor....

EvAlex:
Ok, I've signed the CLA. I'ff fix the issues with the code in a while.

muratg:
@EvAlex Any updates?

Tratcher:
Closing as stale. Please re-submit when get a chance to update it.

jkotalik:
Going to fix issues here and resend this pull request.

Issue #528, opened 2016-01-05
Share StreamCopyOperation implementations


Tratcher:
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http.Extensions/StreamCopyOperation.cs
https://github.com/aspnet/StaticFiles/blob/dev/src/Microsoft.AspNet.StaticFiles/StreamCopyOperation.cs

glennc:
If StaticFiles already uses Http.Extensions then we should just use that location, otherwise find somewhere else for it to live.

davidfowl:
Internal helper in common?

BrennanConroy:
We can make it public in Http.Extensions since StaticFiles already depends on that and then we'll copy Ben's changes to the StaticFiles one over to Http.Extensions

BrennanConroy:
Fixed via 3f84e992f48f52d86404732915ebe86bfdd2a610

PR #537, opened 2016-01-14
Make StreamCopyOperation public and update it to same as StaticFiles

BrennanConroy:
Part 1 of https://github.com/aspnet/HttpAbstractions/issues/528

@Tratcher @davidfowl

### DIFF
--- a/src/Microsoft.AspNet.Http.Extensions/StreamCopyOperation.cs
+++ b/src/Microsoft.AspNet.Http.Extensions/StreamCopyOperation.cs
@@ -2,55 +2,67 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Http.Extensions/StreamCopyOperation.cs
```diff
@@ -10,47 +11,58 @@
 namespace Microsoft.AspNet.Http
```
Tratcher: Change the namspace to Microsoft.AspNet.Extensions. This will not be a commonly used type.

Tratcher: Woops, Microsoft.AspNet.Http.Extensions.

Tratcher:
:shipit:

Issue #536, opened 2016-01-14
Remove IsLocal


Tratcher:
We keep getting questions like this: https://github.com/aspnet/Hosting/issues/570#issuecomment-171571555

Environments are the correct way to handle these scenarios most of the time.

We should remove IsLocal, and possibly replace it with an extension method that just returns RemoteIP.Equals(LocalIP).

@davidfowl @muratg @blowdart

blowdart:
Yea, kill it IMO

davidfowl:
> We should remove IsLocal, and possibly replace it with an extension method that just returns RemoteIP.Equals(LocalIP).

If we add it back as an extension method won't be in the same place? It seems like an OK thing to remove from `ConnectionInfo` though.

Tratcher:
Current recommendation is to remove it without adding an extension.

muratg:
Yup, no extension method. This was redundant anyway.

BrennanConroy:
Fixed via 1f21540fd53a839dcfe7004788abe1e9a1d820a4

dazinator:
I found this useful for authorising a user based on if it is a Local request or not. I am aware that this information (ip address) could be spoofed, but I am happy to accept that risk given that this functionality is in place only when the web application is first installed, and it treats the local user as a local admin whilst they set up the site, and blocks non local requests.

What's the replacement?

Tratcher:
IsLocal is more dangerous than it once was because of the prevalence of local reverse proxies (e.g. Nginx, HttpPlatformHandler, AspNetCoreModule, etc.), most requests now appear to be local.

How would your setup process work for a site deployed to a hoster like Azure Web Sites where there is no local access?

Consider using a Staging Environment mode or real credentials based authentication.

dazinator:
@Tratcher understood. If there is a chance of IsLocal no longer being accurate then I don't wish to use it. I'll have to re-think my setup process. :(

blowdart:
So what I've seen working for some folks is

1. Check for the presence of a marker file or database entry to see if already configured. If present goto 6.
2. Check for the presence of a first run password in a settings file, if not present throw error with instructions on how to set it. Never set a default one in any installer.
3. Login page which asks for the setup password
4. Run setup
5. Drop the "I've been configured" marker
6. Profit.

dazinator:
@blowdart haha thanks - I like that suggestion, Team City uses that concept too, it generates an system admin login token which it writes to a local log file. Cheers

PR #538, opened 2016-01-15
Remove IsLocal

BrennanConroy:
#536
@Tratcher 

Will need to update a couple repos when this goes in

### DIFF
--- a/src/Microsoft.AspNet.Http.Abstractions/ConnectionInfo.cs
+++ b/src/Microsoft.AspNet.Http.Abstractions/ConnectionInfo.cs
@@ -18,8 +18,6 @@ namespace Microsoft.AspNet.Http
 
         public abstract int LocalPort { get; set; }
 
-        public abstract bool IsLocal { get; set; }
-
         public abstract X509Certificate2 ClientCertificate { get; set; }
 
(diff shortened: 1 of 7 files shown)


Tratcher:
:shipit: Check Hosting, WebListener, Kestrel, and BasicMiddleware.

davidfowl:
:shipit:

davidfowl:
Related to this idea, should we move IsHttps to an extension method?

Tratcher:
@davidfowl *shrug*. Last chance if you're going to, but it's not a priority.

davidfowl:
The weird thing about it not being a computed property is that the actual scheme and IsHttps can be out of sync. I'm not aware of that being an intentional design choice.

Tratcher:
Well, it is actually computed: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http/DefaultHttpRequest.cs#L98-L102, but I guess that's not a guaranteed implementation detail.

brockallen:
I'd very much like the `IsHttps` property to be kept. My $0.02.

davidfowl:
Might make sense to make these things all computed and not settable.

BrennanConroy:
:up: :date: Removed ```IsLocal``` from `IHttpConnectionFeature`

Tratcher:
:shipit:

Issue #527, opened 2016-01-01
Fix untestable extension methods


davidfowl:
- https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Extensions/SendFileResponseExtensions.cs
- https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Extensions/FormFileExtensions.cs

These methods demand that files be on disk. What makes it worse is that these are extension methods so they cannot be stubbed.

Maybe we should make these virtual on `HttpResponse` and `IFormFile` (interface method).

/cc @Eilon @Tratcher @lodejard

glennc:
If we are going to do this we should do it now, moving methods around after RC2 should be avoided.

davidfowl:
another alternative might be to add an overload that takes `IFileInfo`

davidfowl:
Ughm `IFileInfo` is read only...

khellang:
> another alternative might be to add an overload that takes IFileInfo

Add an overload where? For SendFile?

What about the `FormFileExtensions`? Promote them to `IFormFile`?

davidfowl:
@khellang Those are the 2 options. Either promote to interface or use some overload that takes a writable file abstraction.

davidfowl:
For SendFile, we should allow passing `IFileInfo`

muratg:
Moving this to Backlog as we will be in RC2 ask mode very soon.  If you feel strongly about this issue, please ping me.

lodejard:
On the Form file extensions, why not `Task SaveAsync(Stream target)`?

The `IFileInfo` for SendFile is good though - because if it's has a non-null-or-empty PhysicalFilePath some servers can accelerate that.

khellang:
I think `Task SaveAsync(Stream target)` makes sense, then the `Task SaveAsync(string path)` overloads could be extension methods again :stuck_out_tongue:

davidfowl:
Only problem with extension methods is that it reintroduces the original problem if you ever use them... Which is what the original problem and why this bug was filed :smile:

khellang:
> Only problem with extension methods is that it reintroduces the original problem if you ever use them

Yes and no. You can still mock the `SaveAs`method on the interface, which is an improvement.

khellang:
Also, if it's changed to a `Stream`, does the `SaveAs` name make sense?

davidfowl:
[...]

PR #543, opened 2016-01-22
Changed SaveAs[Async](string) to CopyTo[Async](Stream)

khellang:
#527


src/Microsoft.AspNet.Http/Features/FormFile.cs
```diff
@@ -74,28 +74,20 @@ public Stream OpenReadStream()
...
-        public void SaveAs(string path)
+        /// <param name="target">The stream to save the file to.</param>
+        public void Save(Stream target)
         {
-            using (var fileStream = File.Create(path, DefaultBufferSize))
-            {
-                var inputStream = OpenReadStream();
-                inputStream.CopyTo(fileStream, DefaultBufferSize);
-            }
+            OpenReadStream().CopyTo(target, DefaultBufferSize);
```
davidfowl: Shouldn't we be disposing our stream? If we don't who does?

khellang: Yeah, I've seen it being disposed other places...

muratg:
@Tratcher Could you have a look?

Tratcher:
@lodejard @davidfowl What's the point of having IFormFile.CopyTo(Stream) when you already have IFormFile.OpenReadStream().CopyTo(Stream)?

Should we just remove SaveAs(string)?

khellang:
> What's the point of having IFormFile.CopyTo(Stream) when you already have IFormFile.OpenReadStream().CopyTo(Stream)?

Good point :stuck_out_tongue_closed_eyes::+1:

Tratcher:
Hmm I guess it would actually be something like:
```
IFormFile.CopyTo(Stream output)
{
 using (var readStream = OpenReadStream())
 {
  reasStream.CopyTo(output);
 }
}
```
This might be worth it if it was actually important to dispose the readStream, but so far it isn't.

Tratcher:
Rebased and merged.

PR #544, opened 2016-01-22
Promote SendFile extension methods to HttpResponse

khellang:
#527 

I think we can remove some code duplication in after this as well:
https://github.com/aspnet/StaticFiles/blob/dev/src/Microsoft.AspNetCore.StaticFiles/StaticFileContext.cs#L379-L397
https://github.com/aspnet/StaticFiles/blob/dev/src/Microsoft.AspNetCore.StaticFiles/StaticFileContext.cs#L331-L346


src/Microsoft.AspNet.Http.Abstractions/HttpResponse.cs
```diff
@@ -103,5 +106,89 @@ public abstract class HttpResponse
...
+            if (sendFile == null)
+            {
+                return SendFileAsync(Body, file.PhysicalPath, offset, count, cancellationToken);
+            }
+
+            return sendFile.SendFileAsync(file.PhysicalPath, offset, count, cancellationToken);
+        }
+
+        // Not safe for overlapped writes.
+        private static async Task SendFileAsync(Stream outputStream, string fileName, long offset, long? length, CancellationToken cancellationToken)
```
khellang: Do we still want this fallback?

davidfowl:
I'm not sure this should be on HttpResponse but I'll let others chime in

src/Microsoft.AspNetCore.Http.Abstractions/project.json
```diff
@@ -11,11 +11,13 @@
   },
   "dependencies": {
     "Microsoft.AspNetCore.Http.Features": "1.0.0-*",
+    "Microsoft.Extensions.FileProviders.Abstractions": "1.0.0-*",
```
Tratcher: HttpAbstractions picking up a dependency on FileProviders is pretty strange.

src/Microsoft.AspNetCore.Http.Abstractions/HttpResponse.cs
```diff
@@ -103,5 +106,95 @@ public abstract class HttpResponse
...
         /// <param name="permanent"><c>True</c> if the redirect is permanent (301), otherwise <c>false</c> (302).</param>
         public abstract void Redirect(string location, bool permanent);
+
+        /// <summary>
+        /// Sends the given file using the SendFile extension.
+        /// </summary>
+        /// <param name="file">The file to send.</param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public virtual Task SendFileAsync(IFileInfo file, CancellationToken cancellationToken = default(CancellationToken))
```
Tratcher: These should be abstract and the implementation should be in DefaultHttpResponse like most of the other APIs. (if these should even be here. we'll discuss that separately.)

Tratcher:
:-1: We don't even let [WriteAsync(string)](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/HttpResponseWritingExtensions.cs) live directly on HttpResponse. SendFileAsync is even less common.

The IFileInfo overloads are interesting, but they belong in Extensions, not Abstractions, especially since they come with an added FileProviders dependency.

src/Microsoft.AspNetCore.Http.Abstractions/project.json
khellang: It's not just FileProviders, it's FileProviders.Abstractions. In my head, it made sense that one abstraction package depends on another :smile:

davidfowl:
+1 I think extension method + IFileInfo is good enough

davidfowl:
closing this

Issue #515, opened 2015-12-17
Reevaluate behavior when cookie parsing fails


rynowak:
Related to issue #390 but this topic is more about what the behavior should be when you're in this state.

I ended up with request cookies like the following for http://localhost:5000/:
```
__RequestVerificationToken=xhhkbQ7Vw0bBpyKt924bA4mEwn-xqrEj8w_Icr5Z2LQnXtugI8qt_glxLRAP3KPQR8RGBUukjfQzPOFCO2sffThjcUmtdgSkzn2iZ4NoObo1;
rm3EK3lo-DQ=CfDJ8DZcqGjS2eNPqurEN8WnNPUoyWWl9b3SROeJgbx_QsBsWViYvx7lI8jIQf5nv9_3B8zVdvmQdTYuqACH5p9Lx07aH8sctN-kt6CJTA4pp39O2Yymvs7M2IBx8eeTqeEn5uXh9_yV66f6dqK4l78r7dQ; 
Sample=36; 
ipt={"v":{"L":3},"pt":{"d":3},"ct":{},"_t":44,"_v":"2"}; 
.AspNet.Session=f5b61640-9efc-a9f8-32c1-6291f903f795; 
BQFxCZqo5n4=CfDJ8GVk6iB7ymZKmY1csOqk7Cds8G0Hd8XFeWgB9u4RB0r73u2gJvYuQmjGY2vn8WtGvFFKwUOLS3LVlkIxH9YmvQBopm69anm0VuXYBePCzYMYRruxRQ9ma34AAZSI76o3eV6cV0xsU3MYytMgTkC_f1U; 
eZYsIZozZe4=CfDJ8GVk6iB7ymZKmY1csOqk7CeNhzXg8fanKIOAcxGcBWXzm5AYBDlOgfSNHG8dLtWG2bFucr_PNhmjyEybiZQgGrdL2CEmvrhxR4MIoh6FgXdcx7N6OHL0tgDn520ud3i-ODYlXGr3aYT-5Eg3M8ulwvc;
XSRF-TOKEN=CfDJ8GVk6iB7ymZKmY1csOqk7CfYDif53fpXVX_mfM1SVKdA4fDDMi3rx98lBVks-zLiGjMv3j6SgAEHMe4VwKqMw5WCdJObXVgF1kTLA62U8SQJ2eLK2cv4vhcmlixQN3dhtOt_yeiPYtAkipDlhPUWiAg
```

[...]

PR #551, opened 2016-02-03
Make forgiving vs strict cookie parsers.

Tratcher:
#515 @rynowak @lodejard @riegelj @muratg 
Provide strict and forgiving versions of the list parsers. The forgiving version attempts to read past and discard invalid values.

This is preliminary only, the other headers will also need to be updated if we like the design changes. Some of their tests are currently broken by the changes.

### DIFF
--- a/src/Microsoft.Net.Http.Headers/HttpHeaderParser.cs
+++ b/src/Microsoft.Net.Http.Headers/HttpHeaderParser.cs
@@ -39,13 +39,23 @@ namespace Microsoft.Net.Http.Headers
             T result;
             if (!TryParseValue(value, ref index, out result))
             {
-                throw new FormatException(string.Format(CultureInfo.InvariantCulture, "Invalid value '{0}'.",
-                    value?.Substring(index) ?? "<null>"));
+                throw new FormatException(string.Format(CultureInfo.InvariantCulture,
+                    "The header contains invalid values at index {0}: '{1}'", index, value ?? "<null>"));
             }
             return result;
         }
 
         public virtual bool TryParseValues(IList<string> values, out IList<T> parsedValues)
+        {
+            return TryParseValues(values, strict: false, parsedValues: out parsedValues);
+        }
+
+        public virtual bool TryParseStrictValues(IList<string> values, out IList<T> parsedValues)
+        {
+            return TryParseValues(values, strict: true, parsedValues: out parsedValues);
+        }
+
+        protected virtual bool TryParseValues(IList<string> values, bool strict, out IList<T> parsedValues)
         {
             Contract.Assert(_supportsMultipleValues);
             // If a parser returns an empty list, it means there was no value, but that's valid (e.g. "Accept: "). The caller
@@ -71,9 +81,14 @@ namespace Microsoft.Net.Http.Headers
                             results.Add(output);
                         }
                     }
-                    else
+                    else if (strict)
                     {
                         return false;
+                    }
+                    else
+                    {
+                        // Skip the invalid values and keep trying.
+                        index++;
                     }
                 }
             }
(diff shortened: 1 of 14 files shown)


Tratcher:
*ping*

src/Microsoft.Net.Http.Headers/HttpHeaderParser.cs
```diff
@@ -85,7 +100,17 @@ public virtual bool TryParseValues(IList<string> values, out IList<T> parsedValu
...
+        {
+            return ParseValues(values, strict: false);
+        }
+
+        public virtual IList<T> ParseStrictValues(IList<string> values)
+        {
+            return ParseValues(values, strict: true);
+        }
+
+        protected virtual IList<T> ParseValues(IList<string> values, bool strict)
```
moozzyk: do you expect this to be overwritten or is it just maybe for testing/mocking?

test/Microsoft.Net.Http.Headers.Tests/CookieHeaderValueTest.cs
```diff
@@ -110,8 +111,48 @@ public static TheoryData<string> InvalidCookieValues
...
+
+                var header2 = new CookieHeaderValue("name2", "value2");
+                var validString2 = "name2=value2";
+
+                var header3 = new CookieHeaderValue("name3", "value3");
+                var validString3 = "name3=value3";
+
+                var invalidString1 = "ipt={\"v\":{\"L\":3},\"pt\":{\"d\":3},\"ct\":{},\"_t\":44,\"_v\":\"2\"}";
+
+                dataset.Add(null, new[] { invalidString1 });
```
moozzyk: Can you add some cases where quotes were not closed?

moozzyk: and maybe an empty list?

test/Microsoft.Net.Http.Headers.Tests/CookieHeaderValueTest.cs
```diff
@@ -225,5 +266,41 @@ public void CookieHeaderValue_TryParseList_AcceptsValidValues(IList<CookieHeader
...
+            // ParseList aways returns a list, even if empty. TryParseList may return null (via out).
+            Assert.Equal(cookies ?? new List<CookieHeaderValue>(), results);
+        }
+
+        [Theory]
+        [MemberData(nameof(ListOfInvalidCookieHeaderDataSet))]
+        public void CookieHeaderValue_TryParseList_ExcludesInvalidValues(IList<CookieHeaderValue> cookies, string[] input)
+        {
+            IList<CookieHeaderValue> results;
+            bool result = CookieHeaderValue.TryParseList(input, out results);
```
moozzyk: nit var

test/Microsoft.Net.Http.Headers.Tests/CookieHeaderValueTest.cs
```diff
@@ -110,8 +111,48 @@ public static TheoryData<string> InvalidCookieValues
                 return dataset;
             }
         }
+        public static TheoryData<IList<CookieHeaderValue>, string[]> ListOfInvalidCookieHeaderDataSet
```
moozzyk: super nit: add a new line before the method

test/Microsoft.Net.Http.Headers.Tests/CookieHeaderValueTest.cs
Tratcher: TryParse returns null out values and Parse returns an empty list. See line 293

src/Microsoft.Net.Http.Headers/HttpHeaderParser.cs
```diff
@@ -71,10 +81,15 @@ public virtual bool TryParseValues(IList<string> values, out IList<T> parsedValu
...
                     }
-                    else
+                    else if (strict)
                     {
                         return false;
                     }
+                    else
+                    {
+                        // Skip the invalid values and keep trying.
+                        index++;
```
moozzyk: I guess `++` is fine because TryParse will skip to the separator?

Tratcher: TryParse will update index to the first invalid character, so we skip it and try again.

moozzyk:
LGTM

src/Microsoft.Net.Http.Headers/HttpHeaderParser.cs
Tratcher: In theory. This is the base parser and some of the implementations may need to customize it.

moozzyk: :+1:

Tratcher:
:up: :date: Cleaned up. Expanded to the other types with list parsers.

moozzyk:
LGTM

Issue #553, opened 2016-02-04
Allocations in FormReader


rynowak:
#### There's a lot of overhead right now in FormReader

Data is from 3000 requests to https://github.com/aspnet/Performance/tree/dev/testapp/BigModelBinding (x64). The client is doing a form post of about 100 form fields - we kinda consider this the upper bound for the amount of data (and shape of data) that you'd want to put through MVC model binding.

![image](https://cloud.githubusercontent.com/assets/1430011/12822124/596eacc2-cb1b-11e5-8448-167d49b4d552.png)

Based on this profile, my napkin math shows about 77mb of allocations from our underlying representation (`string` + `System.Collections.Generic.Dictionary<String, StringValues>` + etc) and about 122mb coming from various overheads.

I'm excluding from the 122mb stuff that's in theory covered by @benaadams current work (byte/char buffers, etc).

------------------

##### `Task<string>` - 47mb
![image](https://cloud.githubusercontent.com/assets/1430011/12822658/e7b6d598-cb1d-11e5-82d5-47885f6604e3.png)
Cost of being async

##### `Task<Nullable<KeyValuePair<string, string>>>` - 28mb
![image](https://cloud.githubusercontent.com/assets/1430011/12822696/1859fee6-cb1e-11e5-89b0-62d885221ebf.png)
Cost of being async

##### `Dictionary<string, List<String>>`, `List<string>`, `string[]` - 25mb 12mb 9mb
![image](https://cloud.githubusercontent.com/assets/1430011/12822782/80c70b86-cb1e-11e5-9c37-fefecead57de.png)
Scratch data used in `KeyValueAccumulator` to build the final `Dictionary<string, StringValues>`

Note that 5mb of `string[]` allocations are coming from a hashset in MVC, so I excluded it from this total

==================

##### Some thoughts

We should consider a design for `FormReader` where we pass the accumulator around or store values in fields/properties instead of returning them via `Task<T>`. Using `async` plus `Task<T>` at such a chatty level is what causes all of this overhead. Using `Task` on the other hand can avoid this issue. 

If we're in a state where we know the entire body is buffered in memory, we might want to just optimize that path to do a synchronous read. That's probably a simpler fix, and it will result in running much more efficient code without *any* async overhead.

[...]

PR #554, opened 2016-02-04
Lower alloc KeyValueAccumulator

benaadams:
Avoid allocating `List<string>` per value 

@rynowak did you mean something like this in #553 ?


rynowak:
Yeah, this is exactly suggesting.

One possible pivot here (and I don't think we need to do it without evidence that it's good) would be to use geometric growth on the backing `string[]` for `StringValues` and teach it to ignore null elements - but I'm not sure we need that.

src/Microsoft.AspNetCore.WebUtilities/KeyValueAccumulator.cs
```diff
@@ -9,24 +9,33 @@ namespace Microsoft.AspNetCore.WebUtilities
...
             {
-                values.Add(value);
+                if (values.Count <= 1)
+                {
+                    _accumulator[key] = new string[] { values[0], value };
+                }
+                else
+                {
+                    var currentValues = values.ToArray();
+                    Array.Resize(ref currentValues, values.Count + 1);
```
Tratcher: Yeah, this growth by one is what the List was intended to avoid.

benaadams: Will save allocs for values with counts  > 3 (or > 5 if you include lists internals); any idea on how often that occurs?
e.g. when I get into that situation I'll generally move to a json post as the array post is easier to reason about what's going on.

benaadams: There is the checkbox hack double pattern 
```html
<input name="foo" type="checkbox" value="true" />
<input name="foo" type="hidden" value="false" />
```
However this change should still have less allocations than `List` in this case

davidfowl: Yes this change isn't great. Easy to create a form that causes tons of resize calls.

benaadams: Betterz?

src/Microsoft.AspNetCore.WebUtilities/KeyValueAccumulator.cs
```diff
@@ -9,24 +9,55 @@ namespace Microsoft.AspNetCore.WebUtilities
...
+                }
+                else
+                {
+                    // Third value for this key
+                    // Add zero count entry and move to data to expanding list dictionary
+                    _accumulator[key] = default(StringValues);
+
+                    if (_expandingAccumulator == null)
+                    {
+                        _expandingAccumulator = new Dictionary<string, List<string>>();
```
benaadams: needs `StringComparer.OrdinalIgnoreCase`

benaadams:
Addressed multi-value growth issues.

Tratcher:
@rynowak for verification.

rynowak:
Sure bro

rynowak:
This is definitely having the intended effect.

**Before**
image

**After**
image

Saves 19mb in this case.

muratg:
@Tratcher Are we taking this?

rynowak:
I'm not @Tratcher - but this is a huge perf win for big and small inputs

davidfowl:
@Tratcher is this on your radar?

Tratcher:
Yeah, it's next on the list...

Tratcher:
Rebased, squashed, and merged.

PR #555, opened 2016-02-04
Add IFileInfo overloads for SendFileAsync.

Tratcher:
#527 @davidfowl @lodejard @khellang

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Extensions/SendFileResponseExtensions.cs
+++ b/src/Microsoft.AspNetCore.Http.Extensions/SendFileResponseExtensions.cs
@@ -7,6 +7,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.FileProviders;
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Http.Extensions/SendFileResponseExtensions.cs
```diff
@@ -19,9 +20,66 @@ public static class SendFileResponseExtensions
...
+            {
+                throw new ArgumentNullException(nameof(file));
+            }
+            CheckRange(offset, count, file.Length);
+
+            if (string.IsNullOrEmpty(file.PhysicalPath))
+            {
+                using (var fileContent = file.CreateReadStream())
+                {
+                    fileContent.Seek(offset, SeekOrigin.Begin);
```
davidfowl: Is the stream always seekable?

davidfowl:
Small comment around seek-ability of the stream. We should probably throw if we can't seek and the index isn't 0.

lodejard:
and don't seek if the offset is 0

Tratcher:
:up::date: Seek will throw if it's not seekable. No need to add our own exception unless it becomes a frequent point of confusion.

PR #550, opened 2016-02-03
Use System.Buffers

benaadams:
Resolves #545

Added System.Buffer

 - [X] OwinWebSocketAdapter.CloseAsync
 - [X] Microsoft.AspNetCore.WebUtilities.BufferedReadStream.ctor
 - [X] Microsoft.AspNetCore.WebUtilities.FormReader.ctor (variable initialized)
 - [X] Microsoft.AspNetCore.WebUtilities.HttpRequestStreamReader.ctor
 - [X] Microsoft.AspNetCore.WebUtilities.HttpResponseStreamWriter.ctor
 - [X] Microsoft.AspNetCore.WebUtilities.MultipartReaderStream.Read
 - [X] Microsoft.AspNetCore.WebUtilities.MultipartReaderStream.ReadAsync
 - [X] Microsoft.AspNetCore.WebUtilities.StreamHelperExtensions.DrainAsync
 - [X] Microsoft.AspNetCore.WebUtilities.WebEncoders.Base64UrlDecode
 - [X] Microsoft.AspNetCore.WebUtilities.WebEncoders.Base64UrlEncode
 - [X] Microsoft.Net.Http.Headers.ContentDispositionHeaderValue.TryDecode5987

Extra 

- [X] Microsoft.AspNetCore.WebUtilities.FileBufferingReadStream

Now also resolves #548 

...except when content is buffered to disk; which needs this issue resolving in corefx https://github.com/dotnet/corefx/issues/5598 
Due to `FileStream`'s internal buffer is not being exposed/settable.

PR #556, opened 2016-02-07
Use System.Buffers + ValueTask in FormReading

benaadams:
Built on #550
Addressing Task allocations in  #553 (last commit in this PR)

ValueTask reduces Task allocation by FormReader to 1%
Post for 30,000 requests - byte[] allocations are as follows 0.5MB on rent vs 440MB on MemoryStream originally

/cc @rynowak @davidfowl @Tratcher @stephentoub

### DIFF
--- a/src/Microsoft.AspNetCore.Owin/WebSockets/OwinWebSocketAdapter.cs
+++ b/src/Microsoft.AspNetCore.Owin/WebSockets/OwinWebSocketAdapter.cs
@@ -34,6 +34,7 @@ namespace Microsoft.AspNetCore.Owin
 
     public class OwinWebSocketAdapter : WebSocket
     {
+        private const int _rentedBufferSize = 1024;
         private IDictionary<string, object> _websocketContext;
         private WebSocketSendAsync _sendAsync;
         private WebSocketReceiveAsync _receiveAsync;
@@ -127,7 +128,7 @@ namespace Microsoft.AspNetCore.Owin
                 await CloseOutputAsync(closeStatus, statusDescription, cancellationToken);
             }
 
-            var buffer = ArrayPool<byte>.Shared.Rent(1024);
+            var buffer = ArrayPool<byte>.Shared.Rent(_rentedBufferSize);
             try
             {
                 while (State == WebSocketState.CloseSent)
--- a/src/Microsoft.AspNetCore.Owin/WebSockets/OwinWebSocketAdapter.cs
+++ b/src/Microsoft.AspNetCore.Owin/WebSockets/OwinWebSocketAdapter.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
@@ -126,11 +127,18 @@ namespace Microsoft.AspNetCore.Owin
                 await CloseOutputAsync(closeStatus, statusDescription, cancellationToken);
             }
 
-            byte[] buffer = new byte[1024];
-            while (State == WebSocketState.CloseSent)
+            var buffer = ArrayPool<byte>.Shared.Rent(1024);
+            try
             {
-                // Drain until close received
-                await ReceiveAsync(new ArraySegment<byte>(buffer), cancellationToken);
+                while (State == WebSocketState.CloseSent)
+                {
+                    // Drain until close received
+                    await ReceiveAsync(new ArraySegment<byte>(buffer), cancellationToken);
+                }
+            }
+            finally
+            {
+                ArrayPool<byte>.Shared.Return(buffer);
             }
         }
 
--- a/src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
+++ b/src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
@@ -17,6 +17,7 @@ namespace Microsoft.AspNetCore.WebUtilities
     /// </summary>
     public class FileBufferingReadStream : Stream
     {
+        private const int _maxRentedBufferSize = 1024 * 1024; // 1MB
         private readonly Stream _inner;
         private readonly ArrayPool<byte> _bytePool;
         private readonly int _memoryThreshold;
@@ -25,7 +26,7 @@ namespace Microsoft.AspNetCore.WebUtilities
 
         private Stream _buffer;
         private byte[] _rentedBuffer;
-        private int _readIntoBuffer = 0;
+        private int _readIntoBuffer;
         private bool _inMemory = true;
         private bool _completelyBuffered;
 
@@ -57,7 +58,7 @@ namespace Microsoft.AspNetCore.WebUtilities
             }
 
             _bytePool = bytePool;
-            if (memoryThreshold < 1024 * 1024)
+            if (memoryThreshold < _maxRentedBufferSize)
             {
                 _rentedBuffer = bytePool.Rent(memoryThreshold);
                 _buffer = new MemoryStream(_rentedBuffer);
@@ -95,7 +96,7 @@ namespace Microsoft.AspNetCore.WebUtilities
             }
 
             _bytePool = bytePool;
-            if (memoryThreshold < 1024 * 1024)
+            if (memoryThreshold < _maxRentedBufferSize)
             {
                 _rentedBuffer = bytePool.Rent(memoryThreshold);
                 _buffer = new MemoryStream(_rentedBuffer);
--- a/src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
+++ b/src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
@@ -25,6 +25,7 @@ namespace Microsoft.AspNetCore.WebUtilities
 
         private Stream _buffer;
         private byte[] _rentedBuffer;
+        private int _readIntoBuffer = 0;
         private bool _inMemory = true;
         private bool _completelyBuffered;
 
@@ -73,6 +74,7 @@ namespace Microsoft.AspNetCore.WebUtilities
 
         // TODO: allow for an optional buffer size limit to prevent filling hard disks. 1gb?
         public FileBufferingReadStream(Stream inner, int memoryThreshold, string tempFileDirectory)
+            : this (inner, memoryThreshold, tempFileDirectory, ArrayPool<byte>.Shared)
         {
         }
 
@@ -177,26 +179,27 @@ namespace Microsoft.AspNetCore.WebUtilities
         public override int Read(byte[] buffer, int offset, int count)
         {
             ThrowIfDisposed();
-            if (_buffer.Position < _buffer.Length || _completelyBuffered)
+            if (_buffer.Position < _readIntoBuffer || _completelyBuffered)
             {
                 // Just read from the buffer
-                return _buffer.Read(buffer, offset, (int)Math.Min(count, _buffer.Length - _buffer.Position));
+                return _buffer.Read(buffer, offset, (int)Math.Min(count, _readIntoBuffer - _buffer.Position));
             }
 
             int read = _inner.Read(buffer, offset, count);
 
-            if (_inMemory && _buffer.Length + read > _memoryThreshold)
+            if (_inMemory && _readIntoBuffer + read > _memoryThreshold)
             {
                 _inMemory = false;
                 checked
                 {
-                    int length = (int)_buffer.Length;
+                    int length = _readIntoBuffer;
                     _buffer = CreateTempFile();
                     _buffer.Write(_rentedBuffer, 0, length);
                     _bytePool.Return(_rentedBuffer);
                     _rentedBuffer = null;
                 }
             }
+            _readIntoBuffer += read;
 
             if (read > 0)
             {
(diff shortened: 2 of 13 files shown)


stephentoub:
> This is because  ValueTask  isn't supported in net451

Why? The main point of putting it in a new assembly was so it would be usable there:
https://github.com/dotnet/corefx/blob/master/src/System.Threading.Tasks.Extensions/src/System.Threading.Tasks.Extensions.csproj#L9

benaadams:
> > This is because ValueTask isn't supported in net451

> Why? The main point of putting it in a new assembly was so it would be usable there:

Entirely my fault as I couldn't follow instructions; when it said it needed `"System.Threading.Tasks": "4.0.0.0"` it literally meant exactly that, not  `"System.Threading.Tasks": "4.0.0-*"`

benaadams:
Current

current

Buffers + ValueTask

valuetask

benaadams:
Reduces Task allocation by FormReader to 1% https://github.com/aspnet/HttpAbstractions/issues/553#issuecomment-182826023

benaadams:
Pre-for 3,000 requests 44MB on MemoryStream

pre-alloc

Post for 30,000 requests - `byte[]` allocations are as follows 0.5MB on rent vs 440MB on MemoryStream originally

allocs

rynowak:
Nice :+1:

benaadams:
Rebased; removed WebEncoders changes as its changed somewhat.

SajayAntony:
Sweet :+1: /cc @DavidObando @mnltejaswini

benaadams:
Need to sort out some `System.Runtime` references :-/

DavidObando:
Love this.

src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
```diff
@@ -43,6 +56,17 @@ public class FileBufferingReadStream : Stream
                 throw new ArgumentNullException(nameof(tempFileDirectoryAccessor));
             }
 
+            _bytePool = bytePool;
+            if (memoryThreshold < 1024 * 1024)
```
rynowak: Would be nice to put this in a constant.

src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
```diff
@@ -17,11 +18,14 @@ namespace Microsoft.AspNetCore.WebUtilities
...
         private readonly Stream _inner;
+        private readonly ArrayPool<byte> _bytePool;
         private readonly int _memoryThreshold;
         private string _tempFileDirectory;
         private readonly Func<string> _tempFileDirectoryAccessor;
 
-        private Stream _buffer = new MemoryStream(); // TODO: We could have a more efficiently expanding buffer stream.
+        private Stream _buffer;
+        private byte[] _rentedBuffer;
+        private int _readIntoBuffer = 0;
```
rynowak: Default init not needed

src/Microsoft.AspNetCore.WebUtilities/FormReader.cs
```diff
@@ -44,6 +61,9 @@ public FormReader(Stream stream, Encoding encoding)
                 throw new ArgumentNullException(nameof(encoding));
             }
 
+            _buffer = charPool.Rent(8192);
```
rynowak: Would be nice to put this in a constant - same with the buffer for streamreader

src/Microsoft.AspNetCore.WebUtilities/FormReader.cs
```diff
@@ -186,7 +263,7 @@ private void Buffer()
         /// <param name="stream">The HTTP form body to parse.</param>
         /// <param name="cancellationToken">The <see cref="CancellationToken"/>.</param>
         /// <returns>The collection containing the parsed HTTP form body.</returns>
-        public static Task<Dictionary<string, StringValues>> ReadFormAsync(Stream stream, CancellationToken cancellationToken = new CancellationToken())
+        public static ValueTask<Dictionary<string, StringValues>> ReadFormAsync(Stream stream, CancellationToken cancellationToken = default(CancellationToken))
```
rynowak: What's the impact of putting `ValueTask<>` in a public API - can mere mortals still call this method?

src/Microsoft.AspNetCore.WebUtilities/FormReader.cs
```diff
@@ -198,19 +275,69 @@ private void Buffer()
...
             }
+        }
 
-            return accumulator.GetResults();
+        private static async Task<Dictionary<string, StringValues>> ReadFormAsyncAwaited(Task<KeyValuePair<string, string>?> pairTask, FormReader reader, KeyValueAccumulator accumulator, CancellationToken cancellationToken)
+        {
+            using (reader)
+            {
+                var pair = await pairTask;
+                while (pair.HasValue)
```
rynowak: how does this work? Am I missing something that way over my head?

src/Microsoft.AspNetCore.WebUtilities/HttpRequestStreamReader.cs
```diff
@@ -34,40 +34,13 @@ public class HttpRequestStreamReader : TextReader
         private bool _isBlocked;
 
         public HttpRequestStreamReader(Stream stream, Encoding encoding)
-            : this(stream, encoding, DefaultBufferSize)
+            : this(stream, encoding, DefaultBufferSize, ArrayPool<byte>.Shared, ArrayPool<char>.Shared)
```
rynowak: :heart_eyes_cat:

src/Microsoft.AspNetCore.WebUtilities/StreamHelperExtensions.cs
```diff
@@ -9,14 +10,25 @@ namespace Microsoft.AspNetCore.WebUtilities
 {
     public static class StreamHelperExtensions
```
rynowak: Do we really have these as public extensions on stream?

src/Microsoft.AspNetCore.WebUtilities/FormReader.cs
```diff
@@ -44,6 +61,9 @@ public FormReader(Stream stream, Encoding encoding)
                 throw new ArgumentNullException(nameof(encoding));
             }
 
+            _buffer = charPool.Rent(8192);
+            _charPool = charPool;
+            _bytePool = bytePool;
             _reader = new StreamReader(stream, encoding, detectEncodingFromByteOrderMarks: true, bufferSize: 1024 * 2, leaveOpen: true);
```
rynowak: Should we use `HttpRequestBodyStreamReader` here?

src/Microsoft.AspNetCore.WebUtilities/FormReader.cs
benaadams: Lol, yes is fine and happily awaitable

src/Microsoft.AspNetCore.WebUtilities/StreamHelperExtensions.cs
benaadams: Yes :)

src/Microsoft.AspNetCore.WebUtilities/FormReader.cs
benaadams: ¯\\_(ツ)_/¯

benaadams:
Rebased, somewhat squashed, feedback made

@rynowak you asked a question on one of the algos, but it seemed to get lost in the rebase; can you ask again?

benaadams:
@rynowak found it in email; will reply in-line

src/Microsoft.AspNetCore.WebUtilities/FormReader.cs
```diff
@@ -198,19 +276,69 @@ private void Buffer()
...
             }
+        }
 
-            return accumulator.GetResults();
+        private static async Task<Dictionary<string, StringValues>> ReadFormAsyncAwaited(Task<KeyValuePair<string, string>?> pairTask, FormReader reader, KeyValueAccumulator accumulator, CancellationToken cancellationToken)
+        {
+            using (reader)
+            {
+                var pair = await pairTask;
+                while (pair.HasValue)
```
benaadams: > how does this work? Am I missing something that way over my head?

@rynowak do you mean the splitting of the `async` version from the non-`async` version (until it becomes async) or the end of read being indicated by a `null`?

rynowak: I just hadn't read these two lines super well.

davidfowl:
@benaadams Can you rebase?

src/Microsoft.AspNetCore.WebUtilities/FormReader.cs
benaadams: I imagine a golden time in the future when the compiler automatically does these patterns for you...

benaadams:
Rebased; note: I had to bump ` src/Microsoft.Net.Http.Headers/project.json` from `"netstandard1.0"` to `"netstandard1.1"`

src/Microsoft.AspNetCore.WebUtilities/project.json
```diff
@@ -16,12 +16,14 @@
...
-    "System.Text.Encodings.Web": "4.0.0-*"
+    "System.Text.Encodings.Web": "4.0.0-*",
+    "System.Threading.Tasks.Extensions": "4.0.0-*"
   },
   "frameworks": {
     "net451": {
       "frameworkAssemblies": {
-        "System.Runtime": { "type": "build" }
+        "System.Runtime": { "type": "build" },
+        "System.Threading.Tasks": "4.0.0.0"
```
davidfowl: This needs to be type build like the other facades

benaadams: Done

src/Microsoft.AspNetCore.Owin/WebSockets/OwinWebSocketAdapter.cs
```diff
@@ -126,11 +128,18 @@ public override Task SendAsync(ArraySegment<byte> buffer, WebSocketMessageType m
...
+            try
             {
-                // Drain until close received
-                await ReceiveAsync(new ArraySegment<byte>(buffer), cancellationToken);
+                while (State == WebSocketState.CloseSent)
+                {
+                    // Drain until close received
+                    await ReceiveAsync(new ArraySegment<byte>(buffer), cancellationToken);
+                }
+            }
```
muratg: Would this ever actually throw?

benaadams: Potentially, could throw if socket already disconnected or cancellation token cancelled.

src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
```diff
@@ -135,22 +180,27 @@ private Stream CreateTempFile()
         public override int Read(byte[] buffer, int offset, int count)
         {
             ThrowIfDisposed();
-            if (_buffer.Position < _buffer.Length || _completelyBuffered)
+            if (_buffer.Position < _readIntoBuffer || _completelyBuffered)
```
Tratcher: Why was this changed?

benaadams: The buffer pool can return arrays larger than you requested, so we use the requested length rather than the array length

Tratcher: MemoryStream.Length is not the length of the buffer, it's the length of the data. Contrast with Capacity which is the length of the buffer.

benaadams: Bleh, sorry has been a while since this PR :)

Using a fixed sized memory stream rather than a resizable one; length in the fixed size case is the size of the underlying array not its used space (different behavior between the two inits :-/)

Tratcher: Ah, it assumes the `byte[]` started full of data. You could call `SetLength(0)` just after creating the stream to reset it.

benaadams: Hmm... poked around in MemoryStream, looks like that would work; had some issues when trying to get it to work before but maybe I was changing capacity

Tratcher:
This ValueTask pattern causes substantial code duplication and maintenance issues. I don't think it's piratical. Can we split this and just take the buffer pooling?

benaadams:
The ValueTask does cut the allocations on 30k requests from 750MB to 7MB though

/cc @rynowak

Tratcher:
The buffer pooling is ready to merge and unrelated to the ValueTask work. Let's get that merged.

We need to come up with a simpler pattern if we're going to use ValueTasks.

benaadams:
K, will split them later today

benaadams:
> We need to come up with a simpler pattern if we're going to use ValueTasks.

Know what you mean, there are a bunch of higher performance patterns with async that do make the code more confusing as you need to split the code into multiple fragments.

I wonder if there is scope to push this into the compiler instead? As it already does that for general async.

Would rosyln be the correct repo to raise this? /cc @stephentoub

stephentoub:
> Would roslyn be the correct repo to raise this?

https://github.com/dotnet/roslyn/issues/7169

Tratcher:
I recommend we close this pending the roslyn discussion. @benaadams @rynowak?

rynowak:
That seems fine, I don't think we're going to take this PR in it's current form

PR #558, opened 2016-02-12
Add AuthenticationManager.HttpContext. Clean up contructors.

Tratcher:
@benaadams @muratg @HaoK 
Adding AuthenticationManager.HttpContext to mirror HttpRequest.HttpContext. We want to do some extensions on AuthenticationManager that need access to HttpContext.User.

Also cleaning up some of the constructors since they no longer need to pass both HttpContext and IFeatureCollection. We exposed HttpContext.Features directly.

### DIFF
--- a/samples/SampleApp/PooledHttpContext.cs
+++ b/samples/SampleApp/PooledHttpContext.cs
@@ -21,11 +21,11 @@ namespace SampleApp
         {
             if (_pooledHttpRequest != null)
             {
-                _pooledHttpRequest.Initialize(this, Features);
+                _pooledHttpRequest.Initialize(this);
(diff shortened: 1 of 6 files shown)


davidfowl:
What extension methods?

Tratcher:
These: https://github.com/aspnet/Security/pull/696/files#diff-3250882a7bd6ad3e385666151cf12fb4

Tratcher:
OK, so I ended up not requiring AuthenticationManager.HttpContext, but I don't see any problem with adding it.  We should still take the other cleanup changes regardless.

HaoK:
Lets add the context, it doesn't hurt and may be useful eventually... :shipit:

PR #559, opened 2016-02-15
UninitializeXxx() should null out, rather than Uninitialize()

benaadams:
To allow finer grained control of what is nulled out


benaadams:
nvm

benaadams:
Not sure which way you want to go; is currently doing a double clear

benaadams:
Otherwise `private FeatureReferences<FeatureInterfaces> _features;` should be `protected`

benaadams:
G'ah can't make this approach work without making the internals protected

PR #563, opened 2016-02-16
Extend `WebEncoders` API to avoid allocations within the methods

dougbu:
- rewrite existing methods in terms of the new ones
- don't allocate multiple 0-length arrays

nit:
- clarify a couple of doc comments e.g. using `<paramref/>`
- move an error message into a resource
 - pass parameter names into new resource
- rename parameters for consistency e.g. `inputLength` -> `count`


src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
```diff
@@ -48,50 +51,125 @@ public static byte[] Base64UrlDecode(string input, int offset, int count)
...
+            // Add the padding characters back.
+            for (; paddingCharsToAdd > 0; i++, paddingCharsToAdd--)
+            {
+                input[i] = '=';
+            }
+
+            // Decode.
             // If the caller provided invalid base64 chars, they'll be caught here.
-            return Convert.FromBase64CharArray(completeBase64Array, 0, completeBase64Array.Length);
+            return Convert.FromBase64CharArray(input, offset, arraySizeRequired);
```
dougbu: Yes, this call allocates. `Convert` doesn't offer a way around it.

src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
```diff
@@ -48,50 +51,125 @@ public static byte[] Base64UrlDecode(string input, int offset, int count)
...
             {
-                return new byte[0];
+                return EmptyBytes;
+            }
+
+            // Copy to an array large enough to Base64 characters (not just shorter Base64-URL-encoded form).
+            char[] completeBase64Array = new char[GetArraySizeRequiredToDecode(count)];
+            input.CopyTo(offset, completeBase64Array, 0, count);
+
+            return Base64UrlDecode(completeBase64Array, 0, count);
```
Tratcher: shouldn't count be modified here to completeBase64Array.Length?

src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
```diff
@@ -48,50 +51,125 @@ public static byte[] Base64UrlDecode(string input, int offset, int count)
...
-            input.CopyTo(offset, completeBase64Array, 0, count);
-            for (int i = 1; i <= numPaddingCharsToAdd; i++)
+            int paddingCharsToAdd = GetNumBase64PaddingCharsToAddForDecode(count);
+            var arraySizeRequired = checked(count + paddingCharsToAdd);
+            Debug.Assert(arraySizeRequired % 4 == 0, "Invariant: Array length must be a multiple of 4.");
+
+            if (input.Length - offset < arraySizeRequired)
             {
-                completeBase64Array[completeBase64Array.Length - i] = '=';
+                throw new ArgumentException(
```
Tratcher: Blah, wrong formatting for this repo

src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
```diff
@@ -48,50 +51,125 @@ public static byte[] Base64UrlDecode(string input, int offset, int count)
...
+
+            return Base64UrlDecode(completeBase64Array, 0, count);
+        }
+
+        /// <summary>
+        /// Decodes a base64url-encoded <paramref name="input"/> into a <c>byte[]</c>.
+        /// </summary>
+        /// <param name="input">
+        /// The <see cref="char"/>s to decode. Array must be large enough to hold <paramref name="offset"/> and
+        /// <paramref name="count"/> characters as well as Base64 padding characters. Content is not preserved.
```
Tratcher: It's really bad form to modify your input array.

Tratcher:
Why are you making this change?

src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
dougbu: Huh? What's the "right" formatting in this repo?

src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
dougbu: Which is why this overload is heavily-commented.

dougbu:
> Why are you making this change?

Because the allocations done inside `WebEncoders` are significant for the Antiforgery use of this code. See numbers in aspnet/Antiforgery#56.

davidfowl:
Seems like it might be worthwhile to consider a low allocation pattern where the user provides all of the input and output structures. They way they can pool arrays or instead of us having to new things up. The only issue might be knowing what the size of the output array is. I wonder of there are any nice patterns for doing this.

src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
rynowak: Yeah, it would be way less surprising to take an input and output array separately.

dougbu: Completely agree but it's not useful because the "output array" is always allocated within `Convert.FromBase64CharArray()`. One option would be
``` c#
public static byte[] Base64UrlDecode(
    string input, // All callers start w/ a string; char[] choice below was due only to reuse as an intermediate buffer.
    int offset,
    int count,
    char[] buffer) // Intermediate buffer of at least size GetArraySizeRequiredToDecode() returns.
{
    ...
}
```

Thoughts?

rynowak: Seems slightly better.

dougbu:
:up::date:

src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
```diff
@@ -124,38 +216,104 @@ public static string Base64UrlEncode(byte[] input, int offset, int count)
...
+        /// <paramref name="input"/>, including padding characters.
+        /// </param>
+        /// <param name="outputOffset">
+        /// The offset into <paramref name="output"/> at which to begin writing the base64url-encoded form of
+        /// <paramref name="input"/>.
+        /// </param>
+        /// <returns>
+        /// The number of characters written to <paramref name="output"/>, less any padding characters.
+        /// </returns>
+        public static int Base64UrlEncode(byte[] input, int offset, int count, char[] output, int outputOffset)
```
rynowak: By convention, the `count` in these methods always comes last (`Buffer.BlockCopy`, `string.Compare`)

src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
```diff
@@ -48,50 +51,139 @@ public static byte[] Base64UrlDecode(string input, int offset, int count)
...
 
             // Assumption: input is base64url encoded without padding and contains no whitespace.
 
-            // First, we need to add the padding characters back.
-            int numPaddingCharsToAdd = GetNumBase64PaddingCharsToAddForDecode(count);
-            char[] completeBase64Array = new char[checked(count + numPaddingCharsToAdd)];
-            Debug.Assert(completeBase64Array.Length % 4 == 0, "Invariant: Array length must be a multiple of 4.");
-            input.CopyTo(offset, completeBase64Array, 0, count);
-            for (int i = 1; i <= numPaddingCharsToAdd; i++)
+            int paddingCharsToAdd = GetNumBase64PaddingCharsToAddForDecode(count);
```
rynowak: There's a healthy mix of `var` and non-var in the new codesss

rynowak:
:shipit: from me with some minor comments - this will meet our needs. 

Another option here is to use `stackalloc` with the intermediate array, but that has its own gotchas.

src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
dougbu: Not always and I followed what `Convert.ToBase64CharArray()` exposes. See https://github.com/aspnet/HttpAbstractions/pull/563/files#diff-3fdfb7eef02148bf1131ee1c77af1fe6R282 But I don't feel strongly. Do you?

src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
dougbu: Thx, I missed that aspect of the code I was changing.

src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
dougbu: Thought about this a bit more, `Convert.ToBase64CharArray()` is the odd one out. I'll move `count` to the end here and in the new `Base64UrlDecode()` overload.

dougbu:
8c120a0

dougbu:
BTW @rynowak 

> Another option here is to use `stackalloc` with the intermediate array

This option is now open for callers. For example, if the DataProtections use of `WebEncoders` in [`public static string Protect(this IDataProtector protector, string plaintext)`](https://github.com/aspnet/DataProtection/blob/342676925ff23cf823b9234772e77c6b9565f542/src/Microsoft.AspNetCore.DataProtection.Abstractions/DataProtectionExtensions.cs#L220) or [`public static string Unprotect(this IDataProtector protector, string protectedData)`](https://github.com/aspnet/DataProtection/blob/342676925ff23cf823b9234772e77c6b9565f542/src/Microsoft.AspNetCore.DataProtection.Abstractions/DataProtectionExtensions.cs#L254) show up in tested scenarios, we could use `stackalloc` there.

Issue #542, opened 2016-01-21
Add a ConnectionId property to IHttpConnectionFeature to help correlate logs


halter73:
Adding a ConnectionId property to IHttpConnectionFeature, should enable application logs to be correlated with server logs.

See the original discussion here: https://github.com/aspnet/KestrelHttpServer/pull/592#issuecomment-173532535

PR #564, opened 2016-02-17
Add IHttpConnectionFeature.ConnectionId

Tratcher:
#542 
@halter73

### DIFF
--- a/src/Microsoft.AspNetCore.Owin/OwinConstants.cs
+++ b/src/Microsoft.AspNetCore.Owin/OwinConstants.cs
@@ -73,6 +73,7 @@ namespace Microsoft.AspNetCore.Owin
             public const string RemotePort = "server.RemotePort";
             public const string LocalIpAddress = "server.LocalIpAddress";
             public const string LocalPort = "server.LocalPort";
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Owin/OwinConstants.cs
```diff
@@ -73,6 +73,7 @@ internal static class CommonKeys
             public const string RemotePort = "server.RemotePort";
             public const string LocalIpAddress = "server.LocalIpAddress";
             public const string LocalPort = "server.LocalPort";
+            public const string ConnectionId = "server.ConnectionId";
```
davidfowl: Did you make this up?

davidfowl:
:shipit:

src/Microsoft.AspNetCore.Owin/OwinConstants.cs
Tratcher: Yes. It's ok, I made up many of the other ones too :-)

davidfowl: Nice

PR #566, opened 2016-02-18
Reset fields on AuthenticateContext

rynowak:
This is needed for cases where IAuthenticationHandler instances delegate
or modify the output of each-other.


Tratcher:
@HaoK do you think this is going to have any interesting side-effects?

rynowak:
I think if we had any code that called multiple different of these methods already we would have found this :laughing: 

Claims transformer calls `Authenticated` again only if the context is already authenticated, which works in the existing code.

HaoK:
Yeah this change seems fine, before we were not expecting these to be modified more than once, basically expectation was one of these methods is called and that's it...

Tratcher:
:shipit:

rynowak:
Ok cool, does someone want to stamp a :shipit: on here then?

rynowak:
beat me to it :laughing:

PR #568, opened 2016-02-18
Enabled xml doc generation

ajaybhargavb:
@NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/HeaderDictionaryExtensions.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/HeaderDictionaryExtensions.cs
@@ -11,6 +11,7 @@ namespace Microsoft.AspNetCore.Http
         /// <summary>
         /// Add new values. Each item remains a separate array entry.
         /// </summary>
+        /// <param name="headers">The <see cref="IHeaderDictionary"/> to use.</param>
         /// <param name="key">The header name.</param>
         /// <param name="value">The header value.</param>
         public static void Append(this IHeaderDictionary headers, string key, StringValues value)
@@ -21,6 +22,7 @@ namespace Microsoft.AspNetCore.Http
         /// <summary>
         /// Quotes any values containing comas, and then coma joins all of the values with any existing values.
         /// </summary>
+        /// <param name="headers">The <see cref="IHeaderDictionary"/> to use.</param>
         /// <param name="key">The header name.</param>
         /// <param name="values">The header values.</param>
         public static void AppendCommaSeparatedValues(this IHeaderDictionary headers, string key, params string[] values)
@@ -32,6 +34,7 @@ namespace Microsoft.AspNetCore.Http
         /// Get the associated values from the collection separated into individual values.
         /// Quoted values will not be split, and the quotes will be removed.
         /// </summary>
+        /// <param name="headers">The <see cref="IHeaderDictionary"/> to use.</param>
         /// <param name="key">The header name.</param>
         /// <returns>the associated values from the collection separated into individual values, or StringValues.Empty if the key is not present.</returns>
         public static string[] GetCommaSeparatedValues(this IHeaderDictionary headers, string key)
@@ -42,6 +45,7 @@ namespace Microsoft.AspNetCore.Http
         /// <summary>
         /// Quotes any values containing comas, and then coma joins all of the values.
         /// </summary>
+        /// <param name="headers">The <see cref="IHeaderDictionary"/> to use.</param>
         /// <param name="key">The header name.</param>
         /// <param name="values">The header values.</param>
         public static void SetCommaSeparatedValues(this IHeaderDictionary headers, string key, params string[] values)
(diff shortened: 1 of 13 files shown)


src/Microsoft.AspNetCore.Http.Abstractions/Extensions/HeaderDictionaryExtensions.cs
```diff
@@ -11,6 +11,7 @@ public static class HeaderDictionaryExtensions
         /// <summary>
         /// Add new values. Each item remains a separate array entry.
         /// </summary>
+        /// <param name="headers">The <see cref="IHeaderDictionary"/> that this method extends.</param>
```
Eilon: Not sure I understand the current comment - this method isn't "extending the dictionary".

How about something like: `The <see cref="IHeaderDictionary"/> to append the specified values to.`?

@Tratcher can you comment on this comment?

Eilon:
The change in general is :shipit: but need to get that doc comment update clarified.

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/HeaderDictionaryExtensions.cs
Tratcher: It's named Append because it will append to existing values for the same key (if any). You're not just appending to the dictionary.

src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
```diff
@@ -156,7 +156,7 @@ public static byte[] Base64UrlDecode(string input, int offset, char[] buffer, in
 
         /// <summary>
         /// Gets the minimum <c>char[]</c> size required for decoding of <paramref name="count"/> characters
-        /// with the <see cref="Base64UrlDecode(char[], int, int)"/> method.
+        /// with the <see cref="M:Base64UrlDecode"/> method.
```
dougbu: Should be `Base64UrlDecode(string, int, char[], int, int)`. Definitely should not refer to the method group because the other overloads do not have a `char[]` parameter. (I neglected to update the signature here when I changed that method's parameters.)

src/Microsoft.AspNetCore.WebUtilities/WebEncoders.cs
```diff
@@ -305,7 +305,7 @@ public static int Base64UrlEncode(byte[] input, int offset, char[] output, int o
 
         /// <summary>
         /// Get the minimum output <c>char[]</c> size required for encoding <paramref name="count"/>
-        /// <see cref="byte"/>s with the <see cref="Base64UrlEncode(byte[], int, int, char[], int)"/> method.
+        /// <see cref="byte"/>s with the <see cref="M:Base64UrlEncode"/> method.
```
dougbu: `Base64UrlEncode(byte[], int, char[], int, int)`. Again, the problem was in the signature and not the reference to one overload.

src/Microsoft.AspNetCore.WebUtilities/MultipartReaderStream.cs
```diff
@@ -25,6 +25,7 @@ internal class MultipartReaderStream : Stream
         /// </summary>
         /// <param name="stream"></param>
         /// <param name="boundary"></param>
+        /// <param name="expectLeadingCrlf"></param>
```
dougbu: Please fill these in.

dougbu:
:shipit: with fixes requested above.

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/HeaderDictionaryExtensions.cs
ajaybhargavb: Isn't this method an extension method to `IHeaderDictionary`?

Eilon: @ajaybhargavb yes but the docs for an extension method don't say "it's extending type Foo", rather, they say what the method does, i.e. "frobs a fribble using a frabulator."

src/Microsoft.AspNetCore.WebUtilities/MultipartReaderStream.cs
Eilon: This type is `internal`...

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/HeaderDictionaryExtensions.cs
ajaybhargavb: I agree. But this is the doc for the param and not the method itself.

Eilon: Adding/modifying a dictionary or whatever isn't "extending" it, though. It's just adding/modifying stuff in it...

ajaybhargavb: Okay. How about making it simple and generic?
 `The <see cref="IHeaderDictionary"/> to use.`

Eilon: Sure.

PR #569, opened 2016-02-19
[Fixes #567] Adding port and domain parsing in HostString

### LINKED ISSUE #567: Expose port and domain in HostString
Talked with @Tratcher and he suggested this.
It's missing for some code that needs to manipulate them without implementing our own logic. c.f. https://github.com/aspnet/Mvc/pull/4113

sebastienros:


src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -25,6 +26,31 @@ public HostString(string value)
         }
 
         /// <summary>
+        /// Creates a new HostString without modification.
+        /// </summary>
+        /// <param name="value">The value should be Unicode rather than punycode. IPv6 addresses must use square braces.</param>
```
dougbu: > `value`

Really? Looks more like the `domain` from how it's handled.

> must use square braces

No "must"; the constructor adds missing square braces.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -25,6 +26,31 @@ public HostString(string value)
...
+            if (value.IndexOf('[') == -1
+                && (index = value.IndexOf(':')) >= 0
+                && index < value.Length - 1
+                && value.IndexOf(':', index + 1) >= 0)
+            {
+                // IPv6 without brackets ::1 is the only type of host with 2 or more colons
+                value =  $"[{value}]";
+            }
+
+            _value = value + ":" + port;
```
dougbu: Why is this done if `!port.HasValue`?

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -25,6 +26,31 @@ public HostString(string value)
...
         /// <summary>
+        /// Creates a new HostString without modification.
+        /// </summary>
+        /// <param name="value">The value should be Unicode rather than punycode. IPv6 addresses must use square braces.</param>
+        /// <param name="port">A positive, greater than 0 value representing the port in the host string.</param>
+        public HostString(string value, int? port)
+        {
+            if(port.HasValue && port <= 0)
+            {
+                throw new ArgumentException(Resources.FormatException_PortMustBeGreaterThanZero(nameof(port)), nameof(port));
```
dougbu: Why does the resource include the parameter name? That'll be in the final `Message` regardless.

dougbu: Suggest `ArgumentOutOfRangeException`.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -38,6 +64,92 @@ public bool HasValue
...
+
+                // normalize the value
+                var uriComponent = ToUriComponent();
+
+                int index;
+                int port = 0;
+                if ((index = uriComponent.IndexOf(']')) >= 0)
+                {
+                    // IPv6 in brackets [::1], maybe with port
+                    if ((index = uriComponent.IndexOf(':', index)) >= 0)
```
dougbu: Is it legal for the colon before the port to appear anywhere other than immediately after the closing square bracket? If not, the `uriComponent.Length == index + 1` check in `Domain` then checking `uriComponent[index + 1] == ':'` seems more straightforward.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -38,6 +64,92 @@ public bool HasValue
...
+                    return string.Empty;
+                }
+
+                // normalize the value
+                var uriComponent = ToUriComponent();
+
+                int index;
+                if ((index = uriComponent.IndexOf(']')) >= 0)
+                {
+                    // optimization if uriComponent is exactly the domain
```
dougbu: FYI `Substring()` does this optimization for you already. Won't allocate in the whole string case.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -38,6 +64,92 @@ public bool HasValue
...
+                    {
+                        // no port
+                        return uriComponent;
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Returns the value of the port part of the host, or <value>0<code> if none is found.
```
dougbu: Why doesn't this return `int?` and `null` in the no port case? That would align better with the new constructor.

dougbu:
:shipit:

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -25,6 +26,31 @@ public HostString(string value)
         }
 
         /// <summary>
+        /// Creates a new HostString without modification.
+        /// </summary>
+        /// <param name="value">The value should be Unicode rather than punycode. IPv6 addresses must use square braces.</param>
+        /// <param name="port">A positive, greater than 0 value representing the port in the host string.</param>
+        public HostString(string value, int? port)
```
Tratcher: Don't use a nullable. If there's no port then make them call the other constructor.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
Tratcher: ToString InvariantCulture

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -25,6 +26,31 @@ public HostString(string value)
...
+        {
+            if(port.HasValue && port <= 0)
+            {
+                throw new ArgumentException(Resources.FormatException_PortMustBeGreaterThanZero(nameof(port)), nameof(port));
+            }
+
+            int index;
+            if (value.IndexOf('[') == -1
+                && (index = value.IndexOf(':')) >= 0
```
Tratcher: how about `value.Count(':') >= 2`?

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -38,6 +64,92 @@ public bool HasValue
         }
 
         /// <summary>
+        /// Returns the value of the domain part of the host. The port is removed if it was present.
+        /// IPv6 addresses will have brackets added if they are missing.
+        /// </summary>
+        /// <returns></returns>
+        public string Domain
```
Tratcher: Officially it's called the Host, not the Domain. https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -38,6 +64,92 @@ public bool HasValue
...
+        {
+            get
+            {
+                if (string.IsNullOrEmpty(_value))
+                {
+                    return string.Empty;
+                }
+
+                // normalize the value
+                var uriComponent = ToUriComponent();
```
Tratcher: ToUriComponent returns the punycode form. We'd rather return the unicode form from this property. That's the form you'd get from Value and pass into the constructor.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
Tratcher: :+1:

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
sebastienros: I initially used `domain` like you said, but Chris preferred `value`. 2 vs 1 ;)

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
sebastienros: I followed an example in `PathString` and I assumed this was the convention. Will do.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
sebastienros: I used the same logic as in `ToUriComponent`

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
sebastienros: Should I use host instead then (based on another comment stating it's not domain but host actually)

Tratcher: Hmm
```
HostString(string value)
HostString(string host, int port)
```

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
sebastienros: Does it mean we should unencode it if the value contains punycode ?

Tratcher: No, _value is documented to be in it's decoded state.

Tratcher: FromUriComponent takes care of that.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -38,6 +64,88 @@ public bool HasValue
...
+                    return _value;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Returns the value of the port part of the host, or <value>0</value> if none is found.
+        /// </summary>
+        /// <returns></returns>
+        public int Port
```
Tratcher: `int?`

sebastienros: `0` means no port, you prefer `int?` ?

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -38,6 +64,88 @@ public bool HasValue
...
+                }
+                else
+                {
+                    // Has a port
+                    int.TryParse(_value.Substring(index + 1), out port);
+                }
+
+                return port;
+            }
+        }
```
Tratcher: There really is a lot of duplicate code here. How about sharing most of it with:
`private void GetParts(out string host, out string port)`

sebastienros: Ok.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
Tratcher: Yes, no magic values please. `var port = hostString.Port; if (port.HasValue)` is much better than `if (port != 0)`

Tratcher:
:shipit:

PR #571, opened 2016-02-20
Downgrade to use CoreFX NET46 compatible packages

mjsabby:
Similar to the change in DependencyInjection repo.


mjsabby:
cc @NTaylorMullen

NTaylorMullen:
Merged.

Issue #575, opened 2016-03-01
MultipartStreamReader has suboptimal boundary search in stream


DavidObando:
Method [SubMatch](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.WebUtilities/MultipartReaderStream.cs#L309) in class MultipartStreamReader has a naïve implementation of a substring search that make this method a hotpath in multpart upload scenarios.

In an ad-hoc test uploading a number of files ranging from 10 MB to 1 GB, method SubMatch consumed 53.4% exclusive samples during a profiling session.

![image](https://cloud.githubusercontent.com/assets/950290/13437003/e70812be-df96-11e5-8139-dedd1336c9cc.png)

Its current implementation has an O(NM) complexity that can be improved by using a faster algorithm such as [Boyer–Moore–Horspool](https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore%E2%80%93Horspool_algorithm).

Method SubMatch looks for two things:
  1. Whether the byte array segment fully contains the boundary, or
  2. Whether the byte array ends with the start of the boundary.

Case 1 can implement a faster algorithm. Case 2 can be more natural and check character by character as typically the boundary string is very short compared to the length of the byte array segment.

DavidObando:
Implemented a modified Boyer–Moore–Horspool algorithm with an average-case complexity of O(N) on random text. Worst case, it behaves similarly to the previous implementation O(MN), where M is the length of the boundary and N is the length of the buffer to operate on.

This took the cost down to 16.1%. I think I know how to bring it a bit lower, will try that. My skip table is currently a `Dictionary<byte, int>` when it could simply be a `byte[256]`.

![image](https://cloud.githubusercontent.com/assets/950290/13443961/7a408d9c-dfb8-11e5-9f26-bfb91f1b2820.png)

DavidObando:
Indeed, changing the skip table type from `Dictionary<byte, int>` to `int[256]` brought times down as expected. SubMatch now costs 7.6%. I'll run some more tests and then submit a PR.

![image](https://cloud.githubusercontent.com/assets/950290/13444446/31ed3de4-dfbb-11e5-874d-0a23ba1e3fae.png)

davidfowl:
W00t

DavidObando:
Fixed 15649b7.

PR #576, opened 2016-03-01
Faster SubMatch implementation

DavidObando:
SubMatch has been sped up by implementing a modified Boyer–Moore–Horspool algorithm with an average-case complexity of O(N) on random text. Worst case, it behaves similarly to the previous implementation O(MN), where M is the length of the boundary and N is the length of the buffer to operate on.

Method SubMatch looks for two things:

  1. Whether the byte array segment fully contains the boundary, or
  2. Whether the byte array ends with the start of the boundary.

Case 1 is now a lot faster than the previous implementation. Case 2 remains using the same code as before. The method will do Case 1 until the matchOffset is equal to N-M. It then switches to Case 2, unless a match is found.

The code can be further sped up with a full Boyer–Moore implementation, or something more sophisticated. This however can be evaluated in the case that this implementation is insufficiently performant for our main scenarios.

This commit resolves issue #575.

Using test [MultipartPOST](https://github.com/aspnet/Performance/tree/dev/testapp/MultipartPOST) the work percentage (inclusive %) for method SubMatch under the profiler was:
  * Before this change: 53.4%
  * After this change: 7.6%

@SajayAntony @Tratcher @pranavkm @davidfowl


src/Microsoft.AspNetCore.WebUtilities/MultipartReaderStream.cs
```diff
@@ -40,15 +38,8 @@ public MultipartReaderStream(BufferedReadStream stream, string boundary, bool ex
...
-            {
-                _boundaryBytes = Encoding.UTF8.GetBytes("\r\n--" + boundary);
-            }
-            else
-            {
-                _boundaryBytes = Encoding.UTF8.GetBytes("--" + boundary);
-            }
-            _finalBoundaryLength = _boundaryBytes.Length + 2; // Include the final '--' terminator.
+            _boundary = boundary;
+            _boundary.ExpectLeadingCrlf = expectLeadingCrlf;
```
Tratcher: It's weird to take a constructor parameter that only modifies another parameter, especially one that already takes the same parameter.

davidfowl: Yea it should be set from the caller on the multipart boundary object

SajayAntony: I can't understand why is there an optional parameter. But with your change it doesn't make sense to pass it in anymore since it can be set on the `MultipartBoundary` object. 
@Tratcher @davidfowl couldn't we just remove this now and set it at the call site as mentioned.

DavidObando: Yup, I failed to see that. Will send an update to fix this.

SajayAntony:
Awesome drop in CPU.

DavidObando:
Just updated this PR to address feedback. Let me know if you see anything else. Thanks!

DavidObando:
Ping @SajayAntony @Tratcher @pranavkm @davidfowl

davidfowl:
Do we have any tests for this code? Unit tests would be best

DavidObando:
@davidfowl we do have tests for MultipartReader, and they seem adequate. Do you have any scenario in mind, I'd gladly add it.

davidfowl:
@DavidObando test cases for each of the cases called out in the algorithm.

DavidObando:
@davidfowl Just sent an update with a test. The only missing case was testing that a partial boundary at the end of the buffer still signals the end of a section, so I added a test for that (effectively exercising case 2). Case 1 is fully covered by this and other tests.

davidfowl:
Nice!

test/Microsoft.AspNetCore.WebUtilities.Tests/MultipartReaderTests.cs
```diff
@@ -218,6 +239,47 @@ private static MemoryStream MakeStream(string text)
...
+            var stream = MakeStream(TwoPartBodyIncompleteBuffer);
+            var reader = new MultipartReader(Boundary, stream);
+            var buffer = new byte[128];
+
+            //first section can be read successfully
+            var section = await reader.ReadNextSectionAsync();
+            Assert.NotNull(section);
+            Assert.Equal(1, section.Headers.Count);
+            Assert.Equal("form-data; name=\"text\"", section.Headers["Content-Disposition"][0]);
+            section.Body.Read(buffer, 0, buffer.Length);
```
davidfowl: why not use the value returned from read in `GetStringASCIIZ`?

davidfowl:
:shipit:

test/Microsoft.AspNetCore.WebUtilities.Tests/MultipartReaderTests.cs
DavidObando: @davidfowl Yes, thanks for the observation. Fixed now.

DavidObando:
Merged 15649b7.

PR #579, opened 2016-03-07
Don't reference design-time facades in nuspec on Full .NET

bricelam:
This may need to propagate into everything that depends on this... :cry:

See also dotnet/cli#164


Eilon:
Looks like the other changes, so :shipit:

pranavkm:
:shipit:

davidfowl:
:shipit:

PR #581, opened 2016-03-10
Don't reference facades in NuSpec

bricelam:
These can be removed entirely after dotnet/cli#164


bricelam:
:grimacing: Oops, this is a dupe of #581. Closing.

PR #584, opened 2016-03-11
Target dnx451

bricelam:
This reverts a051244faffe99302dad919827261b79aa99e487 for DNX


davidfowl:
Nah

Eilon:
@davidfowl what's the issue?

davidfowl:
Don't want DNX creeping into anything. We should just have an announcement telling people to add framework assemblies. The .NET CLI will handle the main cases for us.

Eilon:
@davidfowl this is temporary until we get closer to having tooling available.

bricelam:
Recalling. After chatting with @Eilon, we decided that adding the facades to user's `project.json` is an acceptable workaround.

Issue #561, opened 2016-02-16
`ResponseCookies.Append()` allocates lots, big lots


dougbu:
Despite using `StringValues`s, `ResponseCookies.Append()` is very allocation-heavy. Even just reusing the `StringBuilder` created in `SetCookieHeaderValue.ToString()` would be a significant improvement.

For example, running a scenario that adds an Antiforgery token to every response (~156 characters in the token in this case) allocates almost 1 KB extra _per response_. That's more than 2% of the allocations in this scenario.

Data for 3000 requests:
![responsecookies append strings](https://cloud.githubusercontent.com/assets/6431421/13084660/1be72258-d48f-11e5-9798-ccda36d1935f.png)
![responsecookies append chars](https://cloud.githubusercontent.com/assets/6431421/13084662/1e8c93a8-d48f-11e5-9e29-e35ccf0b64cb.png)

muratg:
@dougbu Are you able to send a PR to reuse the StringBuilder?

dougbu:
Sure. I'll do that today.

dougbu:
Offline discussion w/ @Tratcher: Reusing a `StringBuilder` per `ResponseCookies` instance won't do much because few cookies are set per request. Will instead add an object pool of a small class containing (at least for now) just a `StringBuilder` instance.

muratg:
@dougbu That makes sense.

dougbu:
For exposing the feature interfaces (no longer in `.Internal` namespaces):
- 6f24508
- aspnet/Hosting@bb3809d
- aspnet/KestrelHttpServer@f13c418
- aspnet/Security@9361960

dougbu:
Part two for the actual fix:
- @80813f7
- aspnet/Hosting@b06a844 (always register an `ObjectPoolProvider`)
- aspnet/WebListener@f0a5a01 (react to `HttpContextFactory` constructor change)
- aspnet/Antiforgery@0c81df8 (no need to register an `ObjectPoolProvider`)
- aspnet/Mvc@43836ae  (no need to register an `ObjectPoolProvider`)
- aspnet/Routing@39b4f6f  (no need to register an `ObjectPoolProvider`)

PR #587, opened 2016-03-15

dougbu:
- #561


src/Microsoft.Net.Http.Headers/SetCookieHeaderValue.cs
```diff
@@ -90,42 +90,53 @@ public string Value
...
+        }
+
+        /// <summary>
+        /// Add string representation of this <see cref="SetCookieHeaderValue"/> to given <paramref name="builder"/>.
+        /// </summary>
+        /// <param name="builder">
+        /// The <see cref="StringBuilder"/> to receive the string representation of this
+        /// <see cref="SetCookieHeaderValue"/>.
+        /// </param>
+        public void AddStringValue(StringBuilder builder)
```
Tratcher: This name is confusing. `ToStringBuilder`? `CopyToStringBuilder`? `SerializeToStringBuilder`? `AppendToStringBuilder`?

dougbu: Sure, let's go w/ `AppendToStringBuilder`.

Tratcher:
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.CookiePolicy/CookiePolicyMiddleware.cs#L29 will also need to be updated.

Looks OK. Check with @davidfowl on the new hosting service.

dougbu:
> [`CookiePolicyMiddleware`'s `ResponseCookieFeature` instantiation](https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.CookiePolicy/CookiePolicyMiddleware.cs#L29) will also need to be updated

That line is fine because `ResponseCookieFeature` has the same constructor it did before. I've successfully built aspnet/KestrelHttpServer and aspnet/Security with this and aspnet/Hosting#652 in place.

dougbu:
When using this branch and aspnet/Hosting#652, overall allocations go down approximately 1.7%. Big savings in `char[]` and `StringBuilder` instances:

###### Before
dev allocations

###### After
maker allocations

maker builders

dougbu:
Will come back to the pooling once we've got a few other ducks in line.

PR #589, opened 2016-03-17
Move remaining feature interfaces into `Microsoft.AspNetCore.Http.Features` package and namespace

dougbu:
- related to #561
- move required interfaces and classes down from `Microsoft.AspNetCore.Http.Abstractions`
- move everything in the `Microsoft.AspNetCore.Http.Features` into their own directory

nit: remove transient dependencies listed in `Microsoft.AspNetCore.Http.Abstractions`'s `project.json`


dougbu:
/cc @lodejard @davidfowl 
Will get reaction PRs (3 or 4) out later this evening.

src/Microsoft.AspNetCore.Http.Features/project.json
```diff
@@ -21,6 +21,7 @@
     "netstandard1.3": {
       "dependencies": {
         "System.Collections": "4.0.11-*",
+        "System.ComponentModel": "4.0.1-*",
```
davidfowl: Which type requires this?

davidfowl:
We should look at renaming some of these interfaces on the second pass.

/cc @Tratcher

src/Microsoft.AspNetCore.Http.Features/project.json
dougbu: `IServiceProvidersFeature`

dougbu: FYI I moved this from `src/Microsoft.AspNetCore.Http.Abstractions/project.json`, same package that `interface` came from to this one.

dougbu:
> We should look at renaming some of these interfaces on the second pass.

Just curious: Would the renaming be to add `Http`, to remove that, or something else?

In any case, I suggest filing a separate issue b/c such a renaming is unrelated to #561. That issue is specific to `IResponseCookiesFeature`.

davidfowl:
> Just curious: Would the renaming be to add Http, to remove that, or something else?

Probably adding http to some of them (the ones that relate to http things like form)

davidfowl:
@dougbu Any reason you put these in the root instead of where the other features are?

Tratcher:
Having a /Features directory inside the Features package is redundant.

dougbu:
> Having a /Features directory inside the Features package is redundant.

We're supposed to have a directory per namespace and @Eilon asked me to clean up this particular mix.

davidfowl:
I have to agree with @Tratcher, why is it in a features folder?

dougbu:
>  why is it in a features folder?

Huh? Are you suggesting creating a `.Root` folder for everything that's in the `Microsoft.AspNetCore.Http` namespace and putting that _below_ the `.Features` files? Or are you suggesting we ignore @Eilon's guidance?

Tratcher:
The guideline is to use folders for sub-namespaces. There's no guideline for components in parent namespaces. I think it's fine to put everything in the base directory in this case.

dougbu:
> it's fine to put everything in the base directory in this case

That means mixing `Microsoft.AspNetCore.Http` and `Microsoft.AspNetCore.Http.Features` in the root directory. Put another way, `Microsoft.AspNetCore.Http.Features` is a sub-namespace relative to the files in the root directory.

Note moving, say, `IFormCollection` into `Microsoft.AspNetCore.Http.Features` is a non-starter. It's more general than what we have in the sub-namespace. (Actually, should probably move `ISession` and perhaps `WebSocketAcceptContext` up into the root namespace and directory.)

@Eilon ?

Eilon:
Oh I didn't realize the files would effectively be in a `...Features.Features` folder. That's worse.

Just leave it top-level then.

dougbu:
> Just leave it top-level then.

:ok:

dougbu:
:up::date:

Tratcher:
Why not include #590?

dougbu:
> Why not include #590?

Because that class and interface were already in the `Microsoft.AspNetCore.Http.Features` package and the change suggested in #590 was unrelated. Plus I didn't know if #590 was controversial for some reason.

Tratcher:
Meh, do it so you only have to make one pass of updates.

dougbu:
> do it

:ok:

dougbu:
:up::date:

src/Microsoft.AspNetCore.Owin/OwinEnvironment.cs
```diff
@@ -16,9 +16,9 @@
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
-using Microsoft.AspNetCore.Http.Features.Internal;
 using Microsoft.AspNetCore.Http.Features.Authentication;
 using Microsoft.AspNetCore.Http.Features.Authentication.Internal;
+using Microsoft.AspNetCore.Http.Features.Internal;
```
davidfowl: We still have internal features?

Tratcher: Implementations

Tratcher:
:shipit:

src/Microsoft.AspNetCore.Owin/OwinEnvironment.cs
davidfowl: Those should go into an internal folder...

Tratcher: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http/Features/HttpConnectionFeature.cs

davidfowl:
I don't think these should be considered internal

dougbu:
> I don't think these should be considered internal

@davidfowl do you mean I should move all of the implementations into the `Microsoft.AspNetCore.Http.Features` namespace? And, unlike the `interface`s, no files will be moved in the process? (From what I've seen, the implementations are referenced much less than the `interface`s. But they're definitely used, most obviously in KestrelHttpServer.)

davidfowl:
> @davidfowl do you mean I should move all of the implementations into the Microsoft.AspNetCore.Http.Features namespace? And, unlike the interfaces, no files will be moved in the process? (From what I've seen, the implementations are referenced much less than the interfaces. But they're definitely used, most obviously in KestrelHttpServer.)

We can file a bug for this.

:shipit:

dougbu:
> We can file a bug for this.

#592

dougbu:
6f24508

PR #591, opened 2016-03-20

dougbu:
- #561
- new `SetCookieHeaderValue.AppendToStringBuilder()` method; avoids per-call `StringBuilder` allocation
- `ResponseCookies` uses `ObjectPool<StringBuilder>` that `ResponseCookiesFeature` provides
- `ResponseCookiesFeature` creates an `ObjectPoolProvider` instance if none in DI
- `IResponseCookiesFeature` exposes `int` properties supporting middleware overrides of the pool policy

nit: Add some doc comments


dougbu:
Note this is based on #589

/cc @davidfowl here's one idea for addressing #561. This does not require any changes in other repos. Middleware can adjust the pool policies but I doubt our middleware needs that.

###### Details
Looking more closely at the repos that add cookies, I suspect most, even Antiforgery, will always add cookie values shorter than the 4k default maximum size used here. In particular, the Antiforgery cookie token is shorter than I thought because it does not contain claims and most other token values are short identifiers. In addition `CookieOptions` shouldn't make cookie values significantly longer.

Probably worth measuring allocations when using the `TwitterMiddleware`. The `TwitterHandler` uses a `RequestToken` containing a fair amount of information. But it appears that cookie is a session cookie. (So making the test realistic may be difficult :smiling_imp:)

davidfowl:
This still assumes something is coming from services no?

dougbu:
> This still assumes something is coming from services no?

No. If an `ObjectPoolProvider` is a registered service, the feature uses it. But it'll fall back to creating its own `DefaultObjectPoolProvider` otherwise.

src/Microsoft.AspNetCore.Http.Features/IResponseCookiesFeature.cs
```diff
@@ -3,8 +3,33 @@
...
     public interface IResponseCookiesFeature
     {
+        /// <summary>
+        /// Initial capacity of instances used in the implementation of this feature, in <see cref="char"/>s.
+        /// </summary>
+        /// <remarks>
+        /// For example, the initial capacity of <see cref="System.Text.StringBuilder"/> instances obtained
+        /// from an object pool.
+        /// </remarks>
+        int InitialPooledInstanceCapacity { get; set; }
```
Tratcher: :-1: This is an implementation detail, not a generic feature property.

src/Microsoft.AspNetCore.Http/Features/ResponseCookiesFeature.cs
```diff
@@ -25,8 +49,17 @@ public IResponseCookies Cookies
                 if (_cookiesCollection == null)
                 {
                     var headers = HttpResponseFeature.Headers;
-                    _cookiesCollection = new ResponseCookies(headers);
+
+                    var serviceProvider = ServiceProvidersFeature.RequestServices;
+                    var provider = (ObjectPoolProvider)serviceProvider.GetService(typeof(ObjectPoolProvider)) ??
+                        new DefaultObjectPoolProvider();
+                    var pool = provider.CreateStringBuilderPool(
```
Tratcher: Does this actually create a new pool per request? Or is it a GetOrCreate?  How about if you call DefaultObjectPoolProvider?

If so, how is this an improvement over creating a few StringBuilders per request?

dougbu:
Closing. Will reopen against `dev`.

Issue #592, opened 2016-03-21
Move feature implementations into `Microsoft.AspNetCore.Http.Features` namespace


dougbu:
#589 exposes all of the feature interfaces `public`ly. But the feature implementations remain in `Microsoft.AspNetCore.Http.Features.Internal` and `Microsoft.AspNetCore.Http.Features.Authentication.Internal`. Since the feature implementations are used in Kestrel and supported, they should also be moved to `Microsoft.AspNetCore.Http.Features` and `Microsoft.AspNetCore.Http.Features.Authentication`.

See comment [here](https://github.com/aspnet/HttpAbstractions/pull/589#issuecomment-199109202) for example.

dougbu:
Self-assigning because I was playing with this cleanup already.

dougbu:
- 6725d68
- aspnet/Hosting@6e87b51
- aspnet/KestrelHttpServer@c828faf
- aspnet/WebListener@84c8975
- aspnet/IISIntegration@52e5ccd
- aspnet/CORS@5a3ea74
- aspnet/Routing@67f4243
- aspnet/StaticFiles@74d2c13
- aspnet/Antiforgery@65a2b8d
- aspnet/Diagnostics@f1a0243
- aspnet/Security@16a0482
- aspnet/Mvc@6bf25ce
- aspnet/Identity@b2bf1a4
- aspnet/Entropy@9da3ca9
- aspnet/MusicStore@20a24a2

Issue #549, opened 2016-01-30
Move HttpContextAccessor from .Internal namespace


davidfowl:
If users want to use this API they need to call

```C#
services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
```

Today this requires adding `using Microsoft.AspNet.Http.Internal;`

dougbu:
Suggest we take this further and move everything or almost everything our own repos depend on from `.Internal` into a public namespace. Right now Antiforgery, BasicMiddleware, CORS, Diagnostics, Entropy, Hosting, Identity, ... all use HttpAbstraction's `.Internal` namespaces. And all such namespaces defined in this repo except `Microsoft.AspNetCore.Http.Authentication.Internal` have references from other repos.

muratg:
@dougbu Good to close?

dougbu:
No. I'm doing more of this in concert w/ the #592 fix at @davidfowl's urging.

dougbu:
Taking this one because we're tying fix together w/ #592 i.e. will do one set of reaction PRs.

dougbu:
Will move the following types to `Microsoft.AspNetCore.Http`. All of these are currently used in other aspnet repo's product code.

- `DefaultHttpContext`
- `FormCollection` (though not mainstream, necessary for some scenarios e.g. handling a form-URL-encoded response as in aspnet/Security)
- `HeaderDictionary`
- `HttpContextAccessor` (remains the most important of these changes because it's necessary for any use of the `IHttpContextAccessor` feature e.g. in aspnet/Identity)
- `HttpContextFactory`

Will also move files around in the repo so the remaining `Microsoft.AspNetCore.Http.Internal` classes are in an `Internal` sub-folder.

dougbu:
- 6725d68
- aspnet/Hosting@6e87b51
- aspnet/KestrelHttpServer@c828faf
- aspnet/WebListener@84c8975
- aspnet/IISIntegration@52e5ccd
- aspnet/CORS@5a3ea74
- aspnet/Routing@67f4243
- aspnet/StaticFiles@74d2c13
- aspnet/Antiforgery@65a2b8d
- aspnet/Diagnostics@f1a0243
- aspnet/Security@16a0482
- aspnet/Mvc@6bf25ce
- aspnet/Identity@b2bf1a4
- aspnet/Entropy@9da3ca9
- aspnet/MusicStore@20a24a2

PR #594, opened 2016-03-21

dougbu:
- #561
- new `SetCookieHeaderValue.AppendToStringBuilder()` method; avoids per-call `StringBuilder` allocation
- `ResponseCookies` uses `ObjectPool<StringBuilder>` that `ResponseCookiesFeature` provides
- `ResponseCookiesFeature` creates an `ObjectPoolProvider` instance if none in DI
- `IResponseCookiesFeature` exposes `int` properties supporting middleware overrides of the pool policy

nit: Add some doc comments


src/Microsoft.AspNetCore.Http.Features/IResponseCookiesFeature.cs
```diff
@@ -3,8 +3,33 @@
...
+
+        /// <summary>
+        /// Maximum retained capacity of instances used in the implementation of this feature, in <see cref="char"/>s.
+        /// Instances larger than this will be deleted rather than preserved in the pool.
+        /// </summary>
+        /// <remarks>
+        /// For example, the maximum retained capacity of <see cref="System.Text.StringBuilder"/> instances obtained
+        /// from an object pool.
+        /// </remarks>
+        int MaximumRetainedPooledInstanceCapacity { get; set; }
```
dougbu: Copying @Tratcher's comment from previous PR: :-1: This is an implementation detail, not a generic feature property.

src/Microsoft.AspNetCore.Http/Features/ResponseCookiesFeature.cs
```diff
@@ -25,8 +49,17 @@ public IResponseCookies Cookies
...
                 {
                     var headers = HttpResponseFeature.Headers;
-                    _cookiesCollection = new ResponseCookies(headers);
+
+                    var serviceProvider = ServiceProvidersFeature.RequestServices;
+                    var provider = (ObjectPoolProvider)serviceProvider.GetService(typeof(ObjectPoolProvider)) ??
+                        new DefaultObjectPoolProvider();
+                    var pool = provider.CreateStringBuilderPool(
+                        InitialPooledInstanceCapacity,
+                        MaximumRetainedPooledInstanceCapacity);
```
dougbu: Copying @Tratcher's comment from previous PR: Does this actually create a new pool per request? Or is it a GetOrCreate? How about if you call DefaultObjectPoolProvider?

If so, how is this an improvement over creating a few StringBuilders per request?

src/Microsoft.AspNetCore.Http.Features/IResponseCookiesFeature.cs
dougbu: What do you suggest? This is at least abstracted from `StringBuilder` and `ObjectPool<T>`. I'm loath to expose an even-more-specific `ObjectPool<StringBuilder>` property anywhere.

One thought: We could expose these properties only on the feature implementation once I fix #592. Then the required downcast would at least be to a public class.

src/Microsoft.AspNetCore.Http/Features/ResponseCookiesFeature.cs
dougbu: Shoot.

dougbu:
:up::date:

samples/SampleApp/PooledHttpContextFactory.cs
```diff
@@ -1,24 +1,56 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
+using System.Text;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Http.Features.Internal;
```
dougbu: Won't need this addition once I'm done #592.

samples/SampleApp/PooledHttpContextFactory.cs
```diff
@@ -1,24 +1,56 @@
...
+            : this(poolProvider: null, httpContextAccessor: null)
+        {
+        }
+
+        public PooledHttpContextFactory(ObjectPoolProvider poolProvider)
+            : this(poolProvider, httpContextAccessor: null)
+        {
+        }
+
         public PooledHttpContextFactory(IHttpContextAccessor httpContextAccessor)
```
dougbu: Having this be the only constructor would mean the class could only be instantiated if an `IHttpContextAccessor` were registered. sad :panda_face:

src/Microsoft.AspNetCore.Http/HttpContextFactory.cs
```diff
@@ -1,30 +1,59 @@
...
+            : this(poolProvider, httpContextAccessor: null)
         {
         }
 
         public HttpContextFactory(IHttpContextAccessor httpContextAccessor)
+            : this(poolProvider: null, httpContextAccessor: httpContextAccessor)
         {
+        }
+
+        public HttpContextFactory(ObjectPoolProvider poolProvider, IHttpContextAccessor httpContextAccessor)
```
dougbu: Not testing constructor overloads w/ a non-`null` provider b/c that requires something ugly like passing a `Mock<IFeatureCollection>` to `Create()` (test project does not currently use `Moq`) or exposing an otherwise-useless property on the `ResponseCookiesFeature`.

src/Microsoft.AspNetCore.Http.Features/IResponseCookiesFeature.cs
```diff
@@ -3,8 +3,14 @@
 
 namespace Microsoft.AspNetCore.Http.Features
 {
+    /// <summary>
+    /// Feature containing cookies which will be returned with the response.
+    /// </summary>
     public interface IResponseCookiesFeature
     {
+        /// <summary>
+        /// Cookies which will be returned with the response.
```
Tratcher: Misleading, this is not a collection, it's helper methods for setting Headers.

dougbu: Yah want it described as an ease-of-use façade on top of `IHttpResponseFeature`?

Tratcher: Meh, "A helper for creating response cookie headers."

src/Microsoft.AspNetCore.Http/ResponseCookies.cs
```diff
@@ -71,7 +95,19 @@ public void Append(string key, string value, CookieOptions options)
                 HttpOnly = options.HttpOnly,
             };
 
-            Headers[HeaderNames.SetCookie] = StringValues.Concat(Headers[HeaderNames.SetCookie], setCookieHeaderValue.ToString());
+            string cookieValue;
+            var stringBuilder = _builderPool.Get();
```
Tratcher: Check if the pool exists?

test/Microsoft.AspNetCore.Http.Tests/ResponseCookiesTest.cs
```diff
@@ -1,19 +1,24 @@
...
     {
+        private static readonly ObjectPool<StringBuilder> _builderPool =
+            new DefaultObjectPoolProvider().Create<StringBuilder>(new StringBuilderPooledObjectPolicy());
+
         [Fact]
         public void DeleteCookieShouldSetDefaultPath()
         {
             var headers = new HeaderDictionary();
-            var cookies = new ResponseCookies(headers);
+            var cookies = new ResponseCookies(headers, _builderPool);
```
Tratcher: Some tests should use the pool and others shouldn't so you test both code paths. Or you make them all test both scenarios.

Tratcher:
:+1: progress

src/Microsoft.AspNetCore.Http/ResponseCookies.cs
dougbu: Yah, I only fixed one overload.

test/Microsoft.AspNetCore.Http.Tests/ResponseCookiesTest.cs
dougbu: :ok:

davidfowl:
Much better, I'm not 100% comfortable with this yet. Trying to see how it scales when we add things.

dougbu:
> Trying to see how it scales when we add things

1. Main "thing" not visible in this PR is the code in our cookie-related middleware setup to (try)add an `ObjectPoolProvider` to DI. Worst case is something adding cookies doesn't perform as well as it could.
2. If later added "things" are additional `ObjectPool<StringBuilder>` uses, likely the existing one in `HttpContextFactory` will suffice -- the sizes are fine for this and the form parsing use cases. Extra code would instantiate and set additional features with an `ObjectPool<StringBuilder>` constructor parameter.
3. If later added "things" are other pools (or the existing one isn't configured correctly), `HttpContextFactory` could create more `ObjectPool<T>` instances. Would of course also need the extra code from (1) and might need (3) as well.
4. If later added "things" are `IPooledObjectPolicy<T>` instances or `int` values for particular pools, would add `IOptions<HttpContextFactorySettings>` parameters to the `HttpContextFactory` constructor overloads accepting an `ObjectPoolProvider`. That's a bit ugly but it uses our bog-standard options pattern. We'd end up with at most 6 `ObjectPoolProvider` constructor overloads but the existing 3 seems much more likely, at least until we need to pool something besides `StringBuilder`s.

dougbu:
:up::date: addressed @Tratcher's comments

src/Microsoft.AspNetCore.Http.Features/project.json
```diff
@@ -27,7 +27,7 @@
         "System.Net.WebSockets": "4.0.0-*",
         "System.Runtime.Extensions": "4.1.0-*",
         "System.Security.Claims": "4.0.1-*",
-        "System.Security.Cryptography.X509Certificates": "4.1.0-*",
+        "System.Security.Cryptography.X509Certificates": "4.0.0-*",
```
dougbu: Not sure this will be necessary next time CI updates the feed. @pranavkm?

dougbu:
:up::date: (rebased)

ping @Tratcher @davidfowl

Tratcher:
:shipit:

dougbu:
:up::date: (as requested, @davidfowl)

src/Microsoft.AspNetCore.Http.Features/project.json
pranavkm: Yup, the 4.1.0 package is available in the new corefx build.

src/Microsoft.AspNetCore.Http/HttpContextFactory.cs
```diff
@@ -1,30 +1,51 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Text;
 using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Http.Features.Internal;
+using Microsoft.Extensions.ObjectPool;
 
 namespace Microsoft.AspNetCore.Http.Internal
```
muratg: Wasn't the plan to move the Features out of Internal namespace?

dougbu: Yup. I have a second wave of breaking changes (after moving the feature implementations) queued up. That'll cover both #549 and #592.

dougbu:
80813f7

PR #595, opened 2016-03-22
Use System.Buffers for temporary arrays

benaadams:
Buffers only from #556 

/cc @Tratcher


src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
```diff
@@ -135,22 +180,27 @@ private Stream CreateTempFile()
         public override int Read(byte[] buffer, int offset, int count)
         {
             ThrowIfDisposed();
-            if (_buffer.Position < _buffer.Length || _completelyBuffered)
+            if (_buffer.Position < _readIntoBuffer || _completelyBuffered)
```
Tratcher: Did `SetLength(0)` not work for you?

src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
```diff
@@ -613,9 +615,17 @@ private bool TryDecode5987(string input, out string output)
             }
             catch (ArgumentException)
             {
+                if (unescapedBytes != null)
+                {
+                    ArrayPool<byte>.Shared.Return(unescapedBytes);
+                }
                 return false; // Unknown encoding or bad characters
             }
```
Tratcher: finally?

src/Microsoft.AspNetCore.WebUtilities/StreamHelperExtensions.cs
```diff
@@ -9,14 +10,28 @@ namespace Microsoft.AspNetCore.WebUtilities
 {
     public static class StreamHelperExtensions
     {
-        public static async Task DrainAsync(this Stream stream, CancellationToken cancellationToken)
+        private const int _maxReadBufferSize = 1024;
```
Tratcher: This seems a bit small compared with the other read buffers we use. 4kb? 16kb?

src/Microsoft.AspNetCore.WebUtilities/MultipartReaderStream.cs
```diff
@@ -212,22 +227,25 @@ public override int Read(byte[] buffer, int offset, int count)
...
+                read = _innerStream.Read(boundary, 0, length);
+                Debug.Assert(read == length); // It should have all been buffered
+
                 var remainder = _innerStream.ReadLine(lengthLimit: 100); // Whitespace may exceed the buffer.
                 remainder = remainder.Trim();
                 if (string.Equals("--", remainder, StringComparison.Ordinal))
                 {
                     FinalBoundaryFound = true;
                 }
+                _bytePool.Return(boundary);
```
Tratcher: This boundary buffer can be returned on line 239 right after the read, correct?

Tratcher:
Cleaned up, rebased, and merged. Thanks!

Issue #605, opened 2016-03-31
EnableRewind causes data loss on files larger than the buffer threshold


DavidObando:
Multipart posts containing files larger than the [default buffering threshold](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http/Internal/BufferingHelper.cs#L12) (currently 30 KB) used by [EnableRewind](https://github.com/aspnet/HttpAbstractions/blob/6725d68559a55713fdd3d2e38de2ab22059219da/src/Microsoft.AspNetCore.Http/Features/FormFeature.cs#L148) will result in data loss.

The data loss I've seen is usually at the beginning of the file, and usually around 28 KB long. For example, uploading a file that on the client side is 100 KB, the server will save ~72.3 KB on a file. As a proof of concept, when I enlarged the buffering threshold to 200 KB the server was able to save the entire 100 KB file. The solution of course isn't to enlarge the buffer threshold but to make it so we don't lose the bits from the first buffer frame, as we apparently are.

A self-contained repro can be found at: https://github.com/DavidObando/BugRepros/tree/master/2016-03-31/EnableRewindBug. The repro will save the client-side file and the server-side file in C:\Temp\ for easy comparison.

Tratcher:
Hmm, there was a perf change in this code last week, I wonder if it regressed something...
https://github.com/aspnet/HttpAbstractions/commit/bd60507dcd317f2c1f42da37afd9dbc928f50928#diff-6bee66e850efacd8fe96daae334fa1c5R195

davidfowl:
Revert it and test.

Tratcher:
Yes, that was it. I'll send a PR after I add some tests.

PR #606, opened 2016-04-01
Fix regressions in FormReader / FileBufferingReadStream.

Tratcher:
#605 @davidfowl @muratg @benaadams @BrennanConroy 

The recent perf changes caused two regressions in the FileBufferingReadStream used by form reader. The one reported in #605 is that most of the data buffered in memory gets discarded when transitioning to the file buffer. This happened because Write was accidentally given the wrong length when I cleaned up some of @benaadams's code.

The other regression was that the constructor only rented buffers if the in-memory threshold was less than 1mb, but Read always assumed it was using rented buffers. This would have caused ArgumentNullExceptions trying to use an uninitialized buffer. The fix is to rent a buffer when needed and do a manual CopyTo loop.

Compare with https://github.com/aspnet/HttpAbstractions/commit/bd60507dcd317f2c1f42da37afd9dbc928f50928#diff-6bee66e850efacd8fe96daae334fa1c5R192

### DIFF
--- a/src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
+++ b/src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
@@ -192,10 +192,26 @@ namespace Microsoft.AspNetCore.WebUtilities
             if (_inMemory && _buffer.Length + read > _memoryThreshold)
             {
                 _inMemory = false;
+                var oldBuffer = _buffer;
                 _buffer = CreateTempFile();
-                _buffer.Write(_rentedBuffer, 0, (int)_buffer.Length);
-                _bytePool.Return(_rentedBuffer);
-                _rentedBuffer = null;
+                if (_rentedBuffer == null)
+                {
+                    oldBuffer.Position = 0;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
```diff
@@ -272,10 +288,27 @@ public override int EndRead(IAsyncResult asyncResult)
...
             {
                 _inMemory = false;
+                var oldBuffer = _buffer;
                 _buffer = CreateTempFile();
-                await _buffer.WriteAsync(_rentedBuffer, 0, (int)_buffer.Length, cancellationToken);
-                _bytePool.Return(_rentedBuffer);
-                _rentedBuffer = null;
+                if (_rentedBuffer == null)
+                {
+                    oldBuffer.Position = 0;
```
rynowak: It breaks my brain a little bit to look at this code and see 'buffer' used for both `Stream` and arrays

BrennanConroy:
Looks reasonable :shipit:

Issue #572, opened 2016-02-24
Update AuthenticationManager.AuthenticateAsync to return an AuthenticationTicket


PinpointTownes:
In June, the `AuthenticateAsync` method was updated to return a `ClaimsPrincipal` instead of an `AuthenticationResult` (a type similar to `AuthenticationTicket`). As a consequence, retrieving the authentication properties is much harder, and can only be done using the rather unusual overload taking an `AuthenticateContext` parameter:

https://github.com/aspnet/HttpAbstractions/pull/323#discussion_r31666811

@HaoK's main argument was that retrieving authentication properties was not really frequent. Sadly, I don't think it's true, as retrieving the expiration date from the ticket is quite popular: http://stackoverflow.com/a/34535003/542757

If you really think having a method returning a `ClaimsPrincipal` is important, why not adding an extension method?

```csharp
public static async Task<ClaimsPrincipal> GetPrincipalAsync(this AuthenticationManager manager, string scheme) {
    if (manager == null) {
        throw new ArgumentNullException(nameof(manager));
    }

    return (await manager.AuthenticateAsync(scheme))?.Principal;
}
```

brockallen:
+1

HaoK:
I'm not opposed to bringing back AuthenticateResult I guess, perhaps as an AuthenticatePrincipal overload instead of a Get...

```
public abstract class AuthenticationManager {
  public AuthenticationResult AuthenticateAsync(string authenticationScheme);
  public ClaimsPrincipal AuthenticatePrincipalAsync(string authenticationScheme);
```

muratg:
@Eilon

PinpointTownes:
Is it too late for RC2? I think it would be worth fixing that before RTM.

Eilon:
I don't really know what this stuff is :smile: 

If @HaoK and/or @Tratcher think this is useful, sure, we can look into adding it.

Eilon:
And @blowdart for that matter.

blowdart:
I'm fine with it

Tratcher:
@HaoK can you take this? And give us an estimate on churn? I'm concerned that this will be too much churn for RC2.

HaoK:
I can take it, its not too bad if we do something like:

Today's Authenticate => AuthenticateUser
New Autenticate returns AuthenticateResult again

Its a breaking change but not that widespread outside of identity and maybe Authorize filter in MVC

HaoK:
[...]

PR #616, opened 2016-04-18
Switch to GetAuthenticationInfoAsync

HaoK:
Fixes https://github.com/aspnet/HttpAbstractions/issues/572

@Tratcher @Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Authentication/AuthenticateInfo.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Authentication/AuthenticateInfo.cs
@@ -1,0 +1,24 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Security.Claims;
+
+namespace Microsoft.AspNetCore.Http.Authentication
+{
+    /// <summary>
+    /// Used to store the results of an Authenticate call.
+    /// </summary>
+    public class AuthenticateInfo
+    {
+        /// <summary>
+        /// The <see cref="ClaimsPrincipal"/>.
+        /// </summary>
+        public ClaimsPrincipal Principal { get; set; }
+
+        /// <summary>
+        /// The <see cref="AuthenticationProperties"/>.
+        /// </summary>
+        public AuthenticationProperties Properties { get; set; }
+    }
+}
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Authentication/AuthenticationManager.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Authentication/AuthenticationManager.cs
@@ -22,6 +22,9 @@ namespace Microsoft.AspNetCore.Http.Authentication
 
         public abstract Task<AuthenticateInfo> GetAuthenticateInfoAsync(string authenticationScheme);
 
+        // Will remove once callees have been updated
+        public abstract Task AuthenticateAsync(AuthenticateContext context);
+
         public virtual async Task<ClaimsPrincipal> AuthenticateAsync(string authenticationScheme)
         {
             return (await GetAuthenticateInfoAsync(authenticationScheme))?.Principal;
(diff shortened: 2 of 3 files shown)


HaoK:
Note this is a breaking change and all uses of `AuthenticateAsync(AuthenticateContext)` will need to be updated (should mostly be only MVC/Security)

Tratcher:
Not rc2?

src/Microsoft.AspNetCore.Http.Abstractions/Authentication/AuthenticateInfo.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// Used to store the results of an Authenticate call.
+    /// </summary>
+    public class AuthenticateInfo
+    {
+        /// <summary>
+        /// The <see cref="ClaimsPrincipal"/>.
+        /// </summary>
+        public ClaimsPrincipal Principal { get; set; }
```
PinpointTownes: Curious: is there any particular reason to use publicly settable properties instead of get-only properties + public constructor?

src/Microsoft.AspNetCore.Http.Abstractions/Authentication/AuthenticationManager.cs
```diff
@@ -20,18 +20,11 @@ public abstract class AuthenticationManager
 
         public abstract IEnumerable<AuthenticationDescription> GetAuthenticationSchemes();
 
-        public abstract Task AuthenticateAsync(AuthenticateContext context);
+        public abstract Task<AuthenticateInfo> GetAuthenticateInfoAsync(string authenticationScheme);
```
PinpointTownes: FWIW, since this is the main "API", this one should probably be named `AuthenticateAsync` (as indicated by `AuthenticateInfo`'s XML documentation, BTW). The other one is just a wrapper, so it may make more sense to switch the two names.

Tratcher: Switching the names would be a major break that we want to avoid. Not nearly as many people will use this version.

PinpointTownes: Well, TBH, how many people are using passive authentication these days? 5%? 10%?
These APIs are rather complicated for beginners and most will prefer using active authentication (which is the default for both cookies and bearer) or use Identity. In this case, retrieving the principal via `HttpContext.User` or `ControllerBase.User` will be much easier.

src/Microsoft.AspNetCore.Http.Abstractions/Authentication/AuthenticateInfo.cs
HaoK: No particular reason, if you guys feel strongly this should be readonly ctor properties, easy enough change.

src/Microsoft.AspNetCore.Http.Abstractions/Authentication/AuthenticationManager.cs
HaoK: Yeah so of that 5/10%, I'd say only another 5/10% of those passive calls will want the AuthenticationProperties, most will only want the Principal...

PinpointTownes: That's why I think a simpler name like `GetUserAsync` or `GetPrincipalAsync` would be better for the "get just the principal" use case (people often mix up `AuthenticateAsync` and `SignInAsync`).

Anyway, as long as the odd `AuthenticateAsync` overload is removed, I'm fine :smile:

PinpointTownes: (oh, and `GetUserAsync` would be - almost - consistent with `HttpContext.User` and `ControllerBase.User` :smile:)

HaoK:
@Tratcher any final concerns with this?

Tratcher:
:shipit:

HaoK:
3a7f6a7228fe10b14996b97eed83bcb72684088a

PR #618, opened 2016-04-22
Make the form body limits configurable.

Tratcher:
@DamianEdwards @Eilon Please approve for RC2 per @blowdart's request.

@blowdart This enables the developer to configure form parsing settings and limits in Startup:
```c#
services.Configure<FormOptions>(options =>
{
  options.BufferBody = true,
});
```
And override them on a per request basis by doing:
```c#
var form = await httpContext.Request.ReadAsync(new FormOptions() { BufferBody = false });
```
The catch is that you can't override the settings if someone before you in the pipeline has already read the form.

There will be follow up discussion with MVC on how to easily configure this per Action, how to make anti-forgery honor it, and how to tell MVC not to read the form at all so you can stream large file uploads without buffering.

### DIFF
--- a/src/Microsoft.AspNetCore.Http/Features/FormOptions.cs
+++ b/src/Microsoft.AspNetCore.Http/Features/FormOptions.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.WebUtilities;
+
+namespace Microsoft.AspNetCore.Http.Features
+{
+    public class FormOptions
+    {
+        public const int DefaultMemoryBufferThreshold = 1024 * 64;
+        public const int DefaultBufferBodyLengthLimit = 1024 * 1024 * 128;
+        public const int DefaultMultipartBoundaryLengthLimit = 128;
+        public const long DefaultMultipartBodyLengthLimit = 1024 * 1024 * 128;
+
+        public bool BufferBody { get; set; } = false;
+        public int MemoryBufferThreshold { get; set; } = DefaultMemoryBufferThreshold;
+        public long BufferBodyLengthLimit { get; set; } = DefaultBufferBodyLengthLimit;
+        public int KeyCountLimit { get; set; } = FormReader.DefaultKeyCountLimit;
+        public int KeyLengthLimit { get; set; } = FormReader.DefaultKeyLengthLimit;
+        public int ValueLengthLimit { get; set; } = FormReader.DefaultValueLengthLimit;
+        public int MultipartBoundaryLengthLimit { get; set; } = DefaultMultipartBoundaryLengthLimit;
+        public int MultipartHeadersCountLimit { get; set; } = MultipartReader.DefaultHeadersCountLimit;
+        public int MultipartHeadersLengthLimit { get; set; } = MultipartReader.DefaultHeadersLengthLimit;
+        public long MultipartBodyLengthLimit { get; set; } = DefaultMultipartBodyLengthLimit;
+    }
+}
--- a/src/Microsoft.AspNetCore.WebUtilities/BufferedReadStream.cs
+++ b/src/Microsoft.AspNetCore.WebUtilities/BufferedReadStream.cs
@@ -352,7 +352,7 @@ namespace Microsoft.AspNetCore.WebUtilities
                 {
                     if (builder.Length > lengthLimit)
                     {
-                        throw new InvalidOperationException("Line length limit exceeded: " + lengthLimit.ToString());
+                        throw new InvalidDataException($"Line length limit {lengthLimit} exceeded.");
                     }
                     ProcessLineChar(builder, ref foundCR, ref foundCRLF);
                 }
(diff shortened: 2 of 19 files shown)


src/Microsoft.AspNetCore.Http/Features/FormOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+        public int MemoryBufferThreshold { get; set; } = 1024 * 64;
+        public long BufferBodyLengthLimit { get; set; } = 1024 * 1024 * 128;
+        public int KeyCountLimit { get; set; } = FormReader.DefaultKeyCountLimit;
+        public int KeyLengthLimit { get; set; } = FormReader.DefaultKeyLengthLimit;
+        public int ValueLengthLimit { get; set; } = FormReader.DefaultValueLengthLimit;
+        public int MultipartBoundaryLengthLimit { get; set; } = 128;
+        public int MultipartHeadersCountLimit { get; set; } = MultipartReader.DefaultHeadersCountLimit;
+        public int MultipartHeadersLengthLimit { get; set; } = MultipartReader.DefaultHeadersLengthLimit;
+        public long MultipartBodyLengthLimit { get; set; } = 1024 * 1024 * 128;
+    }
```
muratg: Why are some of these defaults are just magic numbers, and other are consts?

Tratcher: The constants are the defaults available on lower level components. The options with magic numbers are implemented directly at the FormFeature level. Would you prefer FormFeatureConstants.DefaultMultipartBodyLengthLimit ? Or FormFeature.DefaultMultipartBodyLengthLimit?

Tratcher: Or FormOptions.DefaultMultipartBodyLengthLimit?

muratg: I see, I guess it's a style question then. Follow whatever we're doing elsewhere.

Tratcher:
@rynowak

src/Microsoft.AspNetCore.WebUtilities/BufferedReadStream.cs
```diff
@@ -352,7 +352,7 @@ public string ReadLine(int lengthLimit)
                 {
                     if (builder.Length > lengthLimit)
                     {
-                        throw new InvalidOperationException("Line length limit exceeded: " + lengthLimit.ToString());
+                        throw new InvalidDataException($"Line length limit {lengthLimit} exceeded.");
```
muratg: :+1: InvalidDataException

muratg:
Looks good to me. Unless you want a second pair of eyes, :shipit:

src/Microsoft.AspNetCore.WebUtilities/MultipartReader.cs
```diff
@@ -79,23 +87,30 @@ public MultipartReader(string boundary, Stream stream, int bufferSize)
...
-                    throw new InvalidOperationException("Total header size limit exceeded: " + TotalHeaderSizeLimit.ToString());
+                    throw new InvalidDataException($"Multipart headers length limit {HeadersLengthLimit} exceeded.");
                 }
+                totalSize += line.Length;
                 int splitIndex = line.IndexOf(':');
-                Debug.Assert(splitIndex > 0, $"Invalid header line: {line}");
+                if (splitIndex <= 0)
+                {
+                    throw new InvalidDataException($"Invalid header line: {line}");
+                }
```
moozzyk: nit: the condition in the `if` below is now always true.

src/Microsoft.AspNetCore.Http.Extensions/RequestFormReaderExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+
+            if (!request.HasFormContentType)
+            {
+                throw new InvalidOperationException("Incorrect Content-Type: " + request.ContentType);
+            }
+
+            var features = request.HttpContext.Features;
+            var formFeature = features.Get<IFormFeature>();
+            if (formFeature == null || formFeature.Form == null)
```
davidfowl: It sucks that you have no idea if the options passed in will apply. This feels like writing to the response after the http headers have been sent.

Tratcher: Agreed, but think through the scenarios:
A) I want to raise a limit but the form has already been read. The form must have already fit within the lower limit so it would also fit into the higher limit. A no-op here is harmless.
B) I want to lower a limit but the form has already been read. The memory has already been allocated, there's not much I can do about it except terminate the request. The alternative here is to verify the limit after the fact and throw.

I expect A to be far more common than B. Devs should set limits low by default and raise them on an as-needed basis.

davidfowl: I'll wait for the stackoverflow questions before I cast my judgement...

src/Microsoft.AspNetCore.Http.Extensions/RequestFormReaderExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http.Features;
+
+namespace Microsoft.AspNetCore.Http
+{
+    public static class RequestFormReaderExtensions
```
davidfowl: Why is this in extensions?

Issue #547, opened 2016-01-30
Continuation of encoding in Cookie discussion


psibernetic:
Refers to #535 & PR #546 

Also open on Katana: http://katanaproject.codeplex.com/workitem/442

@Tratcher 

Forcefully encoding and forcefully decoding breaks interoperability with existing solutions other than Microsoft.Owin. Furthermore cookie handling is now non-symmetrical in operations. Both the write and read use Uri.EscapeDataString, but the read operation does extra work which invalidates Base64. 

```csharp
public static RequestCookieCollection Parse(IList<string> values)
        {
            if (values.Count == 0)
            {
                return Empty;
            }

            IList<CookieHeaderValue> cookies;
            if (CookieHeaderValue.TryParseList(values, out cookies))
            {
                if (cookies.Count == 0)
                {
                    return Empty;
                }

                var store = new Dictionary<string, string>(cookies.Count);
                for (var i = 0; i < cookies.Count; i++)
                {
                    var cookie = cookies[i];
                    var name = Uri.UnescapeDataString(cookie.Name.Replace('+', ' '));
                    var value = Uri.UnescapeDataString(cookie.Value.Replace('+', ' '));
                    store[name] = value;
                }

                return new RequestCookieCollection(store);
            }
            return Empty;
        }
```

psibernetic:
#515 is related too it seems as is #390

psibernetic:
An example of Base64 cookie interop with ASPNET without OWIN and with as was needed in Microsoft.OWIN:
```csharp
            app.UseCookieAuthentication(
                new CookieAuthenticationOptions
                {
                    AuthenticationType = WsFederationAuthenticationDefaults.AuthenticationType,
                    CookieDomain = ".domain.com",
                    CookieName = "PreExistingCookieName",
                    TicketDataFormat = new WifTokenFormat(), //Replaces ' ' with '+' cause yeah.
                    SlidingExpiration = true,
                    Provider = new CookieAuthenticationProvider
                    {
                        OnResponseSignedIn = context =>
                        {
                            context.Response.Headers["Set-Cookie"] = Uri.UnescapeDataString(context.Response.Headers["Set-Cookie"]
                                .Replace(' ', '+'));
                        }
                    }
                });
```

psibernetic:
Would it be possible with the new DI services in ASPNET to inject a CookieEncodingService instead of hard-coding an impl, I think that would satisfy both camps.

osudude:
[...]

Issue #535, opened 2016-01-14
Comma's in cookie value


martijnboland:
I've had an issue where I wanted to store a serialized json array in a cookie. This is not allowed anymore because of the comma that appears in the serialized representation (e.g. "[\"value 1\",\"value 2\"]"). 
Comma's are indeed not supported in cookies, but in the past (System.Web, Owin), these were always encoded and this does not seem to be the case anymore.

I wonder, should we encode and decode these values ourselves or should the framework do that automatically.

Tratcher:
This is for `httpContext.Response.Cookies.Append(key, value);` correct?

martijnboland:
Yes

muratg:
Moving this to Backlog as we will be in RC2 ask mode very soon.  If you feel strongly about this issue, please ping me.

brockallen:
Yes, this seems like an important issue to understand why it's happening.

muratg:
Yep, looks like a bug. Though we are crammed on time and we will revisit backlog bugs after RC2.

CrispinH:
This bug is affecting my use of IdentityServer4 and presumably everyone else's too. Given Dominick Baier's blog post paragraph: "One year ago the ASP.NET team decided to discontinue that middleware and rather focus on consuming tokens instead. They also asked us if IdentityServer can be the replacement going forward." - this is needs fixing.

Tratcher:
@leastprivilege, can you elaborate on how this affects IdentiyServer?

brockallen:
We put the client id the user has logged into into a cookie. This is used for single sign out.

Tratcher:
In a json format? or does the client id happen to contain a comma?

brockallen:
We serailize to json and it's an array of strings. It's the exact same code in IdentityServer3 and that works under katana :)

Basically, the APIs to add the cookie seem to automatically URL encode the value. In ASP.NET Core it's odd that the value is encoded, but there's a "," left in the middle.

brockallen:
Here's a link with the sample text that's being put intot he cookie: https://github.com/IdentityServer/IdentityServer3/issues/2446#issuecomment-172908822

Tratcher:
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http/ResponseCookies.cs#L65

brockallen:
Right, so why is the default url encoder putting a comma in there? I can try to test it in better isolation.

[...]

PR #622, opened 2016-05-02
Remove '+' replacement from request cookies.

Tratcher:
#547 @muratg @BrennanConroy 
When un-escaping cookies there is an extra step that replaces `+` with ` ` (space). This is leftover code from Katana when the parser was shared with queries. This messes up reading base64 encoded values, remove it.

### DIFF
--- a/test/Microsoft.AspNetCore.Http.Tests/RequestCookiesCollectionTests.cs
+++ b/test/Microsoft.AspNetCore.Http.Tests/RequestCookiesCollectionTests.cs
@@ -1,0 +1,46 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using Microsoft.AspNetCore.Http.Internal;
+using Microsoft.Extensions.Primitives;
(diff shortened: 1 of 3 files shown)


muratg:
:shipit:

Could you also check the parser for the forms?

test/Microsoft.AspNetCore.Http.Tests/RequestCookiesCollectionTests.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Microsoft.AspNetCore.Http.Tests
+{
+    public class RequestCookiesCollectionTests
+    {
+        public static TheoryData UnEscapesKeyValues_Data
+        {
+            get
+            {
+                // key, value, object pool, expected
```
khellang: `object pool`?

Tratcher: Woops, copied from the response cookies test... Removed.

brockallen:
Hopefully this issue won't resurface? https://github.com/aspnet/HttpAbstractions/issues/535

davidfowl:
We should have tests for anything previously fixed.

Issue #617, opened 2016-04-19
Should Fetch be made more inline-friendly?


rynowak:
It's common in MVC to see traces like:
![image](https://cloud.githubusercontent.com/assets/1430011/14624142/d65f7648-058c-11e6-9935-77ca8e34ed3e.png)

There's about 1.33% of a request's time spent just doing the initial feature access. 

We should probably see what the JIT has to say about this code path and try to either optimize for:
- first read (`cached == null`) **OR**
- multiple reads (`cached != null && Revision == collection.Revision`)

by breaking this up into an inlinable fast path and non-inlinable slow path

https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Features/FeatureReferences.cs#L26

benaadams:
trickier as cached is often a struct

benaadams:
@rynowak is latest commit on #623 now improved for this?

rynowak:
@benaadams - I'll give this a try when I can - that will probably be next week unf. The profiler should be able to tell whether it's getting inlined or not.

benaadams:
#623 was nothing, #704 is the mother lode...

PR #623, opened 2016-05-03
Reduce Revision check calls via interface

benaadams:
Can cache the first lookup

Revision calls

Revision calls 2

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Features/FeatureReferences.cs
+++ b/src/Microsoft.AspNetCore.Http.Features/FeatureReferences.cs
@@ -26,34 +26,39 @@ namespace Microsoft.AspNetCore.Http.Features
         public TFeature Fetch<TFeature, TState>(
             ref TFeature cached,
             TState state,
-            Func<TState, TFeature> factory)
+            Func<TState, TFeature> factory) where TFeature : class
         {
-            var cleared = false;
-            if (Revision != Collection.Revision)
+            var revision = Collection.Revision;
+            if (Revision == revision)
             {
-                cleared = true;
-                Cache = default(TCache);
+                // collection unchanged, use cached
+                return cached ?? UpdateCached(ref cached, state, factory);
+            }
+
+            // collection changed, clear cache
+            Cache = default(TCache);
+            // empty cache is current revision
+            Revision = revision;
+
+            return UpdateCached(ref cached, state, factory);
+        }
+
+        private TFeature UpdateCached<TFeature, TState>(ref TFeature cached, TState state, Func<TState, TFeature> factory) where TFeature : class
+        {
+            cached = Collection.Get<TFeature>();
+            if (cached == null)
+            {
+                // create if item not in collection
+                cached = factory(state);
+                Collection.Set(cached);
+                // Revision changed by .Set, update revision
                 Revision = Collection.Revision;
             }
 
-            var feature = cached;
-            if (feature == null || cleared)
-            {
-                feature = Collection.Get<TFeature>();
-                if (feature == null)
-                {
-                    feature = factory(state);
-
-                    Collection.Set(feature);
-
-                    Revision = Collection.Revision;
-                }
-                cached = feature;
-            }
-            return feature;
+            return cached;
         }
 
-        public TFeature Fetch<TFeature>(ref TFeature cached, Func<IFeatureCollection, TFeature> factory) =>
-            Fetch(ref cached, Collection, factory);
+        public TFeature Fetch<TFeature>(ref TFeature cached, Func<IFeatureCollection, TFeature> factory)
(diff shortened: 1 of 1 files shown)


rynowak:
:heart_eyes_cat:

rynowak:
Not to derail too much, but do we have a feel for how much benefit we're really getting from using this pattern instead of just calling `Get` all the time?

benaadams:
> a feel for how much benefit we're really getting from using this pattern instead of just calling Get all the time?

If you only access a feature once per request not much; but if you access it more than once it should be significant e.g. the plaintext benchmark accesses `httpContext.Response` 4 times

benaadams:
> a feel for how much benefit we're really getting from using this pattern instead of just calling Get all the time?

If you only access a feature once per request not much; but if you access it more than once it should be significant e.g. the plaintext benchmark accesses `httpContext.Response` 4 times

However instead of doing
```csharp
if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal))
{
    httpContext.Response.StatusCode = 200;
    httpContext.Response.ContentType = "text/plain";
    httpContext.Response.Headers["Content-Length"] = "13";
    return httpContext.Response.Body.WriteAsync(_helloWorldPayload, 0, _helloWorldPayload.Length);
}
```
It could do
```csharp
if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal))
{
    var response = httpContext.Response;
    response.StatusCode = 200;
    response.ContentType = "text/plain";
    response.Headers["Content-Length"] = "13";
    return response.Body.WriteAsync(_helloWorldPayload, 0, _helloWorldPayload.Length);
}
```
To access it only once; but users probably won't do that

benaadams:
@rynowak https://github.com/aspnet/benchmarks/pull/84 :)

benaadams:
My mistake, it still goes via the feature anyway even if the `Reponse` is cached; however the first access on StatusCode is very high

feature

With the `Get` on the frame being 41% of that + casting etc

get

Then the lookups for the other features are very low

rynowak:
Maybe we could make the cost of the extra lookups even lower? :grinning: https://github.com/aspnet/HttpAbstractions/issues/617

The hope is that the cost of abstraction is low and that doing optimizations like using the feature directly aren't super beneficial.

benaadams:
I was trying to see if it could be a compile/jit time lookup though it would only work with generics not `typeof` https://github.com/aspnet/HttpAbstractions/compare/dev...benaadams:strong-typed-features

src/Microsoft.AspNetCore.Http.Features/FeatureReferences.cs
```diff
@@ -26,34 +26,53 @@ public FeatureReferences(IFeatureCollection collection)
         public TFeature Fetch<TFeature, TState>(
             ref TFeature cached,
             TState state,
-            Func<TState, TFeature> factory)
+            Func<TState, TFeature> factory) where TFeature : class
         {
-            var cleared = false;
-            if (Revision != Collection.Revision)
+            if (Revision == 0)
```
rynowak: We could also check `cached == null` here

benaadams: If `cached == null` it couldn't update the revision count as it may have changed and other cached features may be invalid; which would mean on second use it would have to clear the cache.

Tratcher: This makes some assumptions about the implementation of Collection.Revision and the initial value. E.g. WebListener's Revision count starts at 0 so this code will keep executing.

benaadams: Resolved, but 3 nested ifs :(

benaadams: Hmm, but this will go wrong if revision has changed

benaadams: k, now is fixed

davidfowl: This code could do with some comments. If we had fxcop it would probably complain about the cyclomatic complexity

Tratcher: You'd be surprised how high that complicity limit is.

benaadams: Better?

benaadams:
OSX is grumpy on restore

benaadams:
Looks to be faster for the same number of requests

Pre
pre-feature-changes

Post
post-feature-changes

Tratcher:
:shipit:

Tratcher:
Rebased and merged.

Issue #612, opened 2016-04-11
Move CookieSecureOption / SecurePolicy to Http.Abstractions


Tratcher:
This enum controls the behavior of the cookie secure option. We're going to add the same option to Session, so consolidate the enum down to Http.Abstractions.
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.CookiePolicy/SecurePolicy.cs
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.Cookies/CookieSecureOption.cs

PR #631, opened 2016-05-13
Move CookieSecureOption / SecurePolicy to Http.Abstractions

Tratcher:
#612 @blowdart @HaoK @muratg 
Moving this enum to HttpAbstractions so we can de-duplicate the two existing copies, as well as add new uses from Session, etc..


muratg:
:shipit:

Tratcher:
Small update renaming Never to None.

Issue #634, opened 2016-05-19
Add IsAvailable to ISession


JunTaoLuo:
As part of https://github.com/aspnet/Session/pull/113. We realized there is value in adding a `IsAvailable` property to the `ISession` interface to indicate whether the session has loaded correctly. This value can depend on factors such as whether the underlying caching data is available and is valid.

kichalla:
cc @ajaybhargavb 

@JunTaoLuo @Tratcher  is the expectation that where ever session is required to be present, the code should check for IsAvailable explicitly and throw? (for example: MVC's TempData depends on it...can you file an issue on Mvc to update it https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/SessionStateTempDataProvider.cs#L61)

JunTaoLuo:
@kichalla The idea is that if the Session is not available, you should not depend on it to give you any information. You should have some fall back logic to handle it and depending on the case, throwing and failing the request may or may not be the right thing to do. 

Note that if the session is not available, it will no op when you try to access or modify it. `TryGetValue()` will return `false` and `Set`/`Remove` will no-op.

Tratcher:
It's up to the caller to decide what to do. IsAvailable is just a hint to the app so it can fail more gracefully if expected values are not present.

kichalla:
Ok sure, thanks.

PR #635, opened 2016-05-19
Add IsAvailable property to ISession #634

JunTaoLuo:
#634 

cc @Tratcher


davidfowl:
:shipit:

src/Microsoft.AspNetCore.Http.Features/ISession.cs
```diff
@@ -9,6 +9,11 @@ namespace Microsoft.AspNetCore.Http
     public interface ISession
     {
         /// <summary>
+        /// Indicate whether the current session has loaded correctly.
```
Eilon: `has been loaded.` (remove "correctly")

Issue #596, opened 2016-03-24
Add raw request URI property to IHttpRequestFeature


cesarbs:
`Path` and `PathBase` might lose some information. Expose the raw, unaltered request URI via the new property.

Related: https://github.com/aspnet/KestrelHttpServer/issues/594

This will also mitigate the loss of information in https://github.com/aspnet/KestrelHttpServer/issues/709 and https://github.com/aspnet/KestrelHttpServer/issues/666.

Eilon:
@blowdart because we always freak out with regard to security and so-called "raw" URLs. I'm personally a fan of giving raw access to data, but URLs can be dangerous unless you really, really, *really* know what you're doing.

Tratcher:
Yes, sufficiently so that I'd considered making it its own feature, but that seem a little excessive.

Eilon:
I should add a corollary to my earlier comment:

It is a well-known fact that when it comes to URLs, *no one* knows what they're doing.

blowdart:
I have no real objection to them being exposed. However we mustn't use them anywhere ourselves. We must always operate on the canonicialized string. 

The doc comments should reflect the danger of using URLs in their RAW forms.

Tratcher:
Can we add an attribute that makes using that field a compiler warning you need to suppress? Obsolete(Warning)?

halter73:
I think marking Raw URL getter deprecated/obsolete would be very misleading considering we have no plans to expose the same data in a newer/better API.

Tratcher:
I know Obsolete isn't ideal, but it's the only one I'm aware of that triggers a compiler warning. Are there others?

Eilon:
Please don't mark is as obsolete. There is no other attribute that does this. Either we add the property and it's a real property, or we don't add it at all.

muratg:
@cesarbs Let's do this in 1.0

PR #639, opened 2016-05-25
Add RawTarget property to IHttpRequestFeature (#596)

cesarbs:
https://github.com/aspnet/HttpAbstractions/issues/596

Let me know if I'm missing anything on this repo.

Once this is merged I need to change:

* Kestrel
* WebListener
* What else? :grin: 

Do you think `RawTarget` is a good name for the property? I didn't want to use `Uri`, `Url`, or anything like that since this is called "target" in the RFCs.

cc @Tratcher @halter73 @JunTaoLuo @blowdart  @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Features/IHttpRequestFeature.cs
+++ b/src/Microsoft.AspNetCore.Http.Features/IHttpRequestFeature.cs
@@ -49,6 +49,17 @@ namespace Microsoft.AspNetCore.Http.Features
         string QueryString { get; set; }
 
         /// <summary>
+        /// The request target as it was sent in the HTTP request. This property contains the
+        /// raw path and full query, as well as other request targets such as * for OPTIONS
+        /// requests (https://tools.ietf.org/html/rfc7230#section-5.3).
+        /// </summary>
+        /// <remarks>
+        /// This property is not used internally for routing or authorization decisions. It has not
+        /// been UrlDecoded and care should be taken in its use.
+        /// </remarks>
+        string RawTarget { get; set; }
+
+        /// <summary>
         /// Headers included in the request, aggregated by header name. The values are not split
         /// or merged across header lines. E.g. The following headers:
         /// HeaderA: value1, value2
--- a/src/Microsoft.AspNetCore.Owin/OwinFeatureCollection.cs
+++ b/src/Microsoft.AspNetCore.Owin/OwinFeatureCollection.cs
@@ -102,6 +102,12 @@ namespace Microsoft.AspNetCore.Owin
             set { Prop(OwinConstants.RequestQueryString, Utilities.RemoveQuestionMark(value)); }
         }
 
+        string IHttpRequestFeature.RawTarget
+        {
+            get { return string.Empty; }
+            set { throw new NotSupportedException(); }
+        }
+
         IHeaderDictionary IHttpRequestFeature.Headers
         {
             get { return Utilities.MakeHeaderDictionary(Prop<IDictionary<string, string[]>>(OwinConstants.RequestHeaders)); }
--- a/src/Microsoft.AspNetCore.Http/Features/HttpRequestFeature.cs
+++ b/src/Microsoft.AspNetCore.Http/Features/HttpRequestFeature.cs
@@ -17,6 +17,7 @@ namespace Microsoft.AspNetCore.Http.Features
             PathBase = string.Empty;
             Path = string.Empty;
             QueryString = string.Empty;
+            RawTarget = string.Empty;
         }
 
         public string Protocol { get; set; }
@@ -25,6 +26,7 @@ namespace Microsoft.AspNetCore.Http.Features
         public string PathBase { get; set; }
         public string Path { get; set; }
         public string QueryString { get; set; }
+        public string RawTarget { get; set; }
         public IHeaderDictionary Headers { get; set; }
         public Stream Body { get; set; }
     }


src/Microsoft.AspNetCore.Http.Features/IHttpRequestFeature.cs
```diff
@@ -49,6 +49,11 @@ public interface IHttpRequestFeature
         string QueryString { get; set; }
 
         /// <summary>
+        /// The request target as it was sent in the HTTP request.
+        /// </summary>
+        string RawTarget { get; set; }
```
davidfowl: RawTarget? Thats super cryptic

Tratcher:
Check Hosting's TestHost.

src/Microsoft.AspNetCore.Http.Features/IHttpRequestFeature.cs
Tratcher: That's ok in this case, it's not like we want people to use it. It's only the emergency backup.

That said, I'm sure @blowdart has some very strong language to put in the comments to the effect of `Don't even think about using this, you won't do it right and then you'll get hacked.`

cesarbs: It's the technically correct term :stuck_out_tongue: 

What about RawRequest? Or RawPath?

Tratcher: Inaccurate.

davidfowl: I like `RawPath`, it's "hidden" already on the feature and people aren't stupid, let's not try to protect them?

cesarbs: The problem with `RawPath` is that it's incorrect. We'll include the query with it, no just the path. Plus the feature distinguishes between `Path` and `PathBase`, which makes it even more confusing.

Tratcher: Path is wrong, it's Path and Query at a minimum. And for things like `Options *` it's not Path or Query, it's just `*`. Target is the right term.

davidfowl: Call it RawTarget then, we'll be forever explaining this to people 😄

Tratcher: That's what do comments are for :-). Link to the RFC section.

Besides, most people don't even know what feature interfaces are.

davidfowl: Uh huh 😁

cesarbs:
So do I get a squirrel?

cesarbs:
I mean, I guess we have to wait for @blowdart's feedback on stating the new property should not be used unless someone really really really knows what they're doing.

davidfowl:
:shipit:

Tratcher:
You still need more doc comments. An RFC section link and that the filed contains the escaped full path and/or query, or * for Options *.

src/Microsoft.AspNetCore.Owin/OwinFeatureCollection.cs
```diff
@@ -102,6 +102,12 @@ string IHttpRequestFeature.QueryString
             set { Prop(OwinConstants.RequestQueryString, Utilities.RemoveQuestionMark(value)); }
         }
 
+        string IHttpRequestFeature.RawTarget
+        {
+            get { throw new NotSupportedException(); }
```
Tratcher: Property Getters shouldn't throw, return string.Empty. Setters can throw.

davidfowl:
> I mean, I guess we have to wait for @blowdart's feedback on stating the new property should not be used unless someone really really really knows what they're doing.

Leave that for the docs. Let's get the feature in after the changes @Tratcher suggests and worry about that later 😄

blowdart:
Let's worry about it now. It's not just a docs issue, it's an xmldocs issue too.

So how about

/// <remarks>This property is not used internally for routing or authorization decisions. It has not been UrlDecoded and care should be taken in its use.</remarks>

cesarbs:
Updated.

blowdart:
:ship: (I hate that squirrel)

src/Microsoft.AspNetCore.Http.Features/IHttpRequestFeature.cs
```diff
@@ -49,6 +49,15 @@ public interface IHttpRequestFeature
         string QueryString { get; set; }
 
         /// <summary>
+        /// The request target as it was sent in the HTTP request. This property contains the
+        /// escaped path and full query, as well as other request targets such as * for OPTIONS
```
halter73: "escaped path"

Does this mean that the precent encoded characters in path are unescaped or raw? I think the word escaped implies that the server did some sort of escaping.

cesarbs: I was equally confused :smile: But "encoded" here means they are URL-encoded i.e. the `%`s will be there.

cesarbs:
Turns out there was a bit more that I needed to change.

@blowdart I've moved your warning on the property's usage to `<remarks>`. Is that alright?

blowdart:
Yup :)

Tratcher:
:shipit:

src/Microsoft.AspNetCore.Http.Features/IHttpRequestFeature.cs
halter73: Just use "raw" then

halter73:
:shipit:

PR #640, opened 2016-05-25
Another try on form reader

pakrym:
Fixes: https://github.com/aspnet/HttpAbstractions/issues/553

Another possible solution to allocations

@Tratcher @halter73 @rynowak @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.WebUtilities/FormReader.cs
+++ b/src/Microsoft.AspNetCore.WebUtilities/FormReader.cs
@@ -28,6 +28,9 @@ namespace Microsoft.AspNetCore.WebUtilities
         private readonly StringBuilder _builder = new StringBuilder();
         private int _bufferOffset;
         private int _bufferCount;
+        private string _currentKey;
+        private string _currentValue;
+        private bool _endOfStream;
         private bool _disposed;
 
         public FormReader(string data)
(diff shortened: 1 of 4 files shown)


pakrym:
Before:

After:

rynowak:
You should do profiling in release instead of debug :laughing:

rynowak:
Looks right to me

pakrym:
That's why display class is there. :astonished:

rynowak:
Ya, they are structs in release so they aren't allocated unless you actually go async and need to box.

pakrym:
After in Release:

src/Microsoft.AspNetCore.WebUtilities/FormReader.cs
```diff
@@ -114,49 +134,72 @@ public FormReader(Stream stream, Encoding encoding, ArrayPool<char> charPool)
...
+                if (!TryReadWord('&', ValueLengthLimit, ref _currentValue))
                 {
-                    return word;
+                    return false;
                 }
             }
+            return true;
+        }
+
+        private bool TryReadWord(char seperator, int limit, ref string value)
```
muratg: Shouldn't this be `out string value`?

muratg:
This looks good to me. I'll wait a little to see if there are more comments, but if not, I'll put a squirrel here.

src/Microsoft.AspNetCore.WebUtilities/FormReader.cs
```diff
@@ -28,8 +28,12 @@ public class FormReader : IDisposable
         private readonly StringBuilder _builder = new StringBuilder();
         private int _bufferOffset;
         private int _bufferCount;
+        private string _currentKey;
+        private string _currentValue;
+        private bool _eos;
```
Tratcher: `_endOfStream`

test/Microsoft.AspNetCore.WebUtilities.Tests/FormReaderTests.cs
```diff
@@ -39,23 +50,23 @@ public class FormReaderTests
...
+            var formCollection = await ReadForm(new FormReader(body));
 
             Assert.Equal("", formCollection["foo"].ToString());
         }
 
         [Theory]
         [InlineData(true)]
         [InlineData(false)]
-        public async Task ReadFormAsync_EmptyValuedWithAdditionalEntryAllowed(bool bufferRequest)
+        public async Task ReadForm_EmptyValuedWithAdditionalEntryAllowed(bool bufferRequest)
```
Tratcher: Why did you rename the tests after an test extension method?

test/Microsoft.AspNetCore.WebUtilities.Tests/FormReaderAsyncTest.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.WebUtilities
+{
+    public class FormReaderAsyncTest : FormReaderTests
+    {
+        protected override async Task<Dictionary<string, StringValues>> ReadForm(FormReader reader)
```
Tratcher: ReadFormAsync

davidfowl:
:shipit:

Issue #573, opened 2016-02-25
Confusing UriHelper method names


strohhut:
Sorry, I find this helper class a bit confusing.

Why is a method that does more than encoding called `Encode`? It not only encodes it also combines several uri parts into an uri. And why are two methods called `Encode` when one creates an absolute path and the other creates an absolute uri? Why is the unescaped version called so differently (`GetDisplayUrl`)? Why is the convenience method `GetEncodedUrl` named this way although it does exactly the same as the `Encode`method? Only difference is that it is an extension method for `HttpRequest`. Why is there no convenience method for the non-absolute version?

Maybe I just don't get it but here is my proposal:

Why not just have two method names `GetAbsolutePath` and `GetUrl`? Then give those two methods an `encode` or `httpEncode` or `encodeComponents` bool parameter so there is no need for any `Display` methods. Then make overloads of both methods that act as convenience methods for `HttpRequest`.

So we would get

```
string GetAbsolutePath(PathString pathBase, PathString path, QueryString query, FragmentString fragment, bool encodeComponents)
```

gives `/foo/bar`

Note that this would conform to RFC 3986

> A relative reference that begins with a single slash character is
termed an absolute-path reference. A relative reference that does
not begin with a slash character is termed a relative-path reference.

```
string GetUrl(string scheme, HostString host, PathString pathBase, PathString path, QueryString query, FragmentString fragment, bool encodeComponents)
```

gives `http://example.com/foo/bar`

muratg:
Putting in 1.0.0 to consider this post RC2.

NTaylorMullen:
I don't entirely agree with the proposed but I 100% agree that the existing methods are completely confusing and lead users down a pit of failure. Hence spawning my other issue [here](https://github.com/aspnet/HttpAbstractions/issues/647).

We should consider fixing/designing this for RTM seeing how confusing the APIs currently are.

muratg:
@NTaylorMullen @Tratcher Do you guys have a set of proposed names?

cc @Eilon

Tratcher:
One proposal from offline discussion:
Rename `UriHelper.Encode` to `UriHelper.BuildEncoded` or just Build. If you really wanted to give the overloads unique names then how about `BuildAbsolute` vs `BuildRelative`

NTaylorMullen:
:+1: for `BuildAbsolute` / `BuildRelative`

Tratcher:
[...]

PR #648, opened 2016-06-03
Rename UriHelper.Encode

Tratcher:
#573 @NTaylorMullen @muratg @strohhut

Also clean up the sample project so it will run again.

### DIFF
--- a/samples/SampleApp/Program.cs
+++ b/samples/SampleApp/Program.cs
@@ -1,30 +1,21 @@
 using System;
-using System.Diagnostics;
-using Microsoft.Extensions.Primitives;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Extensions;
 
 namespace SampleApp
 {
     public class Program
     {
-        public void Main(string[] args)
+        public static void Main(string[] args)
         {
-            for (int i = 0; i < 10; i++)
+            var query = new QueryBuilder()
             {
-                Stopwatch timer = new Stopwatch();
-                timer.Start();
-                string myString;
-                string[] myArray;
-                StringValues myValues;
-                for (int j = 0; j < 100000000; j++)
-                {
-                    myString = new string('a', 40);
-                    myArray = new[] { myString };
-                    // myValues = new StringValues(myString);
-                    myValues = new StringValues(myArray);
-                }
-                timer.Stop();
-                Console.WriteLine(timer.Elapsed + ", " + Environment.WorkingSet);
-            }
+                { "hello", "world" }
+            }.ToQueryString();
+
+            var uri = UriHelper.BuildAbsolute("http", new HostString("contoso.com"), query: query);
+
+            Console.WriteLine(uri);
         }
     }
 }
--- a/samples/SampleApp/project.json
+++ b/samples/SampleApp/project.json
@@ -1,12 +1,24 @@
 {
   "version": "1.0.0-*",
   "dependencies": {
-    "Microsoft.AspNetCore.Http": "1.0.0-*"
-  },
-  "commands": {
-    "SampleApp": "SampleApp"
+    "Microsoft.AspNetCore.Http": "1.0.0-*",
+    "Microsoft.AspNetCore.Http.Extensions": "1.0.0-*"
   },
   "frameworks": {
-    "net451": {}
+    "net451": { },
+    "netcoreapp1.0": {
+      "imports": [
+        "dnxcore50"
+      ],
+      "dependencies": {
+        "Microsoft.NETCore.App": {
+          "version": "1.0.0-*",
+          "type": "platform"
+        }
+      }
+    }
+  },
+  "buildOptions": {
+    "emitEntryPoint": true
   }
 }
--- a/src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
+++ b/src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
@@ -16,12 +16,12 @@ namespace Microsoft.AspNetCore.Http.Extensions
         /// <summary>
         /// Combines the given URI components into a string that is properly encoded for use in HTTP headers.
         /// </summary>
-        /// <param name="pathBase"></param>
-        /// <param name="path"></param>
-        /// <param name="query"></param>
-        /// <param name="fragment"></param>
+        /// <param name="pathBase">The first portion of the request path associated with application root.</param>
+        /// <param name="path">The portion of the request path that identifies the requested resource.</param>
+        /// <param name="query">The query, if any.</param>
+        /// <param name="fragment">The fragment, if any.</param>
         /// <returns></returns>
-        public static string Encode(
+        public static string BuildRelative(
             PathString pathBase = new PathString(),
             PathString path = new PathString(),
             QueryString query = new QueryString(),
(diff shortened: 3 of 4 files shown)


src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
```diff
@@ -80,7 +80,7 @@ public static string Encode(Uri uri)
         {
             if (uri.IsAbsoluteUri)
             {
-                return Encode(
+                return BuildAbsolute(
```
NTaylorMullen: Curious: Why is there not a corresponding `BuildRelative` call in the `else`?

Tratcher: System.Uri has minimal support for relative uris. It won't parse it into path?query#fragment for you. The most it will do is some basic escaping.

src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
```diff
@@ -101,7 +101,7 @@ public static string Encode(Uri uri)
         /// <returns></returns>
         public static string GetEncodedUrl(this HttpRequest request)
```
NTaylorMullen: Super odd to have an extension method in this class. Also, any reason the above `Encode` method shouldn't be an extension method/renamed `GetEncodedUrl`?

src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
```diff
@@ -21,7 +21,7 @@ public static class UriHelper
         /// <param name="query"></param>
```
NTaylorMullen: Looks like the parameters aren't filled in for the entire class. Do we care?

src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
```diff
@@ -42,7 +42,7 @@ public static class UriHelper
...
         /// <param name="fragment"></param>
         /// <returns></returns>
-        public static string Encode(
+        public static string BuildAbsolute(
             string scheme,
             HostString host,
```
NTaylorMullen: Might make sense to add an overload that takes a string variation of this. Almost every call into this ends up doing `new HostString(...)`

NTaylorMullen:
The Encode => `BuildAbsolute` rename is good. Had a few comments about the API exposed in `UriHelper`, we can talk offline about them if it's easier.

:watch:

samples/SampleApp/Program.cs
```diff
@@ -1,30 +1,21 @@
 using System;
-using System.Diagnostics;
-using Microsoft.Extensions.Primitives;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Extensions;
 
 namespace SampleApp
 {
-    public class Program
+    public static class Program
```
muratg: Nit: no need for static

src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
Tratcher: Encode could be an extension, but I think you'd loose context of what you were encoding it for. It's only used here:
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Extensions/ResponseHeaders.cs#L145

NTaylorMullen: Not sure I follow. Wouldn't it just be:

```C#
Headers.Set(HeaderNames.Location, value == null ? null : value.GetEncodedUrl());
```

Value's a Uri which is the contextual part. It's odd having this method and the below extension method. They both should be consistent whatever is decided.

src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
Tratcher: > Almost every call into this...
Insufficient data. I see 5 references and the ones that call `new HostString` are samples and tests. The actual code references pull the value from `request.Host` or `HostString.FromUriComponent(uri)`.

NTaylorMullen: :ok:, less concerned about this guy. Can always add the additional overloads later if it turns out to be a burden on users.

src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
Tratcher: Encoded in what way and for what purpose? This method does a particular type of encoding for use in Http headers that is not applicable elsewhere.

samples/SampleApp/project.json
```diff
@@ -1,12 +1,29 @@
 {
   "version": "1.0.0-*",
   "dependencies": {
-    "Microsoft.AspNetCore.Http": "1.0.0-*"
+    "Microsoft.NETCore.Platforms": "1.0.1-*",
```
davidfowl: This isn't needed

samples/SampleApp/Program.cs
```diff
@@ -1,30 +1,21 @@
...
     {
-        public void Main(string[] args)
+        public static void Main(string[] args)
```
davidfowl: This sample app is bogus....

src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
davidfowl: Yes

samples/SampleApp/Program.cs
Tratcher: Yes, but it's convenient for debugging things.

src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
NTaylorMullen: Lets talk in person.

samples/SampleApp/project.json
JunTaoLuo: Do we no longer need it for our samples and test projects targeting full clr?

samples/SampleApp/project.json
```diff
@@ -1,12 +1,29 @@
...
+    "Microsoft.AspNetCore.Http": "1.0.0-*",
+    "Microsoft.AspNetCore.Http.Extensions": "1.0.0-*"
   },
   "commands": {
```
JunTaoLuo: remove commands

samples/SampleApp/project.json
davidfowl: That's right. It should be a dependency of the things that need it.

samples/SampleApp/Program.cs
davidfowl: So are unit tests

src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
Tratcher: HttpEncode?

NTaylorMullen: Talked offline, this is :ok:

samples/SampleApp/project.json
```diff
@@ -1,12 +1,25 @@
...
   },
   "frameworks": {
-    "net451": {}
+    "net451": { },
+    "netcoreapp1.0": {
+      "imports": [
+        "dnxcore50"
+      ],
+      "dependencies": {
+        "System.Console": "4.0.0-*",
```
JunTaoLuo: Why do you need this? I think it's pulled in via `Microsoft.NETCore.App` via `NETStandard.Library`.

Tratcher: Ah, removed.

NTaylorMullen:
:shipit:

PR #654, opened 2016-06-13
Added HTTP Status codes from the RFCs 4918 and 7538

0xdeafcafe:
These two RFCs are finalized, and rather useful that I feel we should have in the core framework.

[7538 RFC Spec](https://tools.ietf.org/html/rfc7538)
[4918 RFC Spec](https://tools.ietf.org/html/rfc4918)

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/StatusCodes.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/StatusCodes.cs
@@ -12,6 +12,7 @@ namespace Microsoft.AspNetCore.Http
         public const int Status204NoContent = 204;
         public const int Status205ResetContent = 205;
         public const int Status206PartialContent = 206;
+        public const int Status207MultiStatus = 207;
 
         public const int Status300MultipleChoices = 300;
         public const int Status301MovedPermanently = 301;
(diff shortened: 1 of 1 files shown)


khellang:
Nice. Submitted 308 for Kestrel as well; https://github.com/aspnet/KestrelHttpServer/pull/925

Tratcher:
308 should be fine.

Are you actually using WebDAV? Or are you just adding them for completeness?

0xdeafcafe:
I'm using the `unprocessable_entity` status code, as it's a good one to use if model validation fails, for instance.

I just added the entire RFC while I was at it for completeness sake. Currently I'm doing this, which I'd rather not (I know I could just extend StatusCodes myself, but I thought I'd just do a pull request instead):

``` csharp
if (!ModelState.IsValid)
	return Json(422 /* unprocessable_entity */, ModelState.ToDictionary());

var question = _lawsContext.Questions.FirstOrDefault(q => q.Id == questionId);
if (question == null)
	return Json(StatusCodes.Status404NotFound, new Error<object>("question_not_found"));

// Verify Module Id is real
var scope = _lawsContext.Scopes.FirstOrDefault(r => r.Id == (Guid) model.ScopeId);
if (scope == null)
	return Json(422 /* unprocessable_entity */, new Error<object>("scope_not_found"));
```

Tratcher:
:shipit:

Tratcher:
Rebased and merged. Thanks.

xperiandri:
It would be also nice to have an `UnprocessableEntityObjectResult` class and `UnprocessableEntity()` method on `Controller`

PR #661, opened 2016-06-29
Parse filename* in multipart form

JunTaoLuo:
#651 According to the RFC we should prefer filename* when available. 

@Tratcher


Tratcher:
:shipit:

Issue #351, opened 2015-07-19
MapMiddleware should support rejoining the parent pipeline


davidfowl:
Right now the MapMiddleware either takes the branch or calls next. There should be an option to fallback to the parent pipeline (or maybe that should just be the default), instead of getting a child appbuilder with a 404 by default.

/cc @Tratcher @DamianEdwards

muratg:
Sounds like a good idea. 1.0.0 backlog for now.

Tratcher:
We've theorized about this since the first days of Katana, but nobody has ever actually asked for it.

I think you'd want it to be a distinct keyword rather than just an option because of how much it changes the semantics.

avanderhoorn:
I have a use case for this. We want to have `app.Map("/Glimpse", ...)` which dynamically loads various "child" middleware. If none of these deal with the request, I want it to continue on with the users pipeline, we shouldn't steal everything that is under Glimpse.

avanderhoorn:
With help from @davidfowl, I've done up the following based on the current `app.Map()` implementation - https://gist.github.com/avanderhoorn/00de3272901797060812

Bartmax:
:+1:

In my case, I have an Authorization Server and Resource Server on same project.
I want all my api calls to be excluded from Identity middleware while other routes uses Identity. something like:

```
  public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
        {
            loggerFactory.AddConsole();

            app.UseOAuthBearerAuthentication(opt =>
            {
                opt.AutomaticAuthentication = true;
            });

            app.MapWhen(ctx => !ctx.Request.Path.StartsWithSegments(new PathString("/api")), (subApp) =>
            {
                subApp.UseIdentity();
            });

            // right now, the following lines won't get called if the request doesn't start with the /api path.
            // I would expect to have some kind of (if something do this and continue with the other stuff outside the condition. Like an if flow.
            app.UseMvc();
            ...
        }
```

darrelmiller:
Seems to me that something like that should be called app.UseWhen(...)  It is just a piece of regular middleware that only applies to a subset of the URI space.

tuespetre:
:+1:

Another use case would be to only call 'error-handling' middleware like `app.UseDeveloperExceptionPage()` and `app.UseStatusCodePagesWithReExecute("/Error/{0}")` if the request is not an AJAX request.

PinpointTownes:
[...]

PR #663, opened 2016-06-30
Add UseWhenExtensions and UseWhenExtensionsTests

tuespetre:
Addresses #351


PinpointTownes:
Nice, I had planned to add `UseWhen` to the aspnet-contrib packages, but if it's part of ASP.NET Core, that's cool too :smile:

IMHO, using `Predicate` instead of a delegate accepting a `HttpContext` prevents tons of useful scenarios.

Here's my own version of `UseWhen`, reviewed by @Tratcher last year: https://gist.github.com/PinpointTownes/6d0492cd3d1b3a64a670

tuespetre:
Haha, reinvented wheels :)

The 'Predicate' is just an alias for Func<HttpContext,bool>. I just copied the MapWhenExtensions file and changed it (same for the tests.)

PinpointTownes:
> The 'Predicate' is just an alias for Func.

Sure, but this doesn't allow you to support path-specific middleware registrations.

E.g

```csharp
// Create a new branch where the registered middleware will be executed only for non API calls.
app.UseWhen(context => !context.Request.Path.StartsWithSegments(new PathString("/api")), branch => {
    branch.UseCookieAuthentication(new CookieAuthenticationOptions { ... });
});
```

PinpointTownes:
Okay, my bad. I had not realized you had added an alias at the top of the file :smile:

tuespetre:
Oops, GitHub took my brackets for HTML. I'm on my phone and don't seem to have the backtick key needed for an inline code span. That should have read 'Func(HttpContext,bool)'

PinpointTownes:
Defining custom aliases with common names (like `Predicate`) is a nice trap :smile: 

On a related, an async overload might be nice too.

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UseWhenExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+    public static class UseWhenExtensions
+    {
+        /// <summary>
+        /// Conditionally creates a branch in the request pipeline that is rejoined to the main pipeline.
+        /// </summary>
+        /// <param name="app"></param>
+        /// <param name="predicate">Invoked with the request environment to determine if the branch should be taken</param>
+        /// <param name="configuration">Configures a branch to take</param>
+        /// <returns></returns>
+        public static IApplicationBuilder UseWhen(this IApplicationBuilder app, Predicate predicate, Action<IApplicationBuilder> configuration)
```
Tratcher: This covers the `if (condition) { }` scenario, but is there also a need for `if (condition) { } else { }` or even `if (condition) { } elseif (condition)  { } elseif (condition)  { } else { }` ?

Tratcher: `if (condition) { } else { }` can be done as `if (condition) { } if (!condition) { }`, but it may be inefficient to evaluate the conditional twice.

`if (condition) { } elseif (condition) else { }` can be done as `if (condition) { } else { if (condition) { } else { }  }`

tuespetre: @Tratcher as far as I see it this fills the need of the original issue, if the additional use case(s) ever become in demand they can be designed then...

Tratcher:
Rebased and merged, thanks.

tuespetre:
Thanks!

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UseWhenExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+            configuration(branchBuilder);
+
+            return app.Use(main =>
+            {
+                // This is called only when the main application builder 
+                // is built, not per request.
+                branchBuilder.Run(main);
+                var branch = branchBuilder.Build();
+
+                return async context =>
```
PinpointTownes: nit: the async/await could have been removed here to save a machine state allocation.

tuespetre: :scream: true, true

Issue #660, opened 2016-06-27
PathString over-encodes


Tratcher:
Ported from https://katanaproject.codeplex.com/workitem/462

"The resource path delimiters "(" and ")" are encoded which violates RFC 3986. Please see section 3.3. This breaks a lot of OData clients that rely on a specific URL convention. The offending code can be found in Microsoft.Owin.PathString:ToUriComponent(). 

http://services.odata.org/OData/OData.svc/GetProductsByCategoryId(categoryId=2) 
is encoded as 
http://services.odata.org/OData/OData.svc/GetProductsByCategoryId%28categoryId=2%29 
which is incorrect."

AspNetCore uses a different encoder, but it likely has similar behaviors here.
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs#L67

```
pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
              / "*" / "+" / "," / ";" / "="
pct-encoded   = "%" HEXDIG HEXDIG
unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
```

troydai:
Here's a prototype:  https://github.com/troydai/HttpAbstractions660/blob/master/PrototypeEncoder.cs
```
Category: unreserved
     Input: /abc123.-_~
PathString: /abc123.-_~
 Prototype: /abc123.-_~
              
  Category: colon
     Input: /:
PathString: /%3A
 Prototype: /:
              
  Category: at
     Input: /@
PathString: /@
 Prototype: /@
              
  Category: sub-delims
     Input: /!$&'()*+,;=
PathString: /!$%26%27()*%2B,;%3D
 Prototype: /!$&'()*+,;=
              
  Category: reserved
     Input: /?#[]
PathString: /%3F%23%5B%5D
 Prototype: /%3F%23%5B%5D
              
  Category: pct-encoding
     Input: /单行道
PathString: /%E5%8D%95%E8%A1%8C%E9%81%93
 Prototype: /%E5%8D%95%E8%A1%8C%E9%81%93
              
  Category: mixed
     Input: /index/单行道=(x*y)
PathString: /index/%E5%8D%95%E8%A1%8C%E9%81%93%3D(x*y)
 Prototype: /index/%E5%8D%95%E8%A1%8C%E9%81%93=(x*y)
```

[...]

PR #667, opened 2016-07-07
Fix PathString over-encoding

troydai:
Base on RFC 3986, ensure following characters are not encoded

alpha, digit, "-", "_", ".", "~", "@", ":", "/", "!", "$", ";", "=",
"'", "(", ")", "*", "+", ","

Issue: #660 

/cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Internal/PathStringHelper.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Internal/PathStringHelper.cs
@@ -1,0 +1,32 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Http.Internal
+{
+    internal class PathStringHelper
+    {
+        private static bool[] ValidPathChars = {
+            false, false, false, false, false, false, false, false,     // 0x00 - 0x07
+            false, false, false, false, false, false, false, false,     // 0x08 - 0x0F
+            false, false, false, false, false, false, false, false,     // 0x10 - 0x17
+            false, false, false, false, false, false, false, false,     // 0x18 - 0x1F
+            false, true,  false, false, true,  false, true,  true,      // 0x20 - 0x27
+            true,  true,  true,  true,  true,  true,  true,  true,      // 0x28 - 0x2F
+            true,  true,  true,  true,  true,  true,  true,  true,      // 0x30 - 0x37
+            true,  true,  true,  true,  false, true,  false, false,     // 0x38 - 0x3F
+            true,  true,  true,  true,  true,  true,  true,  true,      // 0x40 - 0x47
+            true,  true,  true,  true,  true,  true,  true,  true,      // 0x48 - 0x4F
+            true,  true,  true,  true,  true,  true,  true,  true,      // 0x50 - 0x57
+            true,  true,  true,  false, false, false, false, true,      // 0x58 - 0x5F
+            false, true,  true,  true,  true,  true,  true,  true,      // 0x60 - 0x67
+            true,  true,  true,  true,  true,  true,  true,  true,      // 0x68 - 0x6F
+            true,  true,  true,  true,  true,  true,  true,  true,      // 0x70 - 0x77
+            true,  true,  true,  false, false, false, true,  false,     // 0x78 - 0x7F
+        };
+
+        public static bool IsValidPathChar(char c)
+        {
+            return c < ValidPathChars.Length && ValidPathChars[c];
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
@@ -2,8 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Text.Encodings.Web;
+using System.Text;
 using Microsoft.AspNetCore.Http.Abstractions;
+using Microsoft.AspNetCore.Http.Internal;
 
 namespace Microsoft.AspNetCore.Http
 {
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
```diff
@@ -335,9 +361,28 @@ public override int GetHashCode()
...
+                    c == 0x2E ||    // period
+                    c == 0x5F ||    // underscore
+                    c == 0x7E ||    // tilde
+                    c == 0x40 ||    // @
+                    c == 0x3A ||    // :
+                    c == 0x2F ||    // forward slash
+                    c == 0x21 ||    // !
+                    c == 0x24 ||    // $
+                    c == 0x3B ||    // ;
+                    c == 0x3D;      // =
```
Tratcher: why use hex instead of `'='`?

Tratcher: `&`?

Tratcher: I have sometimes seen this implemented as a pre-initialized bool array with the max range (0-126) that you just index into.

src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
```diff
@@ -71,20 +72,45 @@ public string ToUriComponent()
...
+                    if (buffer == null)
+                    {
+                        buffer = new StringBuilder();
+                    }
+
+                    // wrap up previous chars
+                    buffer.Append(_value.Substring(start, count));
+
+                    // escape the string
+                    buffer.Append(Uri.EscapeDataString(_value.Substring(i, 1)));
```
Tratcher: Can you make it encode consecutive characters?

src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
```diff
@@ -71,20 +72,45 @@ public string ToUriComponent()
...
+                    count = 0;
                 }
             }
 
-            return changed ? string.Join("/", values) : _value;
+            if (count == _value.Length)
+            {
+                return _value;
+            }
+            else
```
Tratcher: else count > 0?

src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
```diff
@@ -71,20 +72,45 @@ public string ToUriComponent()
...
+                else
                 {
-                    changed = true;
+                    if (buffer == null)
+                    {
+                        buffer = new StringBuilder();
+                    }
+
+                    // wrap up previous chars
+                    buffer.Append(_value.Substring(start, count));
```
pranavkm: `Append(_value, start, count);`

src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
```diff
@@ -71,20 +72,45 @@ public string ToUriComponent()
...
             }
-            var values = _value.Split(splitChar);
-            var changed = false;
-            for (var i = 0; i < values.Length; i++)
-            {
-                var value = values[i];
-                values[i] = UrlEncoder.Default.Encode(value);
 
-                if (!changed && value != values[i])
+            StringBuilder buffer = null;
```
davidfowl: Who usually calls this method and how expensive is this change? Looks like we never measured it..

src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
troydai: Good point. that would be faster.

Tratcher:
Have you tried basic profiling?

troydai:
I haven't finished profiling yet. We don't have baseline either, so I'll find some samples and compare the old and new.

src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
troydai: will do.

Tratcher:
Not really interested in the baseline in this case, just checking for hot spots.

src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
troydai: Actually doesn't matter, if `count == 0` the following `substring` won't generate result. In both case the buffer needs to be returned. To save a needless append operation, I'll change it to:

* `count > 0` => append remaining unescaped chars, and then return buffer.
* `count == 0` (else) => return buffer.

src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
troydai: `&` is `0x26` included on the fourth line.

src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
troydai: can do.

troydai:
Local profiling using this sample: https://github.com/troydai/HttpAbstractions660/blob/master/perf.sh

src/Microsoft.AspNetCore.Http.Abstractions/Internal/PathStringHelper.cs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Http.Internal
+{
+    internal class PathStringHelper
+    {
+        private static bool[] escapeMap = {
```
Tratcher: escapeMap -> ValidPathChars

src/Microsoft.AspNetCore.Http.Abstractions/Internal/PathStringHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+            true,  true,  true,  true,  true,  true,  true,  true,      // 0x48 - 0x4F
+            true,  true,  true,  true,  true,  true,  true,  true,      // 0x50 - 0x57
+            true,  true,  true,  false, false, false, false, true,      // 0x58 - 0x5F
+            false, true,  true,  true,  true,  true,  true,  true,      // 0x60 - 0x67
+            true,  true,  true,  true,  true,  true,  true,  true,      // 0x68 - 0x6F
+            true,  true,  true,  true,  true,  true,  true,  true,      // 0x70 - 0x77
+            true,  true,  true,  false, false, false, true,  false,     // 0x78 - 0x7F
+        };
+
+        public static bool IsLegalPathChar(char c)
```
Tratcher: Legal -> Valid

src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
```diff
@@ -71,20 +72,79 @@ public string ToUriComponent()
...
+
+            var start = 0;
+            var count = 0;
+            var toEscape = false;
```
Tratcher: requiresEscaping

Tratcher:
:shipit:

troydai:
rebase and squashed. waiting for CI to pass.

PR #674, opened 2016-07-21
Add buffer size parameter to CopyTo method

pakrym:
We want to stop using FileStreams buffering, but we'd like to use bigger buffer then 4k for static files.

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Extensions/StreamCopyOperation.cs
+++ b/src/Microsoft.AspNetCore.Http.Extensions/StreamCopyOperation.cs
@@ -15,11 +15,29 @@ namespace Microsoft.AspNetCore.Http.Extensions
     {
         private const int DefaultBufferSize = 4096;
 
-        public static async Task CopyToAsync(Stream source, Stream destination, long? count, CancellationToken cancel)
+        /// <summary>Asynchronously reads the bytes from the source stream and writes them to another stream.</summary>
+        /// <returns>A task that represents the asynchronous copy operation.</returns>
+        /// <param name="source">The stream from which the contents will be copied.</param>
+        /// <param name="destination">The stream to which the contents of the current stream will be copied.</param>
+        /// <param name="count">The count of bytes to be copied.</param>
+        /// <param name="cancel">The token to monitor for cancellation requests. The default value is <see cref="P:System.Threading.CancellationToken.None" />.</param>
+        public static Task CopyToAsync(Stream source, Stream destination, long? count, CancellationToken cancel)
+        {
+            return CopyToAsync(source, destination, count, DefaultBufferSize, cancel);
+        }
+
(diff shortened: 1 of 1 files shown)


pakrym:
cc @davidfowl @rynowak @pranavkm

src/Microsoft.AspNetCore.Http.Extensions/StreamCopyOperation.cs
```diff
@@ -15,7 +15,12 @@ public static class StreamCopyOperation
     {
         private const int DefaultBufferSize = 4096;
 
-        public static async Task CopyToAsync(Stream source, Stream destination, long? count, CancellationToken cancel)
+        public static Task CopyToAsync(Stream source, Stream destination, long? count, CancellationToken cancel)
```
pranavkm: Docs while you're editing this file.

pranavkm:
:shipit:

src/Microsoft.AspNetCore.Http.Extensions/StreamCopyOperation.cs
```diff
@@ -15,7 +15,12 @@ public static class StreamCopyOperation
...
+        {
+            return CopyToAsync(source, destination, count, DefaultBufferSize, cancel);
+        }
+
+        public static async Task CopyToAsync(Stream source, Stream destination, long? count, int bufferSize, CancellationToken cancel)
```
kichalla: you missed supplying this buffer size to the Rent call

src/Microsoft.AspNetCore.Http.Extensions/StreamCopyOperation.cs
```diff
@@ -15,11 +15,29 @@ public static class StreamCopyOperation
...
         private const int DefaultBufferSize = 4096;
 
-        public static async Task CopyToAsync(Stream source, Stream destination, long? count, CancellationToken cancel)
+        /// <summary>Asynchronously reads the bytes from the source stream and writes them to another stream.</summary>
+        /// <returns>A task that represents the asynchronous copy operation.</returns>
+        /// <param name="source">The stream from which the contents will be copied.</param>
+        /// <param name="destination">The stream to which the contents of the current stream will be copied.</param>
+        /// <param name="count">The count of bytes to be copied.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="P:System.Threading.CancellationToken.None" />.</param>
+        public static Task CopyToAsync(Stream source, Stream destination, long? count, CancellationToken cancellationToken)
```
rynowak: Renaming the parameter is a source breaking change. I don't think that's allowed for 1.1.0

pakrym:
:up::date:

kichalla:
:shipit:

PR #675, opened 2016-07-22
Propagate Form and ResponseCookie init options

benaadams:
Also has nice side effect of allowing lazier load of the two features

Before
pre-allocations

After
post-allocations

DefaultHttpContext increases by 16 bytes (120bytes->136bytes)
But it drops ResponseCookiesFeature at 56 bytes and FormFeature at 48 bytes

So overall 88 bytes and 2 allocations down per request or 83MB per 1M


muratg:
Putting in 1.1 for investigation

src/Microsoft.AspNetCore.Http/DefaultHttpContext.cs
```diff
@@ -23,15 +25,20 @@ public class DefaultHttpContext : HttpContext
...
+
         public DefaultHttpContext()
             : this(new FeatureCollection())
         {
             Features.Set<IHttpRequestFeature>(new HttpRequestFeature());
             Features.Set<IHttpResponseFeature>(new HttpResponseFeature());
         }
 
-        public DefaultHttpContext(IFeatureCollection features)
+        public DefaultHttpContext(IFeatureCollection features, ObjectPool<StringBuilder> builderPool = null, FormOptions formOptions = null)
```
Tratcher: Split this into two constructors

src/Microsoft.AspNetCore.Http/HttpContextFactory.cs
```diff
@@ -43,18 +43,12 @@ public HttpContext Create(IFeatureCollection featureCollection)
                 throw new ArgumentNullException(nameof(featureCollection));
             }
 
-            var responseCookiesFeature = new ResponseCookiesFeature(featureCollection, _builderPool);
-            featureCollection.Set<IResponseCookiesFeature>(responseCookiesFeature);
```
Tratcher: This change breaks feature chaining. See https://github.com/aspnet/Security/blob/6294badd972882ba707216c211095474ef8c4709/src/Microsoft.AspNetCore.CookiePolicy/CookiePolicyMiddleware.cs#L29

src/Microsoft.AspNetCore.Http/Internal/DefaultHttpRequest.cs
```diff
@@ -41,7 +41,7 @@ public virtual void Uninitialize()
             _features.Fetch(ref _features.Cache.Query, f => new QueryFeature(f));
 
         private IFormFeature FormFeature =>
-            _features.Fetch(ref _features.Cache.Form, this, f => new FormFeature(f));
+            _features.Fetch(ref _features.Cache.Form, this, f => new FormFeature(f, (f._context as DefaultHttpContext)?.FormOptions));
```
henkmollema: Just hopping in, but isn't tying this class to `DefaultHttpContext` this way a little ugly? Why not flow `FormOptions` through the constructor when initializing it like you [flow it](https://github.com/aspnet/HttpAbstractions/pull/675/files#diff-b5cd72dfa894e4474c6091bc1a696735R46) from `DefaultHttpContextFactory` to `DefaultHttpContext`?

Same goes for `DefaultHttpResponse` with the `BuilderPool` property.

benaadams: Pushing it into the init means the .ctor needs to change to
```csharp
public DefaultHttpContext(IFeatureCollection features, ObjectPool<StringBuilder> builderPool, FormOptions formOptions)
{
    Initialize(features);

    var defaultRequest = (_request as DefaultHttpRequest);
    if (defaultRequest != null)
    {
        defaultRequest.FormOptions = formOptions;
    }

    var defaultResponse = (_response as DefaultHttpResponse);
    if (defaultResponse != null)
    {
        defaultResponse.BuilderPool = builderPool;
    }
}
```
So you are doing to `as` tests upfront whether you use them or not?

src/Microsoft.AspNetCore.Http/HttpContextFactory.cs
benaadams: Might be missing it with that example? It will create its own `ResponseCookiesFeature` with the `??`?

Tratcher: Fair, but it looses the pooling.

src/Microsoft.AspNetCore.Http/Internal/DefaultHttpRequest.cs
henkmollema: I mean pushing it to ctor's of `DefaultHttpRequest` (and `DefaultHttpResponse`):
```csharp
private readonly FormOptions _formOptions;

public DefaultHttpRequest(HttpContext context, FormOptions formOptions)
{
    Initialize(context);
    _formOptions = formOptions;
}

// ...

private IFormFeature FormFeature => _features.Fetch(ref _features.Cache.Form, this, f => new FormFeature(f, f._formOptions);
```

And change initializing `DefaultHttpContext` to:
```csharp
protected virtual HttpRequest InitializeHttpRequest() => new DefaultHttpRequest(this, FormOptions);
```

src/Microsoft.AspNetCore.Http/HttpContextFactory.cs
benaadams: Need DI in the feature collection ;-)

src/Microsoft.AspNetCore.Http/Internal/DefaultHttpRequest.cs
benaadams: @henkmollema Updated with state capturing; does increase the size by 16 bytes as both `DefaultHttpReXX` and `DefaultHttpContext` need the reference due to the way initialisation works

benaadams:
@Tratcher @pakrym @davidfowl updated with currying, which makes creation is less messy and adds 16 more bytes; so only 72 bytes saved :disappointed: 

Still breaks the feature chaining @Tratcher referenced, to make it work this 
```csharp
var feature = context.Features.Get<IResponseCookiesFeature>() ?? 
    new ResponseCookiesFeature(context.Features);
```
Would need to change to
```csharp
var feature = context.Response.Cookies != null ?
    context.Features.Get<IResponseCookiesFeature>() :
    new ResponseCookiesFeature(context.Features);
```
Which isn't great...

benaadams:
@rynowak don't think this causes the feature collection to become dirty as is just regular flow construction? (its discards should have been resolved in https://github.com/aspnet/HttpAbstractions/pull/623)

pakrym:
Closing, even considering allocation gains we cant break feature chaining pattern.

PR #678, opened 2016-08-02
Do not run TryParseExact on null dates

pakrym:
# Before

# After

@davidfowl @Tratcher

### DIFF
--- a/src/Microsoft.Net.Http.Headers/HttpRuleParser.cs
+++ b/src/Microsoft.Net.Http.Headers/HttpRuleParser.cs
@@ -241,11 +241,6 @@ namespace Microsoft.Net.Http.Headers
 
         internal static bool TryStringToDate(string input, out DateTimeOffset result)
         {
-            if (input == null)
-            {
-                return false;
-            }
-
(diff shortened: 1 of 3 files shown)


src/Microsoft.Net.Http.Headers/HttpRuleParser.cs
```diff
@@ -241,6 +241,11 @@ internal static string DateToString(DateTimeOffset dateTime)
 
         internal static bool TryStringToDate(string input, out DateTimeOffset result)
         {
+            if (input == null)
```
pranavkm: Does this help to add a check for empty strings or is that not a common occurence?

pakrym: It usually not. This is for the case when header is absent.

Tratcher:
:shipit:

pakrym:
#  After null short-circuit
!(http://i.imgur.com/ADALzcc.png)[http://i.imgur.com/ADALzcc.png]

davidfowl:
🚢 🇮🇹

Issue #683, opened 2016-08-09
app.Map blocks root path


cnblogs-dudu:
`dotnet run` the following code in Startup.cs

``` csharp
app.Map("/", _ =>
{
    _.Run((context) =>
    {
        context.Response.Redirect("/markdown/ToHtml");
        return Task.FromResult(0);
    });
});
```

Then get the following exception.

> Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.ArgumentException: The path must not end with a '/'
> Parameter name: pathMatch
> at Microsoft.AspNetCore.Builder.MapExtensions.Map(IApplicationBuilder app, PathString pathMatch, Action`1 configuration)

Tratcher:
As noted in the PR, Map is a StartsWith check, not an Equals check. All paths start with `/`, so `Map("/", ...` does not have a meaningful behavior.

khellang:
@Tratcher Isn't that just an implementation detail? I think `Map("/", /* blah */)` makes perfect sense. I'd just assume it'd strip the trailing slash and do a `StartsWith(string.Empty)`.

Tratcher:
Compare to the normal scenario:
`Map("/foo"...` matches requests `/foo`, `/foo/`, and `/foo/bar`.
But now you're expecting `Map("/"...` to match request `/`, but not `/foo` or `/foo/bar`?

khellang:
No, I'm expecting it to handle `/`, `/foo`, `/foo/` `/foo/bar` etc. Everything, basically :smile:

khellang:
I know that it makes little sense, since you could just do a `.Run` call, but I don't see why it couldn't be allowed?

davidfowl:
It could be allowed but we don't see any benefit in doing so. You can achieve the same thing without a call to Map

muratg:
We are closing this issue because no further action is planned for this issue. If you still have any issues or questions, please log a new issue with any additional details that you have.

PR #684, opened 2016-08-09
Fix MapExtensions to work with root path #683

cnblogs-dudu:
See #683


Tratcher:
What's the point of Map("/", ... ?

cnblogs-dudu:
In my case, it's for redirecting.
```csharp
app.Map("/", _ =>
{
    _.Run((context) =>
    {
        context.Response.Redirect("/markdown/tohtml");
        return Task.FromResult(0);
    });
});
```
```csharp
public class MarkdownController : Controller
{
    [Route("markdown/tohtml")]
    public string ToHtml()
    {
    }
}
```

PinpointTownes:
I agree with @Tratcher, I don't think this change makes much sense as the whole point of `Map` is to create "virtual directories".

You probably want to do something like that:

```csharp
app.Use((context, next) => {
    if (context.Request.Path == "/") {
        context.Response.Redirect("/markdown/tohtml");
        return Task.FromResult(0);
   }

   return next();
});
```

Tratcher:
Map is a StartsWith check, not an Equals check. Everything StartsWith `/`, so there's no point in using it with Map. @PinpointTownes sample is your best option.

cnblogs-dudu:
You are right. I missed the point of app.Map.

PR #685, opened 2016-08-10
Performance Optimization for ToUriComponent (called by ToString) for HostString

ZestyBread:
Big optimization for HostString which contains only ASCII characters. Every time we called ToString or set the HttpContext's request or response Host Property, we called ToUriComponent, which if it wasn't an IPV6 hostname, would call into CoreFX for translation from Unicode to Punycode. If all of the characters were ASCII to begin with, there is no need to call into it. 
Perf increase for ASCII only Hostnames:

perfimprovement

Around 6x improvement:
perfimprovement2

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Internal/HostStringHelper.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Internal/HostStringHelper.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Http.Internal
+{
+    internal class HostStringHelper
+    {
+        // Allowed Characters:
+        // A-Z, a-z, 0-9, ., 
+        // -, %, [, ], : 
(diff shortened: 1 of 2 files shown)


ZestyBread:
@Tratcher @davidfowl @muratg @troydai

src/Microsoft.AspNetCore.Http.Abstractions/Internal/HostStringHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+        // Allowed Characters:
+        // A-Z, a-z, 0-9, ., 
+        // -, %, [, ], : 
+        // Above for IPV6
+        private static bool[] SafeHostStringChars = {
+            false, false, false, false, false, false, false, false,     // 0x00 - 0x07
+            false, false, false, false, false, false, false, false,     // 0x08 - 0x0F
+            false, false, false, false, false, false, false, false,     // 0x10 - 0x17
+            false, false, false, false, false, false, false, false,     // 0x18 - 0x1F
+            false, false, false, false, false, true,  false, true,      // 0x20 - 0x27
```
Tratcher: 0x27?

troydai:
:shipit:

PR #689, opened 2016-08-16
Add UsePathBase middleware

JunTaoLuo:
https://github.com/aspnet/IISIntegration/issues/236
I'll update IISIntegration to use this middleware once it's reviewed. I'll also file an issue to remove the path base feature in 2.0.

cc @davidfowl @halter73 @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
@@ -1,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Builder.Extensions;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    /// <summary>
+    /// Extension methods for <see cref="IApplicationBuilder"/>.
+    /// </summary>
+    public static class UsePathBaseExtensions
+    {
+        /// <summary>
+        /// Adds a middleware that extracts the specified path base from request path and postpend it to the request path base.
+        /// </summary>
+        /// <param name="app">The <see cref="IApplicationBuilder"/> instance.</param>
+        /// <param name="pathBase">The path base to extract.</param>
+        /// <returns>The <see cref="IApplicationBuilder"/> instance.</returns>
+        public static IApplicationBuilder UsePathBase(this IApplicationBuilder app, PathString pathBase)
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+
+            // Strip trailing slashes
+            pathBase = pathBase.Value?.TrimEnd('/');
+            if (!pathBase.HasValue)
+            {
+                return app;
+            }
+
+            return app.UseMiddleware<UsePathBaseMiddleware>(pathBase);
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseMiddleware.cs
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -8,32 +8,32 @@ using Microsoft.AspNetCore.Http;
 namespace Microsoft.AspNetCore.Builder.Extensions
 {
     /// <summary>
-    /// Respresents a middleware that maps a request path to a sub-request pipeline.
+    /// Represents a middleware that extracts the specified path base from request path and postpend it to the request path base.
     /// </summary>
-    public class MapMiddleware
+    public class UsePathBaseMiddleware
     {
         private readonly RequestDelegate _next;
-        private readonly MapOptions _options;
+        private readonly PathString _pathBase;
 
         /// <summary>
-        /// Creates a new instace of <see cref="MapMiddleware"/>.
+        /// Creates a new instace of <see cref="UsePathBaseMiddleware"/>.
         /// </summary>
         /// <param name="next">The delegate representing the next middleware in the request pipeline.</param>
-        /// <param name="options">The middleware options.</param>
-        public MapMiddleware(RequestDelegate next, MapOptions options)
+        /// <param name="pathBase">The path base to extract.</param>
+        public UsePathBaseMiddleware(RequestDelegate next, PathString pathBase)
         {
             if (next == null)
             {
                 throw new ArgumentNullException(nameof(next));
             }
 
-            if (options == null)
+            if (!pathBase.HasValue)
             {
-                throw new ArgumentNullException(nameof(options));
+                throw new ArgumentException($"{nameof(pathBase)} cannot be null or empty.");
             }
 
             _next = next;
-            _options = options;
+            _pathBase = pathBase;
         }
 
         /// <summary>
@@ -51,22 +51,21 @@ namespace Microsoft.AspNetCore.Builder.Extensions
             PathString matchedPath;
             PathString remainingPath;
 
-            if (context.Request.Path.StartsWithSegments(_options.PathMatch, out matchedPath, out remainingPath))
+            if (context.Request.Path.StartsWithSegments(_pathBase, out matchedPath, out remainingPath))
             {
-                // Update the path
-                var path = context.Request.Path;
-                var pathBase = context.Request.PathBase;
-                context.Request.PathBase = pathBase.Add(matchedPath);
+                var originalPath = context.Request.Path;
+                var originalPathBase = context.Request.PathBase;
                 context.Request.Path = remainingPath;
+                context.Request.PathBase = originalPathBase.Add(matchedPath);
 
                 try
                 {
-                    await _options.Branch(context);
+                    await _next(context);
                 }
                 finally
                 {
-                    context.Request.PathBase = pathBase;
-                    context.Request.Path = path;
+                    context.Request.Path = originalPath;
+                    context.Request.PathBase = originalPathBase;
                 }
             }
             else
--- a/src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
@@ -203,8 +203,8 @@ namespace Microsoft.AspNetCore.Http
         }
 
         /// <summary>
-        /// Determines whether the beginning of this PathString instance matches the specified <see cref="PathString"/> when compared
-        /// using the specified comparison option and returns the remaining segments.
+        /// Determines whether the beginning of this <see cref="PathString"/> instance matches the specified <see cref="PathString"/> and returns
+        /// the remaining segments.
         /// </summary>
         /// <param name="other">The <see cref="PathString"/> to compare.</param>
         /// <param name="remaining">The remaining segments after the match.</param>
@@ -215,8 +215,8 @@ namespace Microsoft.AspNetCore.Http
         }
 
         /// <summary>
-        /// Determines whether the beginning of this <see cref="PathString"/> instance matches the specified <see cref="PathString"/> and returns
-        /// the remaining segments.
+        /// Determines whether the beginning of this <see cref="PathString"/> instance matches the specified <see cref="PathString"/> when compared
+        /// using the specified comparison option and returns the remaining segments.
         /// </summary>
         /// <param name="other">The <see cref="PathString"/> to compare.</param>
         /// <param name="comparisonType">One of the enumeration values that determines how this <see cref="PathString"/> and value are compared.</param>
@@ -235,6 +235,46 @@ namespace Microsoft.AspNetCore.Http
                 }
             }
             remaining = Empty;
+            return false;
+        }
+
+        /// <summary>
+        /// Determines whether the beginning of this <see cref="PathString"/> instance matches the specified <see cref="PathString"/> and returns
+        /// the matched and remaining segments.
+        /// </summary>
+        /// <param name="other">The <see cref="PathString"/> to compare.</param>
+        /// <param name="matched">The matched segments with the original casing in the source value.</param>
+        /// <param name="remaining">The remaining segments after the match.</param>
+        /// <returns>true if value matches the beginning of this string; otherwise, false.</returns>
+        public bool StartsWithSegments(PathString other, out PathString matched, out PathString remaining)
+        {
+            return StartsWithSegments(other, StringComparison.OrdinalIgnoreCase, out matched, out remaining);
+        }
+
+        /// <summary>
+        /// Determines whether the beginning of this <see cref="PathString"/> instance matches the specified <see cref="PathString"/> when compared
+        /// using the specified comparison option and returns the matched and remaining segments.
+        /// </summary>
+        /// <param name="other">The <see cref="PathString"/> to compare.</param>
+        /// <param name="comparisonType">One of the enumeration values that determines how this <see cref="PathString"/> and value are compared.</param>
+        /// <param name="matched">The matched segments with the original casing in the source value.</param>
+        /// <param name="remaining">The remaining segments after the match.</param>
+        /// <returns>true if value matches the beginning of this string; otherwise, false.</returns>
+        public bool StartsWithSegments(PathString other, StringComparison comparisonType, out PathString matched, out PathString remaining)
+        {
+            var value1 = Value ?? string.Empty;
+            var value2 = other.Value ?? string.Empty;
+            if (value1.StartsWith(value2, comparisonType))
+            {
+                if (value1.Length == value2.Length || value1[value2.Length] == '/')
+                {
+                    matched = new PathString(value1.Substring(0, value2.Length));
+                    remaining = new PathString(value1.Substring(value2.Length));
+                    return true;
+                }
+            }
+            remaining = Empty;
+            matched = Empty;
             return false;
         }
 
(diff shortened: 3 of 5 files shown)


src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <returns>The <see cref="IApplicationBuilder"/> instance.</returns>
+        public static IApplicationBuilder UsePathBase(this IApplicationBuilder app, string pathBase)
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+
+            // Strip trailing slashes
+            pathBase = pathBase?.TrimEnd('/');
```
davidfowl: allocations 😄

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// </summary>
+    public static class UsePathBaseExtensions
+    {
+        /// <summary>
+        /// Adds a middleware that extracts the specified path base from request path and postpend it to the request path base.
+        /// </summary>
+        /// <param name="app">The <see cref="IApplicationBuilder"/> instance.</param>
+        /// <param name="pathBase">The path base to extract.</param>
+        /// <returns>The <see cref="IApplicationBuilder"/> instance.</returns>
+        public static IApplicationBuilder UsePathBase(this IApplicationBuilder app, string pathBase)
```
davidfowl: Can we make this a `PathString`? There's an implicit conversion to `string`

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+
+            // Strip trailing slashes
+            pathBase = pathBase?.TrimEnd('/');
+            if (string.IsNullOrEmpty(pathBase))
+            {
+                return app;
+            }
+
+            return app.Use(main =>
```
davidfowl: main? We usually call this next

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+
+                    if (originalPath.StartsWithSegments(pathBase, out remainingPath))
+                    {
+                        var originalPathBase = context.Request.PathBase;
+                        context.Request.Path = remainingPath;
+                        context.Request.PathBase = originalPathBase.Add(pathBase);
+
+                        try
+                        {
+                            return main(context);
```
davidfowl: This isn't right, you need to await here s the finally does the right thing.

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
Tratcher: Please make it an actual class. It will make the stack traces more legible.

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+                return context =>
+                {
+                    var originalPath = context.Request.Path;
+                    PathString remainingPath;
+
+                    if (originalPath.StartsWithSegments(pathBase, out remainingPath))
+                    {
+                        var originalPathBase = context.Request.PathBase;
+                        context.Request.Path = remainingPath;
+                        context.Request.PathBase = originalPathBase.Add(pathBase);
```
Tratcher: careful, you want to preserve the original casing from the remote user. Here you're switching it to the dev input. Yes we treat it as case insensitive, but that doesn't mean all components must. Besides, ANCM obnoxiously makes their value all upper case.

Tratcher: hmm, MapPath has this wrong too. you can fix that one while you're at it.

Tratcher: for reference, people had complained about this for both Kestrel and WebListener, so I think they do it right.

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
davidfowl: lambdas rule 😄

JunTaoLuo: yikes, meant to use `app.Use(async (context, next)=> {...})` overload. But maybe extracting it out into a middleware is more readable.

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
JunTaoLuo: What's the best way to "trim" then? Find the correct index and then use `StringSegment`?

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
JunTaoLuo: oops

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
davidfowl: Why are we trimming?

JunTaoLuo: This is similar to what was done in kestrel https://github.com/aspnet/KestrelHttpServer/blob/4ece3f4b68d7d754adbd07bd70f4869e39679dd7/src/Microsoft.AspNetCore.Server.Kestrel/ServerAddress.cs#L148-L152

JunTaoLuo: I thought the reason was that we didn't want to end up with a path base with trailing slashes?

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
JunTaoLuo: I need to check with you on this.

test/Microsoft.AspNetCore.Http.Abstractions.Tests/UsePathBaseExtensionsTests.cs
```diff
@@ -121,8 +121,8 @@ public void PathBaseAndPathPreserveRequestCasing(string registeredPathBase, stri
         }
 
         [Theory]
-        [InlineData("/b♫se", "", "/b♫se/Something", "/b♫se", "/something")]
-        [InlineData("/b♫se", "/oldb♫se", "/b♫se/Something", "/oldb♫se/b♫se", "/something")]
+        [InlineData("/b♫se", "", "/b♫se/something", "/b♫se", "/something")]
+        [InlineData("/b♫se", "/oldb♫se", "/b♫se/something", "/oldb♫se/b♫se", "/something")]
```
Tratcher: please add some mixed casing to these values

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseMiddleware.cs
```diff
@@ -54,7 +54,7 @@ public UsePathBaseMiddleware(RequestDelegate next, PathString pathBase)
             {
                 var originalPathBase = context.Request.PathBase;
                 context.Request.Path = remainingPath;
-                context.Request.PathBase = originalPathBase.Add(_pathBase);
+                context.Request.PathBase = originalPathBase.Add(originalPath.Value.Substring(0, _pathBase.Value.Length));
```
Tratcher: And Map?

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseMiddleware.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// </summary>
+        /// <param name="next">The delegate representing the next middleware in the request pipeline.</param>
+        /// <param name="pathBase">The path base to extract.</param>
+        public UsePathBaseMiddleware(RequestDelegate next, PathString pathBase)
+        {
+            if (next == null)
+            {
+                throw new ArgumentNullException(nameof(next));
+            }
+            if (!pathBase.HasValue)
```
troydai: Add empty line to separate two if structures?

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseMiddleware.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                throw new ArgumentException($"{nameof(next)} cannot be null or empty.");
```
troydai: The error message is wrong. `"{nameof(pathBase)} cannot be null or empty."`

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseMiddleware.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// <param name="context">The <see cref="HttpContext"/> for the current request.</param>
+        /// <returns>A task that represents the execution of this middleware.</returns>
+        public async Task Invoke(HttpContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            var originalPath = context.Request.Path;
```
troydai: Move this into the following if-else structure?

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseMiddleware.cs
```diff
@@ -0,0 +1,75 @@
...
+            PathString remainingPath;
+
+            if (originalPath.StartsWithSegments(_pathBase, out remainingPath))
+            {
+                var originalPathBase = context.Request.PathBase;
+                context.Request.Path = remainingPath;
+                context.Request.PathBase = originalPathBase.Add(originalPath.Value.Substring(0, _pathBase.Value.Length));
```
troydai: Is there a possibility to combine this line with line __53__ because they essentially just did a string split.

test/Microsoft.AspNetCore.Http.Abstractions.Tests/UsePathBaseExtensionsTests.cs
```diff
@@ -0,0 +1,165 @@
...
+                .UsePathBase(pathBase);
+
+            // Act
+            builder.Build();
+
+            // Assert
+            Assert.False(useCalled);
+        }
+
+        private class ApplicationBuilderWrapper : IApplicationBuilder
```
troydai: It seems to be a general test component. Can you move it to its own file and make it a internal class so other tests can utilize it?

test/Microsoft.AspNetCore.Http.Abstractions.Tests/UsePathBaseExtensionsTests.cs
```diff
@@ -0,0 +1,165 @@
...
+        [InlineData("/base", "", "/ba/se", "", "/ba/se")]
+        [InlineData("/base", "/oldbase", "/something", "/oldbase", "/something")]
+        [InlineData("/base", "/oldbase", "/baseandsomething", "/oldbase", "/baseandsomething")]
+        [InlineData("/base", "/oldbase", "/ba", "/oldbase", "/ba")]
+        [InlineData("/base", "/oldbase", "/ba/se", "/oldbase", "/ba/se")]
+        public void RequestPathBaseNotContainingPathBase_IsNotSplit(string registeredPathBase, string pathBase, string requestPath, string expectedPathBase, string expectedPath)
+        {
+            TestPathBase(registeredPathBase, pathBase, requestPath, expectedPathBase, expectedPath);
+        }
+
```
troydai: redundant line

test/Microsoft.AspNetCore.Http.Abstractions.Tests/UsePathBaseExtensionsTests.cs
JunTaoLuo: I think I'm the only one who tests this functionality.

src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
```diff
@@ -239,6 +239,46 @@ public bool StartsWithSegments(PathString other, StringComparison comparisonType
         }
 
         /// <summary>
+        /// Determines whether the beginning of this PathString instance matches the specified <see cref="PathString"/> when compared
+        /// using the specified comparison option and returns the remaining segments.
+        /// </summary>
+        /// <param name="other">The <see cref="PathString"/> to compare.</param>
+        /// <param name="matched">The matched segments after the match.</param>
```
Tratcher: that preserves the original casing.

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <returns>The <see cref="IApplicationBuilder"/> instance.</returns>
+        public static IApplicationBuilder UsePathBase(this IApplicationBuilder app, PathString pathBase)
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+
+            // Strip trailing slashes
+            pathBase = pathBase.Value?.TrimEnd('/');
```
davidfowl: Can value be null?

src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
```diff
@@ -239,6 +239,46 @@ public bool StartsWithSegments(PathString other, StringComparison comparisonType
...
+            return StartsWithSegments(other, StringComparison.OrdinalIgnoreCase, out matched, out remaining);
+        }
+
+        /// <summary>
+        /// Determines whether the beginning of this <see cref="PathString"/> instance matches the specified <see cref="PathString"/> and returns
+        /// the remaining segments.
+        /// </summary>
+        /// <param name="other">The <see cref="PathString"/> to compare.</param>
+        /// <param name="comparisonType">One of the enumeration values that determines how this <see cref="PathString"/> and value are compared.</param>
+        /// <param name="matched">The matched segments with the original casing in the request path.</param>
```
Tratcher: request path? source value?

test/Microsoft.AspNetCore.Http.Abstractions.Tests/UsePathBaseExtensionsTests.cs
```diff
@@ -0,0 +1,168 @@
...
+        [Theory]
+        [InlineData("/b♫se", "", "/b♫se/something", "/b♫se", "/something")]
+        [InlineData("/b♫se", "", "/B♫se/something", "/B♫se", "/something")]
+        [InlineData("/b♫se", "", "/b♫se/Something", "/b♫se", "/Something")]
+        [InlineData("/b♫se", "/oldb♫se", "/b♫se/something", "/oldb♫se/b♫se", "/something")]
+        [InlineData("/b♫se", "/oldb♫se", "/b♫se/Something", "/oldb♫se/b♫se", "/Something")]
+        [InlineData("/b♫se", "/oldb♫se", "/B♫se/something", "/oldb♫se/B♫se", "/something")]
+        public void PathBaseCanHaveUTF8Characters(string registeredPathBase, string pathBase, string requestPath, string expectedPathBase, string expectedPath)
```
Tratcher: UTF8 -> Unicode

Tratcher:
:shipit:

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
JunTaoLuo: You can just call `app.UsePathBase(null)`. I think it's okay to just no-op in this case. But we could also throw an `ArgumentNullException` instead.

davidfowl: PathString is a struct and can't be null.

JunTaoLuo: But the value it wraps can.

JunTaoLuo: Ah... typo in the previous comment, now it's right.

troydai:
:shipit:

poke:
Not sure if commenting here is okay, but I have a question about this code:

```c#
try
{
    await _next(context);
}
finally
{
    context.Request.Path = originalPath;
    context.Request.PathBase = originalPathBase;
}
```

Is there a particular reason why we should revert back to the original path after the middleware pipeline ran? Assuming the `UsePathBase` middleware is the first one the user registers, is there anything in the framework that requires the original path to be set after the middlewares are done?

Tratcher:
It's good behavior to clean up like this as there are some components that may re-execute the pipeline like the those in the diagnostics pipeline. We also have no control over the ordering, that's up to the developer.

poke:
I see, so it’s mostly cleanup just in case. I was just wondering because there’s other [kind-of similar middleware](https://github.com/aspnet/BasicMiddleware/tree/dev/src/Microsoft.AspNetCore.HttpOverrides) that doesn’t care about what happens after. I prefer going this safe route though ;) Thanks a lot for the quick answer @Tratcher!

PR #691, opened 2016-08-17
[WIP] Add lazy features support and use for Forms and Cookies

pakrym:
@Tratcher @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Features/FeatureCollectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.Http.Features/FeatureCollectionExtensions.cs
@@ -1,0 +1,20 @@
+using System;
+
+namespace Microsoft.AspNetCore.Http.Features
+{
+    public static class FeatureCollectionExtensions
+    {
+        public static void TryAddLazy<T>(this IFeatureCollection featureCollection, Func<object, T> factory, object state)
+        {
+            var lazyFeatureCollection = featureCollection as ILazyFeatureCollection;
+            if (lazyFeatureCollection != null)
+            {
+                lazyFeatureCollection.Set(factory, state);
+            }
+            else
+            {
+                featureCollection.Set(factory(state));
+            }
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Http.Features/ILazyFeatureCollection.cs
+++ b/src/Microsoft.AspNetCore.Http.Features/ILazyFeatureCollection.cs
@@ -1,0 +1,9 @@
+using System;
+
+namespace Microsoft.AspNetCore.Http.Features
+{
+    public interface ILazyFeatureCollection : IFeatureCollection
+    {
+        void Set<TFeature>(Func<object, TFeature> factory, object state);
+    }
+}
--- a/src/Microsoft.AspNetCore.Http/DefaultHttpContext.cs
+++ b/src/Microsoft.AspNetCore.Http/DefaultHttpContext.cs
@@ -4,17 +4,21 @@
 using System;
 using System.Collections.Generic;
 using System.Security.Claims;
+using System.Text;
 using System.Threading;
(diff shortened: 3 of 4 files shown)


Tratcher:
Hmm, so we have to re-write Kestrel and WebListener's feature collections to implement this? :-1:

pakrym:
It does fallback to old style collection and implementation is not hard.

rynowak:
Is this just an optimization for the case when those features are never created? For instance right now it's relatively expensive when the feature collection **versions**. If this means that accessing the form feature and query feature will bump the feature collection up to version 3, this is concerning.

Tratcher:
Care to demonstrate the changes on Kestrel?

pakrym:
@Tratcher rough implementation in https://github.com/aspnet/KestrelHttpServer/compare/pakrym/lazy-features

pakrym:
@rynowak yes, this can be an issue.

pakrym:
@rynowak Lazy initialization of value should not change Revision, only SetLazy should.

Tratcher:
That makes the Frame class 2-3x as large as before, so isn't it a net loss for memory consumption?

Tratcher:
@benaadams

Tratcher:
@halter73

pakrym:
Frame is per connection, and we can fine tune which features are stored lazily.

davidfowl:
Not a fan of this new interface, seems to specialized.  The IFeatureCollection itself is already lazy.

halter73:
@davidfowl I told @pakrym we should just make DefaultHttpContext implement IResponseCookiesFeature and IFormFeature. 0 allocations and no new interface :trollface:

benaadams:
> Not a fan of this new interface, seems to specialized. The IFeatureCollection itself is already lazy.

DI feature collection with some lambda pre-compilation for injection resolution for features and services?
 
Or
```csharp
IFeatureCollection.Set<TFeature>();
IFeatureCollection.Set<TFeature>(TFeature instance);
IFeatureCollection.Set<TFeature>(Func<TFeature> implementationFactory);
IFeatureCollection.Set<TFeature>(Func<IFeatureCollection, TFeature> collectionFactory);
IFeatureCollection.Set<TFeature, TState>(Func<TState, TFeature> injectionFactory);
IFeatureCollection.Set<TFeature, TState>(Func<TState, TFeature> stateFactory, TState state);
```

pakrym:
@davidfowl any ideas how can we avoid creating new interface and keep this functionality?

benaadams:
The state could be curried/closure captured into the factories as its always the same? 
e.g.
```csharp
    public class HttpContextFactory : IHttpContextFactory
    {
        private readonly IHttpContextAccessor _httpContextAccessor;
        private static readonly Func<IFeatureCollection, IResponseCookiesFeature> _responseCookiesFeatureFactory;
        private static readonly Func<HttpRequest, IFormFeature> _formFeatureFactory;

        public HttpContextFactory(ObjectPoolProvider poolProvider, IOptions<FormOptions> formOptions, IHttpContextAccessor httpContextAccessor)
        {
            if (poolProvider == null)
            {
                throw new ArgumentNullException(nameof(poolProvider));
            }
            if (formOptions == null)
            {
                throw new ArgumentNullException(nameof(formOptions));
            }

            var builderPool = poolProvider.CreateStringBuilderPool();
            var options = formOptions.Value;

            _responseCookiesFeatureFactory = (features) => {
                return new ResponseCookiesFeature(features, builderPool);
            };
            _formFeatureFactory = (request) => {
                return new FormFeature(request, options);
            };

            _httpContextAccessor = httpContextAccessor;
        }
```

benaadams:
Capturing the state doesn't solve, chaining from the FeatureCollection with state, as it doesn't exist in the feature collection yet :(

PR #695, opened 2016-08-30
Add strong and weak ETag comparisons

JunTaoLuo:
The implementation of `Equals` simply checks whether two ETags are equivalent but does not comform to the RFC's specifications in terms of comparing ETags listed here: https://tools.ietf.org/html/rfc7232#section-2.3.2.

cc @Tratcher

### DIFF
--- a/src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
+++ b/src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
@@ -93,6 +93,15 @@ namespace Microsoft.Net.Http.Headers
             return _tag;
         }
 
+        /// <summary>
+        /// Check against another <see cref="EntityTagHeaderValue"/> for equality.
+        /// This equality check should not be used to determine if two values match under the RFC specifications (https://tools.ietf.org/html/rfc7232#section-2.3.2).
+        /// </summary>
+        /// <param name="obj">The other value to check against for equality.</param>
+        /// <returns>
+        /// <c>true</c> if the strength and tag of the two values match,
+        /// <c>false</c> if the other value is null, is not an <see cref="EntityTagHeaderValue"/>, or if there is a mismatch of strength or tag between the two values.
+        /// </returns>
         public override bool Equals(object obj)
         {
             var other = obj as EntityTagHeaderValue;
@@ -103,13 +112,39 @@ namespace Microsoft.Net.Http.Headers
             }
 
             // Since the tag is a quoted-string we treat it case-sensitive.
-            return ((_isWeak == other._isWeak) && (string.CompareOrdinal(_tag, other._tag) == 0));
+            return _isWeak == other._isWeak && string.Equals(_tag, other._tag, StringComparison.Ordinal);
         }
 
         public override int GetHashCode()
         {
             // Since the tag is a quoted-string we treat it case-sensitive.
             return _tag.GetHashCode() ^ _isWeak.GetHashCode();
+        }
+
+        /// <summary>
+        /// Compares against another <see cref="EntityTagHeaderValue"/> to see if they match under the RFC specifications (https://tools.ietf.org/html/rfc7232#section-2.3.2).
+        /// </summary>
+        /// <param name="other">The other <see cref="EntityTagHeaderValue"/> to compare against.</param>
+        /// <param name="useStrongComparison"><c>true</c> to use a strong comparison, <c>false</c> to use a weak comparison</param>
+        /// <returns>
+        /// <c>true</c> if the <see cref="EntityTagHeaderValue"/> match for the given comparison type,
+        /// <c>false</c> if the other value is null or the comparison failed.
+        /// </returns>
+        public bool Compare(EntityTagHeaderValue other, bool useStrongComparison)
+        {
+            if (other == null)
+            {
+                return false;
+            }
+
+            if (useStrongComparison)
+            {
+                return !IsWeak && !other.IsWeak && string.Equals(Tag, other.Tag, StringComparison.Ordinal);
+            }
+            else
+            {
+                return string.Equals(Tag, other.Tag, StringComparison.Ordinal);
+            }
         }
 
         public static EntityTagHeaderValue Parse(string input)
--- a/test/Microsoft.Net.Http.Headers.Tests/EntityTagHeaderValueTest.cs
+++ b/test/Microsoft.Net.Http.Headers.Tests/EntityTagHeaderValueTest.cs
@@ -100,6 +100,97 @@ namespace Microsoft.Net.Http.Headers
         }
 
         [Fact]
+        public void Compare_WithNull_ReturnsFalse()
+        {
+            Assert.False(EntityTagHeaderValue.Any.Compare(null, useStrongComparison: true));
+            Assert.False(EntityTagHeaderValue.Any.Compare(null, useStrongComparison: false));
+        }
+
+        public static TheoryData<EntityTagHeaderValue, EntityTagHeaderValue> NotEquivalentUnderStrongComparison
+        {
+            get
+            {
+                return new TheoryData<EntityTagHeaderValue, EntityTagHeaderValue>
+                {
+                    { new EntityTagHeaderValue("\"tag\""), new EntityTagHeaderValue("\"TAG\"") },
+                    { new EntityTagHeaderValue("\"tag\"", true), new EntityTagHeaderValue("\"tag\"", true) },
+                    { new EntityTagHeaderValue("\"tag\""), new EntityTagHeaderValue("\"tag\"", true) },
+                    { new EntityTagHeaderValue("\"tag\""), new EntityTagHeaderValue("\"tag1\"") },
+                    { new EntityTagHeaderValue("\"tag\""), EntityTagHeaderValue.Any },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(NotEquivalentUnderStrongComparison))]
+        public void CompareUsingStrongComparison_NonEquivalentPairs_ReturnFalse(EntityTagHeaderValue left, EntityTagHeaderValue right)
+        {
+            Assert.False(left.Compare(right, useStrongComparison: true));
+            Assert.False(right.Compare(left, useStrongComparison: true));
+        }
+
+        public static TheoryData<EntityTagHeaderValue, EntityTagHeaderValue> EquivalentUnderStrongComparison
+        {
+            get
+            {
+                return new TheoryData<EntityTagHeaderValue, EntityTagHeaderValue>
+                {
+                    { new EntityTagHeaderValue("\"tag\""), new EntityTagHeaderValue("\"tag\"") },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(EquivalentUnderStrongComparison))]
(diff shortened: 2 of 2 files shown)


src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
```diff
@@ -103,7 +103,7 @@ public override bool Equals(object obj)
             }
 
             // Since the tag is a quoted-string we treat it case-sensitive.
-            return ((_isWeak == other._isWeak) && (string.CompareOrdinal(_tag, other._tag) == 0));
```
JunTaoLuo: Any reason we were going in a roundabout way of comparing strings previously? Why `CompareOrdinal` instead of `Equals`?

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
```diff
@@ -112,6 +112,24 @@ public override int GetHashCode()
...
+            if (left == null || right == null)
+            {
+                return false;
+            }
+            return !left.IsWeak && !right.IsWeak && string.Equals(left.Tag, right.Tag);
+        }
+
+        public static bool CompareWeakly(EntityTagHeaderValue left, EntityTagHeaderValue right)
+        {
+            if (left == null || right == null)
```
JunTaoLuo: What should we do about comparing two null objects? I'm currently leaning towards considering that situation as not equal. We could consider returning false for strong comparisons and true for weak comparisons.

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
```diff
@@ -112,6 +112,24 @@ public override int GetHashCode()
             return _tag.GetHashCode() ^ _isWeak.GetHashCode();
         }
 
+        public static bool CompareStrongly(EntityTagHeaderValue left, EntityTagHeaderValue right)
```
JunTaoLuo: Do we need (EntityTagHeaderValue.CompareStrongly(EntityTagHeaderValue other)` as well?

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
Tratcher: If you're going to use string.Equals you still need to pass in Ordinal

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
```diff
@@ -112,6 +112,24 @@ public override int GetHashCode()
...
+            return !left.IsWeak && !right.IsWeak && string.Equals(left.Tag, right.Tag);
+        }
+
+        public static bool CompareWeakly(EntityTagHeaderValue left, EntityTagHeaderValue right)
+        {
+            if (left == null || right == null)
+            {
+                return false;
+            }
+            return string.Equals(left.Tag, right.Tag);
```
Tratcher: Ordinal

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
Tratcher: Include the RFC link in the doc comments. Also, doc comments

Tratcher: How about Compare(ETHV other, bool strong)?

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
davidfowl: These API names, where do they come from?

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
JunTaoLuo: The default is an ordinal comparison.

Tratcher: Always specify for string operations, the defaults are not consistant

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
JunTaoLuo: api names? do you mean `CompareWeakly`? I just chose something arbitrary, not a big fan of the name though.

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
JunTaoLuo: Interesting... How and when do they vary?

troydai: When then stars align ... 

Also [MSDN recommended](https://msdn.microsoft.com/en-us/library/dd465121(v=vs.110).aspx#specifying_string_comparisons_explicitly) use overload without default option.

> We recommend that you select an overload that does not use default values, for the following reasons:
> - Some overloads with default parameters (those that search for a Char in the string instance) perform an ordinal comparison, whereas others (those that search for a string in the string instance) are culture-sensitive. It is difficult to remember which method uses which default value, and easy to confuse the overloads.
> - The intent of the code that relies on default values for method calls is not clear. In the following example, which relies on defaults, it is difficult to know whether the developer actually intended an ordinal or a linguistic comparison of two strings, or whether a case difference between protocol and "http" might cause the test for equality to return false.

JunTaoLuo: Fascinating, didn't know the defaults can vary in such ways. I guess it's always better to be on the safe side and just be explicit in what you want to do haha.

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
JunTaoLuo: Or maybe an enum EntityTagComparison.Strong and EntityTagComparison.Weak. Otherwise it's a little confusing what true and false correspond to.

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
JunTaoLuo: Guess I'll go update all the instances in ResponseCaching where I relied on the default value.

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
troydai: Being explicit greatly improves readability as it removes confusions.

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
```diff
@@ -103,7 +103,7 @@ public override bool Equals(object obj)
             }
 
             // Since the tag is a quoted-string we treat it case-sensitive.
-            return ((_isWeak == other._isWeak) && (string.CompareOrdinal(_tag, other._tag) == 0));
+            return _isWeak == other._isWeak && string.Equals(_tag, other._tag);
```
troydai: So add options.

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
troydai: A `bool` is good enough. An enum is more confused since it may indicate there are more than two options.

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
```diff
@@ -112,6 +112,24 @@ public override int GetHashCode()
             return _tag.GetHashCode() ^ _isWeak.GetHashCode();
         }
 
+        public static bool CompareStrongly(EntityTagHeaderValue left, EntityTagHeaderValue right)
+        {
+            if (left == null || right == null)
+            {
+                return false;
+            }
+            return !left.IsWeak && !right.IsWeak && string.Equals(left.Tag, right.Tag);
```
troydai: StringComparison

test/Microsoft.Net.Http.Headers.Tests/EntityTagHeaderValueTest.cs
```diff
@@ -100,6 +100,52 @@ public void Equals_UseSameAndDifferentETags_EqualOrNotEqualNoExceptions()
...
+            Assert.False(EntityTagHeaderValue.CompareStrongly(etag1, null), "tag vs. <null>.");
+            Assert.False(EntityTagHeaderValue.CompareStrongly(null, etag1), "<null> vs tag.");
+            Assert.False(EntityTagHeaderValue.CompareStrongly(null, null), "<null> vs. <null>.");
+            Assert.False(EntityTagHeaderValue.CompareStrongly(etag3, etag3), "weak vs. weak.");
+            Assert.False(EntityTagHeaderValue.CompareStrongly(etag1, etag3), "strong vs. weak.");
+            Assert.False(EntityTagHeaderValue.CompareStrongly(etag3, etag1), "weak vs. strong.");
+            Assert.False(EntityTagHeaderValue.CompareStrongly(etag1, etag4), "tag vs. tag1.");
+            Assert.False(EntityTagHeaderValue.CompareStrongly(etag1, etag6), "tag vs. *.");
+            Assert.True(EntityTagHeaderValue.CompareStrongly(etag1, etag5), "tag vs. tag.");
+        }
```
troydai: Make it theory so that all the cases are tested no matter how many of them are failed. In the current form, if the first assertion filed the following assertions won't be executed.

test/Microsoft.Net.Http.Headers.Tests/EntityTagHeaderValueTest.cs
```diff
@@ -100,6 +100,52 @@ public void Equals_UseSameAndDifferentETags_EqualOrNotEqualNoExceptions()
...
+            Assert.False(EntityTagHeaderValue.CompareWeakly(etag1, null), "tag vs. <null>.");
+            Assert.False(EntityTagHeaderValue.CompareWeakly(null, etag1), "<null> vs tag.");
+            Assert.False(EntityTagHeaderValue.CompareWeakly(null, null), "<null> vs. <null>.");
+            Assert.True(EntityTagHeaderValue.CompareWeakly(etag3, etag3), "weak vs. weak.");
+            Assert.True(EntityTagHeaderValue.CompareWeakly(etag1, etag3), "strong vs. weak.");
+            Assert.True(EntityTagHeaderValue.CompareWeakly(etag3, etag1), "weak vs. strong.");
+            Assert.False(EntityTagHeaderValue.CompareWeakly(etag1, etag4), "tag vs. tag1.");
+            Assert.False(EntityTagHeaderValue.CompareWeakly(etag1, etag6), "tag vs. *.");
+            Assert.True(EntityTagHeaderValue.CompareWeakly(etag1, etag5), "tag vs. tag.");
+        }
```
troydai: Same as above

src/Microsoft.Net.Http.Headers/EntityTagHeaderValue.cs
JunTaoLuo: Hmm I guess I'll go with the bool for now but we often use enums even when only two values are possible to make it more explicit and calls more readable. For example in `DeploymentParameters` for runtime flavour and runtime architecture. Otherwise we have to use named parameters everywhere like `new DeploymentParameters(..., useCoreClrRuntime: true, useX64Architecture: true)`

JunTaoLuo:
:up::date:

Tratcher:
:shipit:

PR #699, opened 2016-09-02
Don't allocate for ResponseCookiesFeature

### LINKED ISSUE #676: ResponseCookiesFeature Allocates when unused
response-cookie-allocation

Resolved by: https://github.com/aspnet/HttpAbstractions/pull/675

benaadams:
* Don't allocate ResponseCookiesFeature when not used
* Don't allocate StringBuilder pool
* Don't allocate StringBuilders
* Don't allocate extra strings when constructing cookie string

Resolves #676

/cc @davidfowl @muratg @Tratcher @pakrym

### DIFF
--- a/src/Microsoft.AspNetCore.Http/Features/ResponseCookiesFeature.cs
+++ b/src/Microsoft.AspNetCore.Http/Features/ResponseCookiesFeature.cs
@@ -16,9 +16,6 @@ namespace Microsoft.AspNetCore.Http.Features
         // Lambda hoisted to static readonly field to improve inlining https://github.com/dotnet/roslyn/issues/13624
         private readonly static Func<IFeatureCollection, IHttpResponseFeature> _nullResponseFeature = f => null;
 
-        // Object pool will be null only in test scenarios e.g. if code news up a DefaultHttpContext.
-        private readonly ObjectPool<StringBuilder> _builderPool;
-
         private FeatureReferences<IHttpResponseFeature> _features;
         private IResponseCookies _cookiesCollection;
 
@@ -50,7 +47,6 @@ namespace Microsoft.AspNetCore.Http.Features
             }
 
             _features = new FeatureReferences<IHttpResponseFeature>(features);
-            _builderPool = builderPool;
         }
 
         private IHttpResponseFeature HttpResponseFeature => _features.Fetch(ref _features.Cache, _nullResponseFeature);
@@ -63,7 +59,7 @@ namespace Microsoft.AspNetCore.Http.Features
                 if (_cookiesCollection == null)
                 {
                     var headers = HttpResponseFeature.Headers;
-                    _cookiesCollection = new ResponseCookies(headers, _builderPool);
+                    _cookiesCollection = new ResponseCookies(headers, null);
                 }
 
                 return _cookiesCollection;
--- a/samples/SampleApp/PooledHttpContextFactory.cs
+++ b/samples/SampleApp/PooledHttpContextFactory.cs
@@ -12,7 +12,6 @@ namespace SampleApp
 {
     public class PooledHttpContextFactory : IHttpContextFactory
     {
-        private readonly ObjectPool<StringBuilder> _builderPool;
         private readonly IHttpContextAccessor _httpContextAccessor;
         private readonly Stack<PooledHttpContext> _pool = new Stack<PooledHttpContext>();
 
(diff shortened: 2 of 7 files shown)


src/Microsoft.AspNetCore.Http/Features/ResponseCookiesFeature.cs
```diff
@@ -27,27 +24,13 @@ public class ResponseCookiesFeature : IResponseCookiesFeature
...
-
-        /// <summary>
-        /// Initializes a new <see cref="ResponseCookiesFeature"/> instance.
-        /// </summary>
-        /// <param name="features">
-        /// <see cref="IFeatureCollection"/> containing all defined features, including this
-        /// <see cref="IResponseCookiesFeature"/> and the <see cref="IHttpResponseFeature"/>.
-        /// </param>
-        /// <param name="builderPool">The <see cref="ObjectPool{T}"/>, if available.</param>
-        public ResponseCookiesFeature(IFeatureCollection features, ObjectPool<StringBuilder> builderPool)
```
Tratcher: Side note: we can't remove public API (even if we change the behavior).

benaadams: Restored. Do I need to add back `AppendToStringBuilder` on `SetCookieHeaderValue`?

Tratcher: Yes, it was public in v1.0

benaadams: Restored

benaadams:
Could also use [string.Concat](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/String.Manipulation.cs#L370-L437) with a single `string[]` which would only be an allocation for the array as its quite efficient memory-wise.

Continually appending to the string is obv problematic, but the cookie is both known size and a known number of strings.

Tratcher:
The string.concat option sounds like it's worth a try.

pakrym:
Would be 2 allocations, 1 for array another for string, I'm thinking about extracting inplace string formatting into a struct, it would look like

```
var isf = new InplaceStringFormatter();
isf.AppendLength(s1);
isf.AppendLength(s2);
isf.AppendLength(s3);
isf.AppendValue(s1);
isf.AppendValue(s1);
isf.AppendValue(s1);
var s= isf.ToString();
```
Could reuse it in https://github.com/aspnet/HttpAbstractions/pull/716 too.

pakrym:
@benaadams how do you feel about moving this PR to use https://github.com/aspnet/Common/pull/157 ?

benaadams:
Yep

benaadams:
Need to wait till stuff builds locally :-/

pakrym:
@benaadams ping

benaadams:
Still can't build
```
Errors in I:\Work\GitHub\HttpAbstractions\test\Microsoft.Net.Http.Headers.Tests\project.json
    Package Microsoft.Extensions.Testing.Abstractions 1.0.0-preview2-003121 is not compatible with netcoreapp1.0 (.NETCoreApp,Version=v1.0). Package Microsoft.Extensions.Testing.Abstractions 1.0.0-preview2-003121 supports:
      - net451 (.NETFramework,Version=v4.5.1)
      - netstandard1.6 (.NETStandard,Version=v1.6)
    One or more packages are incompatible with .NETCoreApp,Version=v1.0.
```
Etc

benaadams:
Hmm, does that when I use ./build is ok in VS but can't run tests

benaadams:
@pakrym updated

pakrym:
You have old version of dotnet sdk somewhere, check what dotnet.exe gets used because 
```
    Package Microsoft.Extensions.Testing.Abstractions 1.0.0-preview2-003121 is not compatible with netcoreapp1.0 (.NETCoreApp,Version=v1.0). Package Microsoft.Extensions.Testing.Abstractions 1.0.0-preview2-003121 supports:
      - net451 (.NETFramework,Version=v4.5.1)
      - netstandard1.6 (.NETStandard,Version=v1.6)
```

was fixed long long time ago.

pakrym:
https://github.com/aspnet/HttpAbstractions/commit/45696535048559188dce8b38a9fde82baa88893c

khellang:
Is there a reason why the `(Pooled)HttpContextFactory` ctors taking an `ObjectPoolProvider` wasn't obsoleted with this PR?

Issue #702, opened 2016-09-05
Add PathBase to request as header


eByte23:
Hi,

I recently encountered the issue when reverse proxy with nginx with my site on a sub directory
and I understand this was the fix for future.
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseMiddleware.cs

Would a pr doing something similar to my example here be excepted, or up for discussion as this makes life easy.

https://github.com/eByte23/AspNetCore-pathbase/blob/master/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseMiddleware.cs

Thanks

davidfowl:
It doesn't make sense to send it as a header, it's static configuration and should be sent once. That's what happens for IIS in the same configuration. How are you configuring NGINX to pass that information to the application?

eByte23:
Sure, I understand, even if we utilised the same environment varible sent in iis
ASPNETCORE_APPL_PATH?

Did it that same way?

```
server {
        listen       80;
        server_name  localhost;

        location /api {
            proxy_pass http://localhost:17425;
            proxy_set_header Host $host;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_set_header ASPNETCORE-PATH-BASE "/api";
            proxy_http_version 1.1;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection keep-alive;
            proxy_cache_bypass $http_upgrade;
        }
    }
}
```

eByte23:
I just think it should be nice than saying you must define inside a configuration or hard code when other things are done in environment varibles

davidfowl:
@eByte23 What you're trying to achieve is fine. I just don't want any automatic behavior in lower level components. Maybe there's another component that understands `ASPNETCORE-PATH-BASE` as an env variable or header and does the right thing but it shouldn't be the default experience. It should be code in your application, not code in the framework. We've worked extremely hard to remove as much policy as possible from lower level components.

BTW, what stops a client from setting that header and spoofing the path base? Isn's there a security concern there?

eByte23:
@davidfowl Sure, but how would you suggest one would use this for the circumstance where "hey i'll just put this app on a sub dir..." means you would have to change your app code or have a config in there all the time to read it.

If its a no we are never going to this, that's fine I'll just make my own modified version for the env/header version.

[...]

PR #703, opened 2016-09-05
Add UsePathBaseEnvironment, gets PathBase from Environment Variable

eByte23:
Add UsePathBaseEnvironment, gets PathBase from Environment Variable ASPNETCORE_APPL_PATH


davidfowl:
Yea this shouldn't be part of HttpAbstractions. This should be part of hosting. It would use a variable set in the config and use this middleware to set the environment.

eByte23:
@davidfowler so should I change this to be a pr on hosting and change middleware accordingly to do something like create a IStartupFilter, similar to the IISIntergration?

Or not bother?

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs
```diff
@@ -12,13 +12,14 @@ namespace Microsoft.AspNetCore.Builder
...
     {
+        private static readonly string ServerPath = "APPL_PATH";
         /// <summary>
         /// Adds a middleware that extracts the specified path base from request path and postpend it to the request path base.
         /// </summary>
         /// <param name="app">The <see cref="IApplicationBuilder"/> instance.</param>
         /// <param name="pathBase">The path base to extract.</param>
         /// <returns>The <see cref="IApplicationBuilder"/> instance.</returns>
-        public static IApplicationBuilder UsePathBase(this IApplicationBuilder app, PathString pathBase)
+        public static IApplicationBuilder UsePathBase(this IApplicationBuilder app, PathString pathBase, bool isOptional = false)
```
moozzyk: side note - this is a breaking change.

davidfowl: This doesn't belong here anyways.

davidfowl:
Move this issue to hosting as it has come up more than once. Let's come up with something before sending the PR, I don't think this middleware needs any changes

Tratcher:
https://github.com/aspnet/HttpAbstractions/issues/702#issuecomment-245015384

eByte23:
Create new issue in hosting https://github.com/aspnet/Hosting/issues/849

PR #706, opened 2016-09-08
Added decoding tests for the FormReader

davidfowl:
- This is the first step in work to remove char[] allocations from the FormReader

/cc @Tratcher


Tratcher:
:shipit:

Issue #700, opened 2016-09-04
Make RequestHeaders a struct


Yves57:
`RequestHeaders` are generally allocated one time per request (see the [Response Caching Context](https://github.com/aspnet/ResponseCaching/blob/dev/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachingContext.cs#L36) or my PR about compression in `BasicMiddleware` for example), and it is a small object.
So I suppose it would be be better for the GC to make it `struct` instead of `class`.

davidfowl:
+1, I think these things need an overhaul. They are mostly used to get a typed view of the stringy header collection.

rynowak:
Yes, you can't really use these today in perf-sensitive code. We avoid them in MVC 99% of the time for this reason.

Tratcher:
Breaking change?

Also, when we tried this before with IOwinContext/Request/Response the compiler wouldn't allow it because it thought there were issues with the mutable properties (even though there weren't any actual issues because the properties were mutating non-struct state).

Yves57:
In my compression middleware PR, I just removed 'RequestHeaders' allocation, and I use the low level `StringWithQualityHeaderValue.TryParseList()` code.

But at least I think it would be nice to have extension methods like
```
public static IList<StringWithQualityHeaderValue> GetAcceptEncoding(this IHeaderDictionary headers)
```
* Easy to implement because everything already exists
* Clearer user code
* No perf problem
* Not necessary to use a [KnowListParsers](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Extensions/HeaderDictionaryTypeExtensions.cs#L210) map :-)

davidfowl:
Agreed. We can make the breaking changes for v2. All of the existing APIs can be avoided in our code base though. We can also
Improve the quality of the code to be less allocatey without breaking the contract.

muratg:
Breaking change. We'll reconsider in 2.0

aspnet-hello:
This issue was moved to aspnet/Home#2717

PR #709, opened 2016-09-10
Remove allocations and improve Header Quality Values parsing performance

Yves57:
See #700 
Do not allocate List if not necessary in `HttpHeaderParser.TryParseValues()`
Use the `IList` properties to remove a `foreach`.
Create a new parser (faster perfs and less memory use) for Quality values
Performances (to parse 10 millions `StringWithQualityHeaderValue` values) :
- Before: 3.2s
- After: 1.4s

Memory use (for 10000 parsings):
- Before: 
  before
- After:
  after


src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -60,8 +60,8 @@ internal static void SetQuality(IList<NameValueHeaderValue> parameters, double?
                 // Note that the RFC requires decimal '.' regardless of the culture. I.e. using ',' as decimal
                 // separator is considered invalid (even if the current culture would allow it).
                 double qualityValue;
-                if (double.TryParse(qualityParameter.Value, NumberStyles.AllowDecimalPoint,
-                    NumberFormatInfo.InvariantInfo, out qualityValue))
+                int length;
```
Tratcher: unused parameter

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -208,6 +208,100 @@ public static bool TryParseInt64(string value, out long result)
             return long.TryParse(value, NumberStyles.None, NumberFormatInfo.InvariantInfo, out result);
         }
 
+        internal static bool TryParseQualityDouble(string input, int startIndex, out double quality, out int length)
+        {
+            // Fast RFC2616 3.9 Quality value parser (and without memory allocation)
```
Tratcher: This comment belongs above the method.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
Yves57: I agree, but what do you want me to do?

src/Microsoft.Net.Http.Headers/HttpHeaderParser.cs
```diff
@@ -61,13 +61,14 @@ protected virtual bool TryParseValues(IList<string> values, bool strict, out ILi
...
             // can ignore the value.
             parsedValues = null;
-            var results = new List<T>();
+            List<T> results = null;
             if (values == null)
             {
                 return false;
             }
-            foreach (var value in values)
+            for (var i = 0; i < values.Count; i++)
```
davidfowl: 👍

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
Tratcher: Nevermind, I thought it was also unused by the other caller

Tratcher:
@muratg can we get this assigned for review?

Tratcher:
*ping* @muratg

muratg:
We'll get back to this after 1.1.0 is out.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
cesarbs: With C# 7 we can now do `out var` here.

Issue #693, opened 2016-08-25
Add constants for HTTP Methods


jkotalik:
We already have status codes and header names as constants, so we should add these as constants too.

muratg:
This would be helpful. We'll accept PRs, otherwise we'll take a look later.

PR #712, opened 2016-09-23
Adding HTTP method constants

mikaelm12:
For issue [#693](https://github.com/aspnet/HttpAbstractions/issues/693)
@Tratcher @muratg


src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
```diff
@@ -0,0 +1,24 @@
...
+        public const string Connect = "CONNECT";
+        public const string Delete = "DELETE";
+        public const string Get = "GET";
+        public const string Options = "OPTIONS";
+        public const string Patch = "PATCH";
+        public const string Post = "POST";
+        public const string Put = "PUT";
+        public const string Trace = "TRACE";
+
+        public static bool Equals(string firstMethod, string secondMethod)
```
davidfowl: Do we need this? Where are we going to use it?

Tratcher:
https://github.com/aspnet/ResponseCaching/blob/5e12a103a4c204d4b3fb68d07e43ce2299b29cf9/src/Microsoft.AspNetCore.ResponseCaching/ResponseCachePolicyProvider.cs#L20

davidfowl:
a more optimized approach would be to have methods for each of the verbs that does a reference equality before doing the case insensitive matching.

Since we're declaring static strings we may as well use them....

davidfowl:
Should these be consts or static readonly?

muratg:
Let's not do this in 1.1. We'll consider this later.

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
```diff
@@ -0,0 +1,59 @@
...
+        public const string Get = "GET";
+        public const string Options = "OPTIONS";
+        public const string Patch = "PATCH";
+        public const string Post = "POST";
+        public const string Put = "PUT";
+        public const string Trace = "TRACE";
+
+        public static bool IsConnectMethod(string inputMethod)
+        {
+            return object.ReferenceEquals(Connect ,inputMethod) | StringComparer.OrdinalIgnoreCase.Equals(Connect, inputMethod);
```
davidfowl: Wrong or, we don't want bitwise or here. You need `|`.

davidfowl:
@muratg why?

muratg:
@davidfowl unnecessary churn, no?

davidfowl:
Seems minimal and we have similar things like header names. It would be even better if have Kestrel use these:

/cc @halter73 @cesarbs

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
cesarbs: nit: `Connect ,` -> `Connect, `

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        public static bool IsDeleteMethod(string inputMethod)
```
davidfowl: Change `inputMethod` to `method`

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
cesarbs: @davidfowl I think you mean `||`.

davidfowl: I edited my comment 😄

mikaelm12: Oops 😅

cesarbs:
Why do case-insensitive comparison when the RFC says method names are case-sensitive? https://tools.ietf.org/html/rfc7230#section-3.1.1

Tratcher:
@cesarbs because in practice they are case-insensitive.

test/Microsoft.AspNetCore.Owin.Tests/OwinFeatureCollectionTests.cs
```diff
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Http.Features;
 using Xunit;
+using Microsoft.AspNetCore.Http;
```
davidfowl: Sort usings

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
mikaelm12: When you said "you need `|` " I figured you meant an extra one

davidfowl: Yasss

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
```diff
@@ -0,0 +1,59 @@
...
+        public static bool IsPutMethod(string method)
+        {
+            return object.ReferenceEquals(Put, method) || StringComparer.OrdinalIgnoreCase.Equals(Put, method);
+        }
+
+        public static bool IsTraceMethod(string method)
+        {
+            return object.ReferenceEquals(Trace, method) || StringComparer.OrdinalIgnoreCase.Equals(Trace, method);
+        }
+    }
```
Tratcher: we should still have the Equals(string, string) method

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
```diff
@@ -55,5 +55,10 @@ public static bool IsTraceMethod(string method)
         {
             return object.ReferenceEquals(Trace, method) || StringComparer.OrdinalIgnoreCase.Equals(Trace, method);
         }
+
+        public static bool Equals(string left, string right)
```
davidfowl: This method is pointless... We hard code a few methods there's no point having a case insensitive overload here. It adds nothing..

mikaelm12: I'm gonna let you and @Tratcher settle this...

davidfowl: @Tratcher lets decide on this offline.

khellang:
Where's `HEAD`?

mikaelm12:
Its there now 😄. Good catch

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            return object.ReferenceEquals(Delete, method) || StringComparer.OrdinalIgnoreCase.Equals(Delete, method);
+        }
+
+        public static bool IsGetMethod(string method)
+        {
+            return object.ReferenceEquals(Get, method) || StringComparer.OrdinalIgnoreCase.Equals(Get, method);
+        }
+
+        public static bool IsHeadMethod(string method){
```
Tratcher: formatting?

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            return object.ReferenceEquals(Connect, method) || StringComparer.OrdinalIgnoreCase.Equals(Connect, method);
+        }
+
+        public static bool IsDeleteMethod(string method)
+        {
+            return object.ReferenceEquals(Delete, method) || StringComparer.OrdinalIgnoreCase.Equals(Delete, method);
+        }
+
+        public static bool IsGetMethod(string method)
```
Tratcher: HttpMethod.IsGetMethod(string) seems redundant. What about just HttpMethod.IsGet(string)?

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
```diff
@@ -0,0 +1,70 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Http
+{
+    public static class HttpMethods
```
Tratcher: Should this be singular?
`HttpMethod.Get`
`HttpMethod.IsGet(string)`
`HttpMethod.Equals(string, string)`

halter73: Agreed. This is similar to a non-flag enum.

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            return object.ReferenceEquals(Put, method) || StringComparer.OrdinalIgnoreCase.Equals(Put, method);
+        }
+
+        public static bool IsTrace(string method)
+        {
+            return object.ReferenceEquals(Trace, method) || StringComparer.OrdinalIgnoreCase.Equals(Trace, method);
+        }
+
+        public static bool Equals(string left, string right)
```
halter73: Is this really necessary? If anything this could be private and called by the Is* methods.

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        public const string Connect = "CONNECT";
+        public const string Delete = "DELETE";
+        public const string Get = "GET";
+        public const string Head = "HEAD";
+        public const string Options = "OPTIONS";
+        public const string Patch = "PATCH";
+        public const string Post = "POST";
+        public const string Put = "PUT";
+        public const string Trace = "TRACE";
```
halter73: I'm for keeping these const. ReferenceEquals still works across assembly boundaries. Just don't mess up 😛 .

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
mikaelm12: There was a big discussion on this. We're taking it out

davidfowl: its decided, nuke it.

src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs
davidfowl: static readonly

mikaelm12:
⬆️ 📅

PinpointTownes:
> a more optimized approach would be to have methods for each of the verbs that does a reference equality before doing the case insensitive matching.

What's the point of doing that manually? It's already something `StringComparer` does internally so you'll end up making the reference equality check twice :trollface:

davidfowl:
> What's the point of doing that manually? It's already something StringComparer does internally so you'll end up making the reference equality check twice :trollface:

Good to know! https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/StringComparer.cs#L300

http://referencesource.microsoft.com/#mscorlib/system/stringcomparer.cs,285

We'll still do the work to use the same objects everywhere 😄

PinpointTownes:
> We'll still do the work to use the same objects everywhere :smile:

TBH, I'm not sure to understand why (and how) having a static field would be better than a good old constant. I'm definitely not a perf' expert, but I'd expect the CLR to be smart enough to intern these constants to avoid instantiating new string references every time you use them in your code.

khellang:
> I'm definitely not a perf' expert, but I'd expect the CLR to be smart enough to intern these constants to avoid instantiating new strings every time you use them in your code.

Because of interning, there's little to no difference between `const` and `static readonly` strings AFAIK.

Since the compiler will burn `const` values into all assemblies using them, it makes sense to use `static readonly` for values that might change at some point in the future, but I can't imagine any of the HTTP methods would change, ever. :stuck_out_tongue_winking_eye: 

There's also a drawback to using `static readonly`; it can't be used in attributes :cry:

davidfowl:
> TBH, I'm not sure to understand why (and how) having a static field would be better than a good old constant. I'm definitely not a perf' expert, but I'd expect the CLR to be smart enough to intern these constants to avoid instantiating new strings every time you use them in your code.

Strings are deduped within single assembly boundaries. There's no guarantee that strings are "reference equal" across assembly boundaries. Using the same static readonly field guarantees that.

halter73:
> There's no guarantee that strings are "reference equal" across assembly boundaries.

@davidfowl Did you read @khellang's comment?

> Because of interning, there's little to no difference between const and static readonly strings AFAIK.

I've tested this, and this is true. `const string` are reference equal to literals/statics/etc... across assembly boundaries. That's why I was pushing for `const` earlier. I can't see the compiler/runtime team ever changing this considering how breaking it would be to do so.

davidfowl:
> I've tested this, and this is true. const string are reference equal to literals/statics/etc... across assembly boundaries. That's why I was pushing for const earlier. I can't see the compiler/runtime team ever changing this considering how breaking it would be to do so.

I tested it as well and it *does* work. Still best practice is to avoid const leaking across assembly boundaries no matter how little difference it makes on current runtimes. Maybe it'll break when corert comes around 😉

PR #716, opened 2016-09-30
Added custom DateTimeFormatter

### LINKED ISSUE #715: Custom DateTimeFormatter for rfc header values

khellang:
Closes #715 

## Benchmark Results

About ~60% faster, and ~85% less allocations :grin:

image


src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
```diff
@@ -0,0 +1,142 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using static System.Text.Encoding;
+
+namespace Microsoft.Net.Http.Headers
+{
+    internal static class DateTimeFormatter
+    {
+        private static readonly byte[] MonBytes = UTF8.GetBytes("Mon");
```
khellang: Apparently `ASCII` isn't available in `netstandard1.1` :cry:

src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
```diff
@@ -321,7 +321,7 @@ private void SetDate(string parameter, DateTimeOffset? date)
             else
             {
                 // Must always be quoted
-                var dateString = string.Format(CultureInfo.InvariantCulture, "\"{0}\"", HttpRuleParser.DateToString(date.Value));
+                var dateString = string.Format(CultureInfo.InvariantCulture, "\"{0}\"", HeaderUtilities.FormatDate(date.Value));
```
rynowak: Why isn't the quoting part of the formatting?

src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        private const char Comma = ',';
+        private const char Space = ' ';
+        private const char Colon = ':';
+
+        public static unsafe string ToRfc1123String(this DateTimeOffset dateTime)
+        {
+            var offset = 0;
+            char* target = stackalloc char[Rfc1123DateLength];
+            var universalDateTime = dateTime.ToUniversalTime();
```
BrennanConroy: It may be possible to use `dateTime.UtcDateTime` to save the extra allocation that `ToUniversalTime()` does, but I would test perf for both in case `DateTimeOffset` is better for this scenario than `DateTime`

src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
```diff
@@ -0,0 +1,161 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Globalization;
+using static System.Text.Encoding;
+
+namespace Microsoft.Net.Http.Headers
+{
+    internal static class DateTimeFormatter
```
rynowak: I'm unsure of the level of optimization that's really required for this code, but a few off the cuff thoughts:
1. My intuition says there's gotta be a better way of copying the bytes than a `foreach`
2. It might be beneficial to 'pack' all of these bytes into an array + offset for locality
3. We should look into the inlinability of all of this (most related to 1)

src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
rynowak: Doesn't `ToUniversalTime()` return a struct?

src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
khellang: I think this might be a special requirement for `Content-Disposition`?

src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
```diff
@@ -0,0 +1,161 @@
...
+        private static readonly byte[] MarBytes = UTF8.GetBytes(GetMonthName(3));
+        private static readonly byte[] AprBytes = UTF8.GetBytes(GetMonthName(4));
+        private static readonly byte[] MayBytes = UTF8.GetBytes(GetMonthName(5));
+        private static readonly byte[] JunBytes = UTF8.GetBytes(GetMonthName(6));
+        private static readonly byte[] JulBytes = UTF8.GetBytes(GetMonthName(7));
+        private static readonly byte[] AugBytes = UTF8.GetBytes(GetMonthName(8));
+        private static readonly byte[] SepBytes = UTF8.GetBytes(GetMonthName(9));
+        private static readonly byte[] OctBytes = UTF8.GetBytes(GetMonthName(10));
+        private static readonly byte[] NovBytes = UTF8.GetBytes(GetMonthName(11));
+        private static readonly byte[] DecBytes = UTF8.GetBytes(GetMonthName(12));
```
khellang: Uuh. That's a good question. I've been back and forth between `char*` and `byte*` a couple of times because of missing APIs for different targets. I'll revert back to `char[]`.

src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
```diff
@@ -0,0 +1,161 @@
...
+            offset = Format(GmtBytes, target, offset);
+
+            return new string(target, 0, offset);
+        }
+
+        private static unsafe int FormatDayOfWeek(DayOfWeek dayOfWeek, char* target, int offset)
+        {
+            switch (dayOfWeek)
+            {
+                case DayOfWeek.Sunday: return Format(SunBytes, target, offset);
```
khellang: You want these in an array of arrays? :smile:

pakrym: Why not array of strings?

khellang: Yeah, sure. I guess we could just reuse [`DateTimeFormatInfo.AbbreviatedDayNames`](https://msdn.microsoft.com/en-us/library/system.globalization.datetimeformatinfo.abbreviateddaynames(v=vs.110).aspx) and [`DateTimeFormatInfo.AbbreviatedMonthNames`](https://msdn.microsoft.com/en-us/library/system.globalization.datetimeformatinfo.abbreviatedmonthnames(v=vs.110).aspx)

khellang:
That god rid of quite a bit of code 😝

src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
```diff
@@ -0,0 +1,103 @@
...
+            offset = FormatTimeOfDay(universalDateTime.TimeOfDay, target, offset);
+            offset = Format(Space, target, offset);
+            offset = Format(Gmt, target, offset);
+
+            return new string(target, 0, offset);
+        }
+
+        private static unsafe int FormatDayOfWeek(DayOfWeek dayOfWeek, char* target, int offset)
+        {
+            return Format(FormatInfo.AbbreviatedDayNames[(int) dayOfWeek], target, offset);
```
BrennanConroy: Could we cache this array and similarly for the month names
See http://referencesource.microsoft.com/#mscorlib/system/globalization/datetimeformatinfo.cs,1393 for what it does every `get`

khellang: Yeah, using those regressed perf substantially:

image

Caching the arrays *did* help, but it's still worse than it was:

image

pakrym: Where did additional allocations come from?

khellang: Probably a string enumerator. Let me get rid of that `foreach` loop 😉

khellang: Much better ✨ 

image

src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
khellang: Yeah, `ToUniversalTime` just returns another `DateTimeOffset` using `UtcDateTIme`; http://referencesource.microsoft.com/#mscorlib/system/datetimeoffset.cs,683

src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
khellang: What about now?

@hallatore did some testing with `Buffer.Memcpy` earlier today. It didn't change much. He also verified that stuff was inlined nicely.

hallatore: I was testing changing to foreach to use Buffer.memcpy, but the results where identical.

Ref memcpy: https://gist.github.com/hallatore/a10ffd0d69a508d09b74a4f720c0511a

src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
Tratcher: Correct, the date only needs to be quoted in some contexts like CDHV

src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
rynowak: You could also consider vectorizing this where the lengths are known, using `(int *)` or `(long *)`. I think that what's there is probably fine, and this is important it would come up again.

src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
rynowak: Did you try this measuring packing all of these into a single array? Then your switch just returns an offset into the array.

For instance your format could be like:
```
M O N , T U E , W E D , ...
```

Then you can just cast to `(int *) and blit 4 bytes at once including the comma

src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
hallatore: Do we really need to convert if input is already utc?

```csharp
if (dateTime.Offset > TimeSpan.Zero)
    dateTime = dateTime.ToUniversalTime();
```

src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace Microsoft.Net.Http.Headers
+{
+    internal static class DateTimeFormatter
+    {
+        private static readonly DateTimeFormatInfo FormatInfo = CultureInfo.InvariantCulture.DateTimeFormat;
+        private static readonly string[] DayNames = FormatInfo.AbbreviatedDayNames;
+        private static readonly string[] MonthNames = FormatInfo.AbbreviatedMonthNames;
+
+        // The format is "ddd, dd MMM yyyy HH:mm:ss GMT".
+        private const int Rfc1123DateLength = 29;
```
cesarbs: nitpicking but you could have done `Rfc1123DateLength = "ddd, dd MMM yyyy HH:mm:ss GMT".Length`

khellang:
Alright. Here's the current state of this PR:

image

About ~60% faster, and ~85% less allocations.

There are other, faster alternatives, but I'm not sure how far you'd like to stretch; https://gist.github.com/aL3891/2644c97e55f2f6c67b1ade15228031b8

BrennanConroy:
Where did the extra 3 bytes come from?

khellang:
@BrennanConroy It varies a bit from run to run. Not really sure why. @mattwarren?

pakrym:
@khellang we have multiple PR's commit with similar use case of string construction, can you please try using https://github.com/aspnet/Common/pull/157 to format the date time so we can have common pattern in both of them?

khellang:
@pakrym I already have it sitting here; https://github.com/khellang/HttpAbstractions/commit/6d9345fc0aa22fbad9ba242adb7b7e7fdcdfab66. Ran a quick benchmar and it looks like it regressed perf slightly (not sure why), but I think it's nice for consistency. Want me to include it in this PR?

khellang:
The allocations are still the same, though. That was the main point of this PR (and related issue) before things turned into a perf contest :stuck_out_tongue:

pakrym:
@khellang would be awesome.

src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
pakrym: Maybe it worth adding `quoted` argument to `HeaderUtilities.FormatDate`?

khellang:
:up::date: Moved to `InplaceStringBuilder` and added a `quoted` argument.

src/Microsoft.Net.Http.Headers/DateTimeFormatter.cs
```diff
@@ -0,0 +1,100 @@
...
+        public static string ToRfc1123String(this DateTimeOffset dateTime)
+        {
+            return ToRfc1123String(dateTime, false);
+        }
+
+        public static string ToRfc1123String(this DateTimeOffset dateTime, bool quoted)
+        {
+            var universal = dateTime.UtcDateTime;
+
+            var length = quoted ? QuotedRfc1123DateLength : Rfc1123DateLength;
```
Tratcher: QuotedRfc1123DateLength is only used once, you might as well put the +2 here.

src/Microsoft.Net.Http.Headers/project.json
```diff
@@ -24,6 +24,7 @@
   "frameworks": {
     "netstandard1.1": {
       "dependencies": {
+        "Microsoft.Extensions.Primitives": "1.1.0-*",
```
Tratcher: This goes under the root dependencies, it's not framework specific.

khellang: What about `System.Buffers`? That's not framework specific either?

davidfowl: Move them all

khellang: Shouldn't `NETStandard.Library` be moved to be framework-specific? Like https://github.com/khellang/HttpAbstractions/commit/b5f8a8d5c7c3d3406c631ce64d521fa0a3de6f53?

davidfowl: Why?

khellang: Doesn't it pull down lots of stuff you don't need when you target `net451`?

khellang: But does it hurt to move it under `netstandard`? Is this what you're recommending library authors do?

pakrym:
Thank you  @khellang !

PR #717, opened 2016-09-30
Add InplaceStringFormatter

pakrym:
Intended to be used instead of pooled StringBuilder or string.Concat when all parts of string are known.
Does only 1 allocation of resulting string.

@benaadams @davidfowl @Tratcher @halter73 @rynowak


pakrym:
Use cases: https://github.com/aspnet/HttpAbstractions/pull/699, https://github.com/aspnet/HttpAbstractions/pull/716

Tratcher:
Something this fundamental belongs in Extensions/Common, or at least down in WebUtils

benaadams:
Raised issue for unaligned https://github.com/dotnet/corefx/issues/12243

pakrym:
Moved to https://github.com/aspnet/Common/pull/157

PR #733, opened 2016-10-31
Added Base64UrlTextEncoder utility from Security repo

kichalla:
@Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
+++ b/src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
@@ -7,11 +7,23 @@ namespace Microsoft.AspNetCore.WebUtilities
 {
     public static class Base64UrlTextEncoder
     {
+        /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.WebUtilities
+{
+    public static class Base64UrlTextEncoder
```
davidfowl: Where is this used?

Tratcher: He's moving this from Security so it can be shared with MVC's temp data cookie

kichalla:
4dcde8a32990561495de188622973b9e540a41e9

PR #734, opened 2016-11-08
Reduce string allocations in Base64UrlTextEncoder

kichalla:
@pakrym @pranavkm @Tratcher @rynowak

Issue: https://github.com/aspnet/Mvc/issues/5479

### DIFF
--- a/src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
+++ b/src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Text;
+using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.WebUtilities
 {
@@ -15,7 +17,8 @@ namespace Microsoft.AspNetCore.WebUtilities
         /// <returns>Base64 encoded string modified with non-URL encodable characters</returns>
         public static string Encode(byte[] data)
         {
-            return Convert.ToBase64String(data).TrimEnd('=').Replace('+', '-').Replace('/', '_');
+            var encodedValue = Convert.ToBase64String(data);
+            return EncodeInternal(encodedValue);
         }
 
         /// <summary>
@@ -26,17 +29,76 @@ namespace Microsoft.AspNetCore.WebUtilities
         /// <returns>The decoded data.</returns>
         public static byte[] Decode(string text)
         {
-            return Convert.FromBase64String(Pad(text.Replace('-', '+').Replace('_', '/')));
+            return Convert.FromBase64String(DecodeToBase64String(text));
         }
 
-        private static string Pad(string text)
+        // To enable unit testing
+        internal static string EncodeInternal(string base64EncodedString)
         {
-            var padding = 3 - ((text.Length + 3) % 4);
-            if (padding == 0)
+            var length = base64EncodedString.Length;
+            while (length > 0 && base64EncodedString[length - 1] == '=')
+            {
+                length--;
+            }
+
+            if (length == 0)
+            {
+                return string.Empty;
+            }
+
+            var inplaceStringBuilder = new InplaceStringBuilder(length);
+            for (var i = 0; i < length; i++)
+            {
+                if (base64EncodedString[i] == '+')
+                {
+                    inplaceStringBuilder.Append('-');
+                }
+                else if (base64EncodedString[i] == '/')
+                {
+                    inplaceStringBuilder.Append('_');
+                }
+                else
+                {
+                    inplaceStringBuilder.Append(base64EncodedString[i]);
+                }
+            }
+
+            return inplaceStringBuilder.ToString();
+        }
+
+        // To enable unit testing
+        internal static string DecodeToBase64String(string text)
+        {
+            if (string.IsNullOrEmpty(text))
             {
                 return text;
             }
-            return text + new string('=', padding);
+
+            var padLength = 3 - ((text.Length + 3) % 4);
+            var inplaceStringBuilder = new InplaceStringBuilder(capacity: text.Length + padLength);
+
+            for (var i = 0; i < text.Length; i++)
+            {
+                if (text[i] == '-')
+                {
+                    inplaceStringBuilder.Append('+');
+                }
+                else if (text[i] == '_')
+                {
+                    inplaceStringBuilder.Append('/');
+                }
+                else
+                {
+                    inplaceStringBuilder.Append(text[i]);
+                }
+            }
+
+            for (var i = 0; i < padLength; i++)
+            {
+                inplaceStringBuilder.Append('=');
+            }
+
+            return inplaceStringBuilder.ToString();
         }
     }
 }
--- a/src/Microsoft.AspNetCore.WebUtilities/Properties/AssemblyInfo.cs
+++ b/src/Microsoft.AspNetCore.WebUtilities/Properties/AssemblyInfo.cs
@@ -3,9 +3,11 @@
 
 using System.Reflection;
 using System.Resources;
+using System.Runtime.CompilerServices;
 
 [assembly: AssemblyMetadata("Serviceable", "True")]
-[assembly: NeutralResourcesLanguage("en-us")]
 [assembly: AssemblyCompany("Microsoft Corporation.")]
 [assembly: AssemblyCopyright("© Microsoft Corporation. All rights reserved.")]
 [assembly: AssemblyProduct("Microsoft ASP.NET Core")]
+[assembly: InternalsVisibleTo("Microsoft.AspNetCore.WebUtilities.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
+[assembly: NeutralResourcesLanguage("en-us")]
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
```diff
@@ -26,17 +29,89 @@ public static string Encode(byte[] data)
...
+                {
+                    inplaceStringBuilder.Append('/');
+                }
+                else
+                {
+                    inplaceStringBuilder.Append(text[i]);
+                }
+            }
+
+            switch (padLength)
```
pakrym: `for (i -> padlength) Append('=')` ?

src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
```diff
@@ -15,7 +17,8 @@ public static class Base64UrlTextEncoder
         /// <returns>Base64 encoded string modified with non-URL encodable characters</returns>
         public static string Encode(byte[] data)
         {
-            return Convert.ToBase64String(data).TrimEnd('=').Replace('+', '-').Replace('/', '_');
+            var encodedValue = Convert.ToBase64String(data);
```
pakrym: How bad is `Convert.ToBase64String` itself?

test/Microsoft.AspNetCore.WebUtilities.Tests/Base64UrlTextEncoderTests.cs
```diff
@@ -26,5 +27,43 @@ public void DataOfVariousLengthRoundTripCorrectly()
...
+        [InlineData("==", "")]
+        [InlineData("a+b+c+==", "a-b-c-")]
+        [InlineData("a/b/c==", "a_b_c")]
+        [InlineData("a+b/c==", "a-b_c")]
+        [InlineData("a+b/c", "a-b_c")]
+        [InlineData("abcd", "abcd")]
+        public void EncodeInternal_Replaces_UrlEncodableCharacters(string base64EncodedValue, string expectedValue)
+        {
+            // Arrange & Act
+            var result = Base64UrlTextEncoder.EncodeInternal(base64EncodedValue);
```
pakrym: Why don't you test public methods instead?

dougbu:
What do dotMemory tables for a scenario using the `CookieTempDataProvider` look like before and after this change?

test/Microsoft.AspNetCore.WebUtilities.Tests/Base64UrlTextEncoderTests.cs
kichalla: The first test in this class tests the public methods.

src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
```diff
@@ -26,17 +29,89 @@ public static string Encode(byte[] data)
...
+            for (endIndex = base64EncodedString.Length - 1; endIndex >= 0; endIndex--)
+            {
+                if (base64EncodedString[endIndex] != '=')
+                {
+                    break;
+                }
+            }
+
+            int finalStringLength = endIndex + 1;
+            if (finalStringLength == 0)
```
pranavkm: Isn't this going to be at minimum of size 1? `endIndex` is never negative is it?

src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
```diff
@@ -26,17 +29,89 @@ public static string Encode(byte[] data)
...
+        // To enable unit testing
+        internal static string DecodeToBase64String(string text)
+        {
+            if (string.IsNullOrEmpty(text))
             {
                 return text;
             }
-            return text + new string('=', padding);
+
+            int padLength = 3 - ((text.Length + 3) % 4);
```
pranavkm: `var` all day every day

src/Microsoft.AspNetCore.WebUtilities/project.json
```diff
@@ -11,6 +11,7 @@
     ]
   },
   "buildOptions": {
+    "allowUnsafe": true,
```
pranavkm: Why do we need this?

src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
kichalla: This is actually fine (example: if the encoded string is `=` or `==`...probably this is not realistic but we have unit tests covering these scenarios)

kichalla:
> What do dotMemory tables for a scenario using the CookieTempDataProvider look like before and after this change?

@dougbu  Following is the data:

**Scenarios which cause Encode logic to run:**

Total requests: 3000
Saving 1K characters in TempData in each request
Reduction of 18 MB

Total requests: 3000
Saving 5K characters in TempData (5K because it causes chunking of cookies) in each request
Reduction of 80 MB

**Scenarios which cause Decode logic to run:**

Total requests: 3000
Peeking 1K characters in TempData
Reduction of 36 MB

Total requests: 3000
Peeking 5K characters in TempData
Reduction of 164 MB

Tratcher:
% change?

src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
pranavkm: I'm still confused though. When is `endIndex` < 0? It doesn't look it's ever decremented past 0.

src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
```diff
@@ -26,17 +29,81 @@ public static string Encode(byte[] data)
...
+            int endIndex;
+            for (endIndex = base64EncodedString.Length - 1; endIndex >= 0; endIndex--)
+            {
+                if (base64EncodedString[endIndex] != '=')
+                {
+                    break;
+                }
+            }
+
+            var finalStringLength = endIndex + 1;
```
pranavkm: Could this be simplified:

```C#
var length = base64EncodedString.Length;
while (length > 0 && base64EncodedString[length - 1] != '=')
{
    length--;
}

if (length == 0)
{
    return string.Empty;
}

var builder = new InplaceStringBuilder(length);
```

src/Microsoft.AspNetCore.WebUtilities/Properties/AssemblyInfo.cs
```diff
@@ -3,9 +3,11 @@
 
 using System.Reflection;
 using System.Resources;
+using System.Runtime.CompilerServices;
 
 [assembly: AssemblyMetadata("Serviceable", "True")]
-[assembly: NeutralResourcesLanguage("en-us")]
```
pranavkm: Undo

kichalla:
**For storing & reading 1K characters in TempData:**
Before perf change:
String allocation: 98 of 162MB (60% of total allocations)
Convert.ToBase64String: 9MB (9% of string allocations)

After perf change:
String allocation: 62 of 126MB (49% of total allocations)
Convert.ToBase64String: 9MB (14.5% of string allocations)

**For storing & reading 5K characters in TempData:**
Before perf change:
String allocation: 528 of 726MB (72 % of total allocations)
Convert.ToBase64String: 41MB (8% of string allocations)

After perf change:
String allocation: 365 of 562MB (65 % of total allocations)
Convert.ToBase64String: 41MB (11% of string allocations)

pakrym:
@kichalla can you share an "after" memory dump please?

kichalla:
🆙 📅

src/Microsoft.AspNetCore.WebUtilities/Properties/AssemblyInfo.cs
NTaylorMullen: It's below. Looks like it was sorted alphabetically

src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
```diff
@@ -26,17 +29,76 @@ public static string Encode(byte[] data)
         /// <returns>The decoded data.</returns>
         public static byte[] Decode(string text)
         {
-            return Convert.FromBase64String(Pad(text.Replace('-', '+').Replace('_', '/')));
+            return Convert.FromBase64String(DecodeToBase64String(text));
```
NTaylorMullen: nit: Break this apart like you have in the Encode method.

src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
```diff
@@ -26,17 +29,76 @@ public static string Encode(byte[] data)
...
+            {
+                length--;
+            }
+
+            if (length == 0)
+            {
+                return string.Empty;
+            }
+
+            var inplaceStringBuilder = new InplaceStringBuilder(length);
```
NTaylorMullen: Could just use a char[] and then do new string(char[]). Not super familiar with InplaceStringBuilder though. So this might be misguided.

src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
```diff
@@ -26,17 +29,76 @@ public static string Encode(byte[] data)
...
+        internal static string DecodeToBase64String(string text)
+        {
+            if (string.IsNullOrEmpty(text))
             {
                 return text;
             }
-            return text + new string('=', padding);
+
+            var padLength = 3 - ((text.Length + 3) % 4);
+            var inplaceStringBuilder = new InplaceStringBuilder(capacity: text.Length + padLength);
```
NTaylorMullen: Same comment as above about the char[]

src/Microsoft.AspNetCore.WebUtilities/Properties/AssemblyInfo.cs
pranavkm: Sure, but seems unnecessary to create a diff? We generally don't have this sorted elsewhere.

src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
pranavkm: The issue there is that you end up allocating two things (the char[] and the string). `InplaceStringBuilder` modifies the backing buffer for a string so you avoid the first allocation.

NTaylorMullen: Sounds good. Looks like it was misguided 😄

Issue #737, opened 2016-11-22
HeaderUtilities enhancements


JunTaoLuo:
Relocate Kestrel's implementation of parsing Int64 https://github.com/aspnet/KestrelHttpServer/issues/1227
Move helpers in https://github.com/aspnet/ResponseCaching/pull/70 here
Expose CacheControlHeaderValue constants

PR #738, opened 2016-11-22
HeaderUtilities enhancements

JunTaoLuo:
#737 

cc @BrennanConroy @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Http/Internal/ParsingHelpers.cs
+++ b/src/Microsoft.AspNetCore.Http/Internal/ParsingHelpers.cs
@@ -411,12 +411,11 @@ namespace Microsoft.AspNetCore.Http.Internal
                 throw new ArgumentNullException(nameof(headers));
             }
 
-            const NumberStyles styles = NumberStyles.AllowLeadingWhite | NumberStyles.AllowTrailingWhite;
             long value;
             var rawValue = headers[HeaderNames.ContentLength];
             if (rawValue.Count == 1 &&
                 !string.IsNullOrWhiteSpace(rawValue[0]) &&
-                long.TryParse(rawValue[0], styles, CultureInfo.InvariantCulture, out value))
+                HeaderUtilities.TryParseInt64(new StringSegment(rawValue[0]).Trim(), out value))
             {
                 return value;
             }
@@ -433,7 +432,7 @@ namespace Microsoft.AspNetCore.Http.Internal
 
             if (value.HasValue)
             {
-                headers[HeaderNames.ContentLength] = value.Value.ToString(CultureInfo.InvariantCulture);
+                headers[HeaderNames.ContentLength] = HeaderUtilities.FormatInt64(value.Value);
             }
             else
             {
--- a/src/Microsoft.Net.Http.Headers/CacheControlHeaderValue.cs
+++ b/src/Microsoft.Net.Http.Headers/CacheControlHeaderValue.cs
@@ -11,18 +11,18 @@ namespace Microsoft.Net.Http.Headers
 {
     public class CacheControlHeaderValue
     {
-        private const string MaxAgeString = "max-age";
-        private const string MaxStaleString = "max-stale";
-        private const string MinFreshString = "min-fresh";
-        private const string MustRevalidateString = "must-revalidate";
-        private const string NoCacheString = "no-cache";
-        private const string NoStoreString = "no-store";
-        private const string NoTransformString = "no-transform";
-        private const string OnlyIfCachedString = "only-if-cached";
-        private const string PrivateString = "private";
-        private const string ProxyRevalidateString = "proxy-revalidate";
-        private const string PublicString = "public";
-        private const string SharedMaxAgeString = "s-maxage";
+        public static readonly string PublicString = "public";
+        public static readonly string PrivateString = "private";
+        public static readonly string MaxAgeString = "max-age";
+        public static readonly string SharedMaxAgeString = "s-maxage";
+        public static readonly string NoCacheString = "no-cache";
+        public static readonly string NoStoreString = "no-store";
+        public static readonly string MaxStaleString = "max-stale";
+        public static readonly string MinFreshString = "min-fresh";
+        public static readonly string NoTransformString = "no-transform";
+        public static readonly string OnlyIfCachedString = "only-if-cached";
+        public static readonly string MustRevalidateString = "must-revalidate";
+        public static readonly string ProxyRevalidateString = "proxy-revalidate";
 
         // The Cache-Control header is special: It is a header supporting a list of values, but we represent the list
         // as _one_ instance of CacheControlHeaderValue. I.e we set 'SupportsMultipleValues' to 'true' since it is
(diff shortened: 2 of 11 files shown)


src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -208,6 +292,82 @@ public static bool TryParseInt64(string value, out long result)
             return long.TryParse(value, NumberStyles.None, NumberFormatInfo.InvariantInfo, out result);
         }
 
+        internal static unsafe bool TryParseInt32(StringSegment value, out int result)
```
JunTaoLuo: Not sure if we should use the same name as the older slower `TryParseInt32/64` since there are a few behaviour differences. We parse only positive numbers for example, and we don't check for overflows.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -198,6 +199,89 @@ internal static bool AreEqualCollections<T>(ICollection<T> x, ICollection<T> y,
...
+                }
+            }
+
+            return false;
+        }
+
+        private static bool TryParseInt64FromHeaderValue(int startIndex, string headerValue, out long value)
+        {
+            var found = false;
+            while (startIndex != headerValue.Length)
```
JunTaoLuo: Could potentially turn this into unsafe code too. Not sure if that makes it any faster.

src/Microsoft.Net.Http.Headers/CacheControlHeaderValue.cs
```diff
@@ -11,18 +11,18 @@ namespace Microsoft.Net.Http.Headers
...
-        private const string MustRevalidateString = "must-revalidate";
-        private const string NoCacheString = "no-cache";
-        private const string NoStoreString = "no-store";
-        private const string NoTransformString = "no-transform";
-        private const string OnlyIfCachedString = "only-if-cached";
-        private const string PrivateString = "private";
-        private const string ProxyRevalidateString = "proxy-revalidate";
-        private const string PublicString = "public";
-        private const string SharedMaxAgeString = "s-maxage";
+        public const string MaxAgeString = "max-age";
```
Tratcher: public static readonly string

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
Tratcher: At least add doc comments for the behavior expectations.

Tratcher: Does the old one need to parse negatives? I'd be surprised. You can probably have the old one call the new one.

Add some overflow checks.

src/Microsoft.Net.Http.Headers/CacheControlHeaderValue.cs
```diff
@@ -11,18 +11,18 @@ namespace Microsoft.Net.Http.Headers
...
-        private const string MinFreshString = "min-fresh";
-        private const string MustRevalidateString = "must-revalidate";
-        private const string NoCacheString = "no-cache";
-        private const string NoStoreString = "no-store";
-        private const string NoTransformString = "no-transform";
-        private const string OnlyIfCachedString = "only-if-cached";
-        private const string PrivateString = "private";
-        private const string ProxyRevalidateString = "proxy-revalidate";
-        private const string PublicString = "public";
-        private const string SharedMaxAgeString = "s-maxage";
```
halter73: Have we learned nothing from #712!? I actually prefer const as you can see if you read the PR comments, but we should be consistent and use static readonly instead.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -208,6 +295,82 @@ public static bool TryParseInt64(string value, out long result)
...
+                if (ch != end)
+                {
+                    result = 0;
+                    return true;
+                }
+                return true;
+            }
+        }
+
+        public static unsafe bool TryParseInt64(StringSegment value, out long result)
```
Tratcher: tests?

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
halter73: Why is this internal and TryParseInt64 public?

Even tough this is in the HeaderUtilites class, I think that it should probably be renamed. If not, the doc comments should clearly explain why it's different from int.TryParse.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
JunTaoLuo: I was testing it through TryParseTimeSpan but I'll add some tests specific to this method.

src/Microsoft.Net.Http.Headers/CacheControlHeaderValue.cs
JunTaoLuo: image

JunTaoLuo: Apparently we were using these values in a switch statement and they need to be const for that.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
JunTaoLuo: Yea not sure why the old `ParseInt32` is internal but I kept the same encapsulation. Should we make that public too?

Tratcher:
Consume the new API from Request/Response.ContentLength

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -208,6 +295,82 @@ public static bool TryParseInt64(string value, out long result)
...
+                {
+                    result *= 10;
+                    result += digit;
+                    ch++;
+                }
+
+                if (ch != end)
+                {
+                    result = 0;
+                    return true;
```
JunTaoLuo: false

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -208,6 +295,82 @@ public static bool TryParseInt64(string value, out long result)
...
```
JunTaoLuo: false

src/Microsoft.Net.Http.Headers/CacheControlHeaderValue.cs
halter73: @davidfowl

halter73: I don't know if David will be OK with this because of the switch statement. Like I said, I'm fine with consts given we don't mess up the values, but I would like consistency.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
halter73: Ok then, so TryParseInt64 is public because you're using it from Kestrel then? There should be doc comments on the public one explaining the limitations.

JunTaoLuo:
:up::date:

src/Microsoft.Net.Http.Headers/CacheControlHeaderValue.cs
JunTaoLuo: Modified the switch statement. These strings are now public static readonly.

Tratcher: 👎 This regressed the code readability and maintainability. These constants aren't worth it.

JunTaoLuo: Also avoiding a ToLower() which is required since switch on strings are case-sensitive. I'm seeing a one percent performance increase in ResponseCaching with it. Also it's code within a private method, so I'm not too concerned about its readability.

JunTaoLuo:
:up::date: @BrennanConroy @Tratcher @halter73 @davidfowl 

Here's some performance results tested with ResponseCaching:
Before: 323984 RPS, 1379 bytes per request
After: 350473 RPS, 848 bytes per request

src/Microsoft.AspNetCore.Http/Internal/ParsingHelpers.cs
```diff
@@ -411,12 +411,11 @@ private static string DeQuote(string value)
                 throw new ArgumentNullException(nameof(headers));
             }
 
-            const NumberStyles styles = NumberStyles.AllowLeadingWhite | NumberStyles.AllowTrailingWhite;
             long value;
             var rawValue = headers[HeaderNames.ContentLength];
             if (rawValue.Count == 1 &&
                 !string.IsNullOrWhiteSpace(rawValue[0]) &&
-                long.TryParse(rawValue[0], styles, CultureInfo.InvariantCulture, out value))
+                HeaderUtilities.TryParseInt64(new StringSegment(rawValue[0]).Trim(), out value))
```
Tratcher: We should do SetContentLength while we're at it, but I don't know if we have any optimized code for this yet. Kestrel has one but it's for hex.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -198,14 +199,254 @@ internal static bool AreEqualCollections<T>(ICollection<T> x, ICollection<T> y,
...
+        /// null if the parsing failed. The conversion fails if the <paramref name="targetValue"/> was not
+        /// found or could not be parsed as a <see cref="TimeSpan"/>. This parameter is passed uninitialized;
+        /// any value originally supplied in result will be overwritten.
+        /// </param>
+        /// <returns>
+        /// <code>true</code> if <paramref name="targetValue"/> is found and successfully parsed; otherwise,
+        /// <code>false</code>.
+        /// </returns>
+        // e.g. { "headerValue=10, targetHeaderValue=30" }
+        public static bool TryParseTimeSpan(StringValues headerValues, string targetValue, out TimeSpan? value)
```
Tratcher: This name is misleading, you're not parsing a fully formatted time, only integer seconds. You just happen to represent the output as a TimeSpan. TryParseSeconds?

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -198,14 +199,254 @@ internal static bool AreEqualCollections<T>(ICollection<T> x, ICollection<T> y,
...
+        /// </param>
+        /// <returns>
+        /// <code>true</code> if <paramref name="targetValue"/> is contained in <paramref name="headerValues"/>;
+        /// otherwise, <code>false</code>.
+        /// </returns>
+        public static bool Contains(StringValues headerValues, string targetValue)
+        {
+            for (var i = 0; i < headerValues.Count; i++)
+            {
+                var index = headerValues[i].IndexOf(targetValue, StringComparison.OrdinalIgnoreCase);
```
Tratcher: This will give you false positives for partial matches

JunTaoLuo:
:up::date:

cc @Tratcher @halter73 @BrennanConroy

src/Microsoft.AspNetCore.Http/Internal/ParsingHelpers.cs
JunTaoLuo: Modified the `FormatInt64` implementation.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -198,19 +200,325 @@ internal static bool AreEqualCollections<T>(ICollection<T> x, ICollection<T> y,
...
+                            current += HttpRuleParser.GetWhitespaceLength(cacheControlDirectives[i], current);
+                        }
+                    }
+                }
+            }
+
+            return false;
+        }
+
+        private static unsafe bool TryParseInt64FromHeaderValue(int startIndex, string headerValue, out long result)
```
Tratcher: Bleh...

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -198,19 +200,325 @@ internal static bool AreEqualCollections<T>(ICollection<T> x, ICollection<T> y,
...
+                value = -value;
+            }
+
+            char* charBuffer = stackalloc char[_int64MaxStringLength];
+
+            do
+            {
+                // Consider using Math.DivRem()
+                var quotient = value / 10;
+                charBuffer[--position] = (char)(0x30 + (value - quotient * 10)); // 0x30 = '0'
```
BrennanConroy: Use a char* here so we don't need to load position

JunTaoLuo: I gave this a try but there isn't any improvement.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
JunTaoLuo: Updated the implementation.

src/Microsoft.AspNetCore.Http/Internal/ParsingHelpers.cs
davidfowl: /cc @KrzysztofCwalina

PR #739, opened 2016-11-30
Small typo fix

txdv:


src/Microsoft.AspNetCore.Http.Abstractions/HttpRequest.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.AspNetCore.Http
     public abstract class HttpRequest
     {
         /// <summary>
-        /// Gets the <see cref="HttpContext"/> this request;
+        /// Gets the <see cref="HttpContext"/> to this request;
```
Tratcher: "for this request"?

txdv: Can't even fix a type correctly ;)

txdv:
Fixed the typo fix.

src/Microsoft.AspNetCore.Http.Abstractions/HttpRequest.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.AspNetCore.Http
     public abstract class HttpRequest
     {
         /// <summary>
-        /// Gets the <see cref="HttpContext"/> this request;
+        /// Gets the <see cref="HttpContext"/> for this request;
```
campersau: Maybe also change `;` to `.` as in https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Abstractions/HttpResponse.cs#L24
(As you can see this is also wrong. It says `request` instead of `response`)

Tratcher:
Rebased and merged to dev. Thanks :-)

PR #748, opened 2016-12-28
Add 101 status code

JunTaoLuo:
cc @Tratcher


Tratcher:
How about adding a link to http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml at the top of each class?

Add 100, 421, 426, and 431.

PR #750, opened 2017-01-06
Add all status codes in IANA list and update reason phrases

JunTaoLuo:
Now added all IANA status codes. Also added comments for renamed, removed or unofficial status codes.

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/StatusCodes.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/StatusCodes.cs
@@ -8,6 +8,7 @@ namespace Microsoft.AspNetCore.Http
     {
         public const int Status100Continue = 100;
         public const int Status101SwitchingProtocols = 101;
+        public const int Status102Processing = 102;
 
         public const int Status200OK = 200;
         public const int Status201Created = 201;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Http.Abstractions/StatusCodes.cs
```diff
@@ -66,6 +72,7 @@ public static class StatusCodes
         public const int Status505HttpVersionNotsupported = 505;
```
JunTaoLuo: Notsupported => NotSupported?

src/Microsoft.AspNetCore.Http.Abstractions/StatusCodes.cs
```diff
@@ -41,13 +44,16 @@ public static class StatusCodes
         public const int Status410Gone = 410;
         public const int Status411LengthRequired = 411;
         public const int Status412PreconditionFailed = 412;
-        public const int Status413RequestEntityTooLarge = 413;
-        public const int Status414RequestUriTooLong = 414;
+        public const int Status413RequestEntityTooLarge = 413;          // RFC 2616, renamed
```
Tratcher: please don't us space padding for alignment. alignment is not required.

src/Microsoft.AspNetCore.Http.Abstractions/StatusCodes.cs
Tratcher: Bleh. Just leave it.

Issue #723, opened 2016-10-09
HttpContextAccessor behaves differently under .NETStandard1.3 vs .NET 4.5.1


dotnetjunkie:
The `HttpContextAccessor` class (see [here](https://github.com/aspnet/HttpAbstractions/blob/87cd79d6fc54bb4abf07c1e380cd7a9498a78612/src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs)) implementation contains a compiler directive that contains a different implementation for .NETStandard1.3. The .NET 4.5.1 implementation uses `CallContext`, while the .NETStandard1.3 version uses `AsyncLocal<T>` under the covers.

These two versions however behave differently. While the .NET 4.5.1 version of `HttpContextAccessor` is stateless, the .NETStandard version contains state in the form o an instance field. This could potentially cause a system, when deployed under .NET 4.5.1, to run fine, while completely breaking when running under .NETStandard1.3.

This will happen when an application, or third party library creates its own instance of the `HttpContextAccessor`. While the stateless .NET 4.5.1 version will always request the `HttpContext` from the `CallContext`, the .NETStandard version will create a new async-local container per class instance. This will cause ASP.NET Core to set the HttpContext in a different bucket then where the the application or 3rd party library will read from (since it contains its own `AsyncLocal<HttpContext>` instance).

This will obviously cause bugs that are really hard to track down.

The solution is to change the instance field to a class field:

``` c#
static readonly AsyncLocal<HttpContext> _httpContextCurrent = new AsyncLocal<HttpContext>();
```

davidfowl:
Thats a good find. What were you doing that raised this problem?

dotnetjunkie:
I was looking at the configuration for a .NET Core application of a Simple Injector user. He actually registered that Accessor as singleton where Simple Injector created that instance. This made me trigger looking at the source code of the Accessor where I spotted this bug.

tobiasbunyan:
**ActionContextAccessor** seems to have the same issue.  I presumed this was intentional and spent most of yesterday working round it :\  New bug report required?

davidfowl:
We'll fix this in the next release but the issue is pretty easy to avoid.

Tratcher:
Yes please file a matching bug in MVC.

PR #753, opened 2017-01-09
Make HttpContextAccessor static AsyncLocal

Tratcher:
#723 The .NET 4.5.1 version is static but the .NETStandard1.3 version isn't, which can cause problems if you end up with different instances of HttpContextAccessor.

### DIFF
--- a/src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs
+++ b/src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs
@@ -30,7 +30,8 @@ namespace Microsoft.AspNetCore.Http
         }
 
 #elif NETSTANDARD1_3
-        private AsyncLocal<HttpContext> _httpContextCurrent = new AsyncLocal<HttpContext>();
+        private static AsyncLocal<HttpContext> _httpContextCurrent = new AsyncLocal<HttpContext>();
+
         public HttpContext HttpContext
         {
             get


Tratcher:
@dotnetjunkie

davidfowl:
Part of me doesn't like this change. The http context accessor should be a singleton, newing it up now shares the async local state with all other instances.

@dotnetjunkie Remind me how you ran into this.

dotnetjunkie:
@davidfowl, try not to act purely on feelings. That typically not works out well :)

Here's a simple way to reproduce the issue:

``` c#
services.AddTransient<IHttpContextAccessor, HttpContextAccessor>();
```

davidfowl:
The inconsistency is an issue but I'd rather make .NET Framework act like .NET Core than vice versa. Adding a transient `IHttpContextAccessor` is just broken.

dotnetjunkie:
Yes, making that component transient is clearly broken, but by no way does the framework guide developers into the right direction.

It is really easy for a user to do the wrong thing, while it is really hard for the user to track down the bug that is the result of this. I already shown two completely different scenarios (both in the previous comment and in #723) that can cause bugs.

davidfowl:
Agreed but that can be solved in other ways. I'm just of the opinion this is the wrong fix for this particular problem. There are been some huge improvements made to the allocation profile to async locals in .NET Core vnext so maybe it's time to turn this back on by default again and measure. Also dropping .NET 4.5.1 (and rebasing to .NET Framework 4.6) support would let us use async local on both .NET Core and .NET Framework. 

That solves both the lifetime issue and saves the implementation from hard coding statics.

dotnetjunkie:
The only actual fix here is to make the AsyncLocal field static, because it's very easy for a user to accidentally create a second instance of the `HttpContextAccessor` as explained in #723.

davidfowl:
> The only actual fix here is to make the AsyncLocal field static, because it's very easy for a user to accidentally create a second instance of the HttpContextAccessor as explained in #723.

I'm not convinced that would be common if we were able to add this service back to hosting by default.

dotnetjunkie:
> I'm not convinced that would be common if we were able to add this service back to hosting by default.

Wait. What? You are considering registering it by default? Is that what you're saying? I thought the registration was omitted by default because of performance considerations. Can you elaborate on that?

davidfowl:
I mentioned it here https://github.com/aspnet/HttpAbstractions/pull/753#issuecomment-271504778

Tratcher:
As a data point, we have many uses of AsyncLocal and most of them are static.
https://github.com/search?q=org%3Aaspnet+AsyncLocal&type=Code

@davidfowl how much of your concern applies to these other cases?

PR #755, opened 2017-01-13
let UriHelper.GetEncodedUrl  can  get relative url

John0King:


Tratcher:
This should be a separate method with a more specific name. GetEncodedPathAndQuery

Tratcher:
@sebastienros is this what you were asking for? https://github.com/aspnet/Mvc/issues/5660

sebastienros:
Yes. I am not sure what `BuildRelative` does to the concatenated string though.

Tratcher:
BuildRelative just gives you intellisense for what values to concatenate.

Issue #407, opened 2015-09-16
Expose Content-Length in IHttpResponseFeature


benaadams:
At the moment accessing content-length goes to string and back through parsing and the header collection. It is a `long?` on Response.

The string transitions are alloc heavy for 2000 requests at 108k meaning 54 Meg per 1M requests:

![content-length](http://aoa.blob.core.windows.net/aspnet/cl.png)

What I'd much prefer is to have the `long?` directly so I can output that straight to bytes without going via string or allocating.

Tratcher:
Is the cost you're measuring for long -> string -> ascii bytes, and you're trying to shorten it to long -> ascii bytes?

Background: System.Web has had no end of problems special casing certain headers to bypass the header collection. Response cookies are a particularly bad example where it keeps a separate collection for them and only adds them to the headers at the end of the response. This causes conflicts for anyone using the headers directly. We don't want to repeat this mistake so we've kept the headers collection as the source of truth for all headers.

benaadams:
Trying to shorten it to long -> pre allocated byte buffer; so zero alloc.

Am ok with the other headers, as they come with a user generated cost, and where the user has already created the string; but this is generally always present except if chunked and is a hidden library cost that cannot be worked around.

benaadams:
For comparison its 30% of the alloc cost of DI: https://github.com/aspnet/DependencyInjection/pull/291

Tratcher:
StaticFiles does set it but most of our other components don't, they rely on the server to chunk or to buffer and set it for them. In those cases the server can take shortcuts internally.

I can't think of any of our other libraries that set it by default.

rynowak:
MVC does in some extremely obscure scenarios, but not anything @benaadams or the average user is likely to have encountered.

benaadams:
With the current set up this is ok:

```httpContext.Response.Headers["Content-Length"] = "fred";```

But using the strongly typed ```ContentLength``` on ```HttpResponse``` as used by [plaintext benchmark](https://github.com/aspnet/benchmarks/blob/master/src/Benchmarks/PlaintextMiddleware.cs#L29) is an anti-pattern:

```httpContext.Response.ContentLength = _helloWorldPayload.Length;```

Which feels a bit wrong.

This is because it goes  ContentLength->new HeaderDictionary->ParsingHelpers.SetContentLength->value.ToString()->IDictionary[key]=

[...]

PR #757, opened 2017-01-17
[Perf] Add ContentLength to IHeaderDictionary

Tratcher:
#407 Headers are stored as strings, but multiple components may need to examine the Content-Length header value on a request or response. To avoid redundant conversions allow both representations to be stored. The IHeaderDictionary implementations are responsible for keeping the representations in sync. I did not attempt any optimizations in the default implementations, they are primarily used for testing. The real gains will come from the WebListener and Kestrel implementations.

Do not merge this until the WebListener, Kestrel, and Hosting implementations are ready.

@benaadams Does this give you enough of an abstraction to implement in Kestrel? Do you want to take the first stab at it? The current xproj vs csproj schism may complicate matters.

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Features/IHeaderDictionary.cs
+++ b/src/Microsoft.AspNetCore.Http.Features/IHeaderDictionary.cs
@@ -17,5 +17,10 @@ namespace Microsoft.AspNetCore.Http
         /// <param name="key"></param>
         /// <returns>The stored value, or StringValues.Empty if the key is not present.</returns>
         new StringValues this[string key] { get; set; }
+
+        /// <summary>
+        /// Strongly typed access to the Content-Length header. Implementations must keep this in sync with the string representation.
+        /// </summary>
+        long? ContentLength { get; set; }
     }
 }
--- a/src/Microsoft.AspNetCore.Http/HeaderDictionary.cs
+++ b/src/Microsoft.AspNetCore.Http/HeaderDictionary.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.Http
 {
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNetCore.Http.Features/IHeaderDictionary.cs
```diff
@@ -17,5 +17,10 @@ public interface IHeaderDictionary : IDictionary<string, StringValues>
         /// <param name="key"></param>
         /// <returns>The stored value, or StringValues.Empty if the key is not present.</returns>
         new StringValues this[string key] { get; set; }
+
+        /// <summary>
+        /// Strongly typed access to the Content-Type header. Implementations must keep this in sync with the string representation.
```
campersau: Content-Length

src/Microsoft.AspNetCore.Http/HeaderDictionary.cs
```diff
@@ -97,6 +98,34 @@ public HeaderDictionary(int capacity)
             set { this[key] = value; }
         }
 
+        public long? ContentLength
+        {
+            get
+            {
+                long value;
+                var rawValue = this[HeaderNames.ContentLength];
```
JunTaoLuo: So the value in the string is being parsed every time from the string?

Tratcher: Yes, this is the same as before. Kestrel and WebListener will provide more optimized versions of IHeaderDictionary. I didn't optimize this one because it's primarily used in tests.

JunTaoLuo: Ah yes, I should have read your fir comment more carefully before looking at the code.

src/Microsoft.AspNetCore.Http/HeaderDictionary.cs
```diff
@@ -97,6 +98,34 @@ public HeaderDictionary(int capacity)
             set { this[key] = value; }
         }
 
+        public long? ContentLength
+        {
+            get
```
JunTaoLuo: Looks like duplication here and in `DictionaryStringValuesWrapper`. Should we keep this in a helper method like before and just call it from the getter and setter of the `ContentLength`  properties.

Tratcher: I thought about it but the implementation is actually different, their internal stores work a little differently.

benaadams:
@Tratcher looks good; will have a shot

benaadams:
Probably should be `ulong` but nevermind 😄

halter73:
I think `ulong` is a good idea. I mind.

benaadams:
Is already exposed as `long?` on Request and Response, assume changing them to `ulong?` would be a breaking change?

halter73:
Hadn't considered that. I guess we should stay consistent then. I think 63 usable bits should be enough. We'll just need to keep guarding against negative values.

benaadams:
`HeaderUtilities.TryParseInt64` will return false for a value with a sign in it, so shouldn't get negative values

halter73:
Interesting that HeaderUtilities.FormatInt64 will create a string for a negative number even though HeaderUtilities.TryParseInt64 would throw trying to parse it.

PR #773, opened 2017-02-11
Added support for middleware activation via IMiddlewareFactory

### LINKED ISSUE #754: Add support for IMiddlewareFactory and IMiddleware
The `IMiddlewareFactory` is a new extensibility point for middleware activation that would be registered with the application's  container. The default `IMiddlewareFactory` should be registered as a scoped service, and (maybe) only resolve `middlewareType`s that are already registered in the container.

Hosting will probably need to add the default `IMiddlewareFactory` to the container, but the implementation can stay in HttpAbstractions.

```c#
namespace Microsoft.AspNetCore.Http
{
    /// <summary>
    /// Provides methods to create middlware.
    /// </summary>
    public interface IMiddlewareFactory
    {
        /// <summary>
        /// Creates a middleware instance for each request.
        /// </summary>
        /// <param name="middlewareType">The concrete <see cref="Type"/> of the <see cref="IMiddleware"/>.</param>
        /// <returns>The <see cref="IMiddleware"/> instance.</returns>
        IMiddleware Create(Type middlewareType);

        /// <summary>
        /// Releases a <see cref="IMiddleware"/> instance at the end of each request.
        /// </summary>
        /// <param name="middleware">The <see cref="IMiddleware"/> instance to release.</param>
        void Release(IMiddleware middleware);
    }

    /// <summary>
    /// Defines middleware that can be added to the application's request pipeline.
    /// </summary>
    public interface IMiddleware
    {
        /// <summary>
        /// Request handling method.
        /// </summary>
        /// <param name="context">The <see cref="HttpContext"/> for the current request.</param>
        /// <param name="next">The delegate representing the remaining middleware in the request pipeline.</param>
        /// <returns>A <see cref="Task"/> that represents the execution of this middleware.</returns>
        Task Invoke(HttpContext context, RequestDelegate next);
    }
}
```

The typed `UseMiddleware` extension methods would check if the registered type implements the new `IMiddleware` interface. If it does, it will automatically decide to use the `IMiddlewareFactory` to resolve the `IMiddleware` instead of using the existing middleware activation logic. This will likely be implemented by registering a `Func<RequestDelegate, RequestDelegate>` that resolves the `IMiddlewareFactory` from `HttpContext.RequestServices`.

The existing middleware activation logic will remain for convention-based middleware to preserve backwards compatibility.

davidfowl:
- Fixes #754 
- IMiddlewareFactory and IMiddleware are new extensiblity points for
activating and authoring middleware. Under the covers, middleware is still
very much just a function. This just provides a nice way to get a per request
activated middleware instance that is created and released via the IMiddlewareFactory (this critical for non conforming containers, see https://github.com/dotnetjunkie/Missing-Core-DI-Extensions/blob/7efd3c3f1aa6330d303d0a3e2d28fdc43687b9f4/src/SampleApplication.SimpleInjector/Startup.cs#L52-L55 for more details).
The caveats are that middleware needs to be registered in the container (by default)
and that not possible to explicitly pass arguments directly via UseMiddleware.
- Added tests

/cc @Tratcher @halter73 @pakrym

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UseMiddlewareExtensions.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UseMiddlewareExtensions.cs
@@ -42,6 +42,18 @@ namespace Microsoft.AspNetCore.Builder
         /// <returns>The <see cref="IApplicationBuilder"/> instance.</returns>
         public static IApplicationBuilder UseMiddleware(this IApplicationBuilder app, Type middleware, params object[] args)
         {
+            if (typeof(IMiddleware).GetTypeInfo().IsAssignableFrom(middleware.GetTypeInfo()))
+            {
+                // IMiddleware doesn't support passing args directly since it's 
+                // activated from the container
+                if (args.Length > 0)
+                {
+                    throw new NotSupportedException(Resources.FormatException_UseMiddlewareExplicitArgumentsNotSupported(typeof(IMiddleware)));
+                }
+
+                return UseMiddlewareInterface(app, middleware);
+            }
+
             var applicationServices = app.ApplicationServices;
             return app.Use(next =>
             {
@@ -89,6 +101,41 @@ namespace Microsoft.AspNetCore.Builder
                     }
 
                     return factory(instance, context, serviceProvider);
+                };
+            });
+        }
+
+        private static IApplicationBuilder UseMiddlewareInterface(IApplicationBuilder app, Type middlewareType)
+        {
+            return app.Use(next =>
+            {
+                return async context =>
+                {
+                    var middlewareFactory = (IMiddlewareFactory)context.RequestServices.GetService(typeof(IMiddlewareFactory));
+
+                    if (middlewareFactory == null)
+                    {
+                        // No middleware factory
+                        throw new InvalidOperationException(Resources.FormatException_UseMiddlewareNoMiddlewareFactory(typeof(IMiddlewareFactory)));
+                    }
+
+                    var middleware = middlewareFactory.Create(middlewareType);
+
+                    if (middleware == null)
+                    {
+                        // The factory returned null, it's a broken implementation
+                        throw new InvalidOperationException(Resources.FormatException_UseMiddlewareUnableToCreateMiddleware(middlewareType));
+
+                    }
+
+                    try
+                    {
+                        await middleware.Invoke(context, next);
+                    }
+                    finally
+                    {
+                        middlewareFactory.Release(middleware);
+                    }
                 };
             });
         }
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UseMiddlewareExtensions.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UseMiddlewareExtensions.cs
@@ -112,7 +112,6 @@ namespace Microsoft.AspNetCore.Builder
                 return async context =>
                 {
                     var middlewareFactory = (IMiddlewareFactory)context.RequestServices.GetService(typeof(IMiddlewareFactory));
-
                     if (middlewareFactory == null)
                     {
                         // No middleware factory
@@ -120,12 +119,10 @@ namespace Microsoft.AspNetCore.Builder
                     }
 
                     var middleware = middlewareFactory.Create(middlewareType);
-
                     if (middleware == null)
                     {
                         // The factory returned null, it's a broken implementation
                         throw new InvalidOperationException(Resources.FormatException_UseMiddlewareUnableToCreateMiddleware(middlewareType));
-
                     }
 
                     try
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UseMiddlewareExtensions.cs
```diff
@@ -42,6 +42,18 @@ public static IApplicationBuilder UseMiddleware<TMiddleware>(this IApplicationBu
         /// <returns>The <see cref="IApplicationBuilder"/> instance.</returns>
         public static IApplicationBuilder UseMiddleware(this IApplicationBuilder app, Type middleware, params object[] args)
         {
+            if (typeof(IMiddleware).GetTypeInfo().IsAssignableFrom(middleware.GetTypeInfo()))
+            {
+                // IMiddleware doesn't support passing args directly since it's 
+                // activated from the container
+                if (args.Length > 0)
+                {
+                    throw new NotSupportedException(Resources.FormatException_UseMiddlewareExplicitArgumentsNotSupported(typeof(IMiddleware)));
```
halter73: So you've decide you're OK with throwing? https://github.com/aspnet/HttpAbstractions/issues/754#issuecomment-279120347

src/Microsoft.AspNetCore.Http.Abstractions/Resources.resx
```diff
@@ -144,4 +144,13 @@
   <data name="Exception_PortMustBeGreaterThanZero" xml:space="preserve">
     <value>The value must be greater than zero.</value>
   </data>
+  <data name="Exception_UseMiddlewareNoMiddlewareFactory" xml:space="preserve">
+    <value>No service for type '{0}' has been registered.</value>
+  </data>
+  <data name="Exception_UseMiddlewareUnableToCreateMiddleware" xml:space="preserve">
+    <value>Unable to create middleware '{0}'.</value>
```
halter73: This could be a little more clear about it being the IMiddlewareFactory that's not providing an IMiddleware implementation of type {0}.

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UseMiddlewareExtensions.cs
```diff
@@ -93,6 +105,38 @@ public static IApplicationBuilder UseMiddleware(this IApplicationBuilder app, Ty
...
+                return async context =>
+                {
+                    var middlewareFactory = (IMiddlewareFactory)context.RequestServices.GetService(typeof(IMiddlewareFactory));
+                    if (middlewareFactory == null)
+                    {
+                        // No middleware factory
+                        throw new InvalidOperationException(Resources.FormatException_UseMiddlewareNoMiddlewareFactory(typeof(IMiddlewareFactory)));
+                    }
+
+                    var middleware = middlewareFactory.Create(middlewareType);
```
halter73: It's kind of interesting how the type winds up just being used as a key and nothing else.

src/Microsoft.AspNetCore.Http/MiddlewareFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Http
+{
+    public class MiddlewareFactory : IMiddlewareFactory
+    {
+        // The default middleware factory is just an IServiceProvider proxy.
+        // This should be registered as a scoped service so that the middleware instances
+        // don't end up being singletons.
```
halter73: Is there a companion PR in hosting where this is registered?

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UseMiddlewareExtensions.cs
davidfowl: I put it in the PR for feedback. I considered adding another overload but it would be confusing naming.

src/Microsoft.AspNetCore.Http/MiddlewareFactory.cs
davidfowl: Not yet, I'll do that after we agree we like this.

davidfowl:
🆙 📅

Tratcher:
What does the perf look like for middleware with the various lifetimes? E.g. singleton vs scoped.

davidfowl:
What exactly are you asking? What he cost of resolving middleware instances per request is? What will that data change?

Tratcher:
If there's a high overhead to resolve middleware per request then it will be hard to recommend this to anyone.

davidfowl:
Please read the PR description outlining the reason we're adding this in the first place (non conforming containers). On top of that, mvc controllers are created once per request so are many other objects along the request pipeline. The DI container is also extremely efficient at creating new objects so I'd be extremely suprised if this causes any real problems. People have to do this today if they anyway if they want to use a non conforming container (see the code sample in the PR description).

Tratcher:
Were you planning to convert any of our middleware?

davidfowl:
Nope. Although maybe we should use it somewhere so we can have something validating this code path.

halter73:
:shipit:

PR #779, opened 2017-02-16
Reducing string allocation in Base64UrlTextEncoder's Encode path

kichalla:


src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
```diff
@@ -32,38 +32,30 @@ public static byte[] Decode(string text)
...
+                    {
+                        destination[i] = '-';
+                    }
+                    else if (destination[i] == '/')
+                    {
+                        destination[i] = '_';
+                    }
+                    else if (destination[i] == '=')
+                    {
+                        destination[i] = '.';
```
cesarbs: Why this change?

test/Microsoft.AspNetCore.WebUtilities.Tests/Base64UrlTextEncoderTests.cs
```diff
@@ -65,5 +70,26 @@ public void DecodeToBase64String_ReturnsValid_Base64String(string text, string e
...
+                {
+                    data[index] = (byte)(5 + length + (index * 23));
+                }
+                string text = Base64UrlTextEncoder.Encode(data);
+
+                // Act
+                var urlEncodedText = UrlEncoder.Default.Encode(text);
+
+                // Assert
+                Assert.Equal(text, urlEncodedText);
```
cesarbs: `Assert.NotSame`?

src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
kichalla: This is because `=` character gets url encoded and also if we were to have similar behavior like earlier we would need to allocate a new string (like InplaceStringBuilder did earlier internally)

test/Microsoft.AspNetCore.WebUtilities.Tests/Base64UrlTextEncoderTests.cs
kichalla: Do we care here? UrlEncoder might choose to do no-op since it couldn't find any characters to encode. We only want to make sure that characters remain the same.

cesarbs: I see.

Tratcher:
Perf data?

src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
```diff
@@ -32,38 +32,30 @@ public static byte[] Decode(string text)
             return Convert.FromBase64String(DecodeToBase64String(text));
         }
 
+        // Since the string instnace passed here is owned by the Encode method, its safe to modify
```
davidfowl: nit: instance

test/Microsoft.AspNetCore.WebUtilities.Tests/Base64UrlTextEncoderTests.cs
```diff
@@ -32,30 +33,34 @@ public void DataOfVariousLengthRoundTripCorrectly()
         [InlineData("", "")]
         [InlineData("+", "-")]
         [InlineData("/", "_")]
-        [InlineData("=", "")]
-        [InlineData("==", "")]
-        [InlineData("a+b+c+==", "a-b-c-")]
-        [InlineData("a/b/c==", "a_b_c")]
-        [InlineData("a+b/c==", "a-b_c")]
+        [InlineData("=", ".")]
+        [InlineData("==", "..")]
```
Tratcher: This may be problematic, .. has special meaning in url paths.

Tratcher: so does `.` actually

src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs
```diff
@@ -87,17 +78,16 @@ internal static string DecodeToBase64String(string text)
                 {
                     inplaceStringBuilder.Append('/');
                 }
+                else if (text[i] == '.')
```
Tratcher: You would only want to attempt this at the end of the string.

test/Microsoft.AspNetCore.WebUtilities.Tests/Base64UrlTextEncoderTests.cs
Tratcher: let's find a different replacement char

kichalla: Ah OK

kichalla:
Discussed offline with @Tratcher and @rynowak. Closing this PR as it's not worth the risk we might be taking.

PR #780, opened 2017-02-17
Remove allocations and improve Header Quality Values parsing performance

Yves57:
See #700 and #709 (conversation only).
New PR because I have totally broken my previous PR, like every time I touch Git :-(


test/Microsoft.Net.Http.Headers.Tests/StringWithQualityHeaderValueTest.cs
```diff
@@ -124,6 +124,9 @@ public void Parse_SetOfValidValueStrings_ParsedCorrectly()
             CheckValidParse("t ;  q  =   0", new StringWithQualityHeaderValue("t", 0));
             CheckValidParse("iso-8859-5", new StringWithQualityHeaderValue("iso-8859-5"));
             CheckValidParse("unicode-1-1; q=0.8", new StringWithQualityHeaderValue("unicode-1-1", 0.8));
+            CheckValidParse("quality_limit; q=000000001", new StringWithQualityHeaderValue("quality_limit", 1));
+            CheckValidParse("quality_limit; q=0000001.0", new StringWithQualityHeaderValue("quality_limit", 1));
+            CheckValidParse("quality_limit; q=1.0000000", new StringWithQualityHeaderValue("quality_limit", 1));
```
JunTaoLuo: Something like q=1.0X should still be parsed as 1 where X is any invalid character.

test/Microsoft.Net.Http.Headers.Tests/StringWithQualityHeaderValueTest.cs
```diff
@@ -124,6 +124,9 @@ public void Parse_SetOfValidValueStrings_ParsedCorrectly()
             CheckValidParse("t ;  q  =   0", new StringWithQualityHeaderValue("t", 0));
             CheckValidParse("iso-8859-5", new StringWithQualityHeaderValue("iso-8859-5"));
             CheckValidParse("unicode-1-1; q=0.8", new StringWithQualityHeaderValue("unicode-1-1", 0.8));
+            CheckValidParse("quality_limit; q=000000001", new StringWithQualityHeaderValue("quality_limit", 1));
```
JunTaoLuo: This isn't a valid format per spec.

test/Microsoft.Net.Http.Headers.Tests/StringWithQualityHeaderValueTest.cs
```diff
@@ -124,6 +124,9 @@ public void Parse_SetOfValidValueStrings_ParsedCorrectly()
...
+            CheckValidParse("quality_limit; q=0000001.0", new StringWithQualityHeaderValue("quality_limit", 1));
```
JunTaoLuo: This isn't a valid format per spec.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -480,6 +479,100 @@ public static unsafe bool TryParseInt64(StringSegment value, out long result)
...
+            {
+                ch = input[current];
+
+                if (current >= limit)
+                {
+                    return false;
+                }
+                else if (ch >= '0' && ch <= '9')
+                {
+                    intPart = intPart * 10 + ch - '0';
```
JunTaoLuo: Similar to the previous comment, these are not valid doubles so parsing should fail.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -480,6 +479,100 @@ public static unsafe bool TryParseInt64(StringSegment value, out long result)
             }
         }
 
+        // Fast RFC2616 3.9 Quality value parser (and without memory allocation)
+        internal static bool TryParseQualityDouble(string input, int startIndex, out double quality, out int length)
```
JunTaoLuo: We should optimize the opposite direction too, something like `string FormatQualityDouble(double quality)`

Yves57: I agree, but I prefer to close this PR first.

test/Microsoft.Net.Http.Headers.Tests/StringWithQualityHeaderValueTest.cs
Yves57: In fact `TryParseQualityDouble()` accepts it but it is rejected [later](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.Net.Http.Headers/BaseHeaderParser.cs#L60).

src/Microsoft.Net.Http.Headers/StringWithQualityHeaderValue.cs
```diff
@@ -204,16 +204,9 @@ private static bool TryReadQuality(string input, StringWithQualityHeaderValue re
...
-            if (qualityLength == 0)
-            {
-                return false;
-            }
-
+            var qualityLength = 0;
             double quality;
-            if (!double.TryParse(input.Substring(current, qualityLength), NumberStyles.AllowDecimalPoint,
-                NumberFormatInfo.InvariantInfo, out quality))
+            if (!HeaderUtilities.TryParseQualityDouble(input, current, out quality, out qualityLength))
```
Yves57: Updated

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -13,6 +13,7 @@ namespace Microsoft.Net.Http.Headers
     public static class HeaderUtilities
     {
         private static readonly int _int64MaxStringLength = 20;
+        private static readonly int _qualityValueMaxCharCount = 5;      // See RFC7231 5.3.1
```
JunTaoLuo: I was unsure about being this strict in my last comment. I spoke with @Tratcher and we figured it's probably better to be a little lenient here so let's set this value to 10.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -480,6 +480,89 @@ public static unsafe bool TryParseInt64(StringSegment value, out long result)
...
+                }
+
+                if (ch == '.')
+                {
+                    current++;
+
+                    while (current < inputLength)
+                    {
+                        ch = input[current];
+                        if (ch >= '0' && ch <= '9')
```
JunTaoLuo: This is still going to run for parsing `1.111` for example. Although this is redundant work since any non zero digit after a `1.` it's not a big cost so it's fine as is. 

However, let's add a check at the end before returning to make sure the value is between 0 and 1 inclusive since this helper method parses quality values explicitly and is not a generic double parser. The bounds checking can then be removed here: https://github.com/aspnet/HttpAbstractions/blob/87cd79d6fc54bb4abf07c1e380cd7a9498a78612/src/Microsoft.Net.Http.Headers/StringWithQualityHeaderValue.cs#L221-L224

Yves57: Added the value range check in the method. But it changes the behavior of `MediaTypeHeaderValue.GetQuality()` (any value was accepted before). I find this better, but I just advise you.

JunTaoLuo:
Merged in https://github.com/aspnet/HttpAbstractions/commit/7440d5d29c3814d94c83cdbbea04ee8179e03e64. Thanks @Yves57 for the contribution!

PR #791, opened 2017-03-13
Add Owin extension for convenience

kchanlee:
`IApplicationBuilder` has a very convenient extension that we can add a middleware like
```C#
app.Use((context, next) =>
{
    // do something
    return next();
});
```
rather than
```C#
app.Use(next =>
{
    return context =>
    {
        // do something
        return next(context);
    }
});
```
The former one is much more straightforward and easier to understand and to use.

Asp.Net Owin (Project [Katana](https://github.com/aspnet/AspNetKatana/blob/dev/src/Microsoft.Owin/Extensions/AppBuilderUseExtensions.cs)) also has a similar extension, but Asp.Net Core Owin doesn't. Therefore, I propose to add a similar extension to Asp.Net Core Owin so that we can align them and write code easier like:
```C#
app.UseOwinEx(addToPipeline =>
{
    addToPipeline((env, next) =>
    {
        // do something
        return next();
    });
});
```

Here is the pull request including test case.

### DIFF
--- a/src/Microsoft.AspNetCore.Owin/OwinExtensions.cs
+++ b/src/Microsoft.AspNetCore.Owin/OwinExtensions.cs
@@ -84,6 +84,16 @@ namespace Microsoft.AspNetCore.Builder
 
         public static IApplicationBuilder UseOwin(this IApplicationBuilder builder, Action<AddMiddleware> pipeline)
         {
+            return UseOwinHelp(builder, pipeline, b => b.UseOwin());
+        }
+
+        public static IApplicationBuilder UseOwinEx(this IApplicationBuilder builder, Action<AddMiddlewareEx> pipeline)
+        {
+            return UseOwinHelp(builder, pipeline, b => b.UseOwinEx());
+        }
+
+        private static IApplicationBuilder UseOwinHelp<T>(this IApplicationBuilder builder, Action<T> pipeline, Func<IApplicationBuilder, T> useOwin)
+        {
             if (builder == null)
             {
                 throw new ArgumentNullException(nameof(builder));
@@ -93,7 +103,7 @@ namespace Microsoft.AspNetCore.Builder
                 throw new ArgumentNullException(nameof(pipeline));
             }
 
-            pipeline(builder.UseOwin());
+            pipeline(useOwin(builder));
             return builder;
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Owin/OwinExtensions.cs
```diff
@@ -63,8 +57,43 @@ public static AddMiddleware UseOwin(this IApplicationBuilder builder)
             return add;
         }
 
+        public static AddMiddlewareEx UseOwinEx(this IApplicationBuilder builder)
```
Tratcher: Don't like the Ex name.

test/Microsoft.AspNetCore.Owin.Tests/OwinExtensionTests.cs
```diff
@@ -10,18 +10,36 @@
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Xunit;
+using System.Net.WebSockets;
```
Tratcher: Sort system first

src/Microsoft.AspNetCore.Owin/OwinExtensions.cs
davidfowl: Win32 ftw

kchanlee: I don't like it either, but I didn't come up with a better name... What's your suggestion?

test/Microsoft.AspNetCore.Owin.Tests/OwinExtensionTests.cs
kchanlee: OK, I will update it.

BTW, is there any mandatory code style check in the build system so that this kind of code style inconsistency will cause build error (or at least build warning)?

src/Microsoft.AspNetCore.Owin/OwinExtensions.cs
davidfowl: /cc @Eilon any ideas?

test/Microsoft.AspNetCore.Owin.Tests/OwinExtensionTests.cs
Tratcher: Stylecop used to do that. I think you could do it now with roslyn analysers.

muratg:
@kchanlee Thanks for the PR. Closing it as it became a stale one.  Please file a bug to bring the discussion back on track if you'd still like to pursue this. Thank you.

PR #797, opened 2017-03-21
Use a CopyOnWriteDictionary in ApplicationBuilder...

### LINKED ISSUE #783: new ApplicationBuilder(ApplicationBuilder builder) should copy the Properties dictionary
As I was trying to use `Map` and set a custom service provider `ApplicationServices` within the branch, I found that setting `ApplicationServices` on the branched application was also setting it on the application builder that was branched from, due to both application builders sharing the same properties dictionary. This feels counter-intuitive -- I expected that a new `IApplicationBuilder` in a new branch would have its own properties.

tuespetre:
...so branches can set their own properties

Fixes #783


muratg:
@tuespetre Thanks for the contribution!

PR #798, opened 2017-03-21
Auth 2.0: Add Auth Abstractions, Core

HaoK:
@Tratcher @davidfowl @ajcvickers 

Should I start to obsolete the old Http.Authentication at the same time?

Part of https://github.com/aspnet/Security/pull/1151


src/Microsoft.AspNetCore.Authentication.Abstractions/ISecureDataFormat.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public interface ISecureDataFormat<TData>
```
HaoK: The Data interfaces are kinda specific to our Security repo implementations, these could live inside of the Security auth package instead...  Thoughts @Tratcher ?

Tratcher:
They won't work anymore right? Then yes, start Obsoleting them.

src/Microsoft.AspNetCore.Authentication.Abstractions/ISecureDataFormat.cs
Tratcher: It's only used on Options types, right? Then it's fine to move to the concrete package.

src/Microsoft.AspNetCore.Authentication.Core/AuthAppBuilderExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// Extension methods to add authentication capabilities to an HTTP application pipeline.
+    /// </summary>
+    public static class AuthAppBuilderExtensions
+    {
+        /// <summary>
+        /// Adds the <see cref="AuthenticationMiddleware"/> middleware to the specified <see cref="IApplicationBuilder"/>, which enables authentication capabilities.
+        /// </summary>
+        /// <param name="app">The <see cref="IApplicationBuilder"/> to add the middleware to.</param>
+        /// <returns>A reference to this instance after the operation has completed.</returns>
+        public static IApplicationBuilder UseAuthentication(this IApplicationBuilder app)
```
Tratcher: Why did Auth.Core have to move to HttpAbstractions?

HaoK: Do we want to be able to test auth without the security infrastructure, i.e. IISIntegration would like to add auth but not depend on Security

HaoK: So I moved the 'Core' pieces that are needed for the auth infrastructure service here.  But the scheme specific helpers and base classes are still in Security.

So you can add schemes, and call into auth via what's in HttpAbstractions, but all the sugar around implementing an auth middleware lives in the Security PR (still to be recreated)

src/Microsoft.AspNetCore.Authentication.Abstractions/ISecureDataFormat.cs
PinpointTownes: :+1: for moving the "data handling" back to the security repo :smile:

PinpointTownes:
Now that an important part of the new security stuff is back in this repo, is there a particular reason the authentication methods (`AuthenticateAsync`/`ChallengeAsync` and co.) couldn't be first-class citizen and added directly in the `Http.Abstractions` package, like what we used to have in 1.0.0 with `IAuthenticationManager`?

HaoK:
@PinpointTownes extensions on HttpContext aren't sufficient?

HaoK:
Altho there is a bit of wierdness in the new stuff that I don't like today, that you have to call AddAuthenticationCore for context.Authenticate to not blow up...

PinpointTownes:
> @PinpointTownes extensions on HttpContext aren't sufficient?

They are. I'm just wondering if there's a technical reason for opting for extensions in this case :smile:

HaoK:
@PinpointTownes did you have an alternative in mind?  We expect this to be mostly called by infrastructure/framework code like Security repo stuff/Identity/Authorize code.

HaoK:
I went with extensions mostly to keep roughly the same shape as the old context.Authentication.Authenticate look and feel.  As it made updating the existing usages very easy in a Find/Replace.

PinpointTownes:
> @PinpointTownes did you have an alternative in mind?

I suppose you could keep the existing `IAuthenticationManager` as-is and simply update `DefaultAuthenticationManager` to use your `IAuthenticationService` (resolved via `HttpContext.RequestServices`)... or better, you could merge them.

It would have a pro: you wouldn't have to deprecate the existing abstractions.

HaoK:
We can't touch the existing old interfaces as that's the interop story for the old 1.X packages still continuing to work in a new MVC.  They will still work since they are using the existing unchanged IAuthenticationHandler and the old packages are still running the old code

PinpointTownes:
> We expect this to be mostly called by infrastructure/framework code like Security repo stuff/Identity/Authorize code.

+ MVC, since it relies on these methods for `SignInResult`/`SignIn` and co.

HaoK:
Yeah...so basically the main consumer of the new APIs is myself :)

PinpointTownes:
> We can't touch the existing old interfaces as that's the interop story for the old 1.X packages still continuing to work in a new MVC.

Hum my bad, I thought these things were supposed to be removed in 2.0.

It's a bit strange tho'... as there's no guarantee code written for 1.0 will still compile with 2.0 (that's even the opposite, otherwise you wouldn't increase the major version, right?)

HaoK:
Yeah so clearly 1.0 code won't compile, but 1.0 packages should still work/run in a 2.0 app as there's binary compat.  Hence we can't really touch the HttpAbstractions auth behavior if we want those to still work...  They can continue to party there like its 2016

PinpointTownes:
> but 1.0 packages should still work/run in a 2.0 app as there's binary compat.

Really? So how do you explain things like https://github.com/openiddict/openiddict-core/issues/359 (code written for EF Core 1.0 no longer works with 1.2/2.0 due to a signature change) or https://github.com/aspnet/SignalR/issues/155 ?

HaoK:
So, my guess is the reality will be more complicated yes :)

HaoK:
Obsoleted old Auth APIs

HaoK:
@davidfowl How do things look here?  I'd like to start getting the additive changes in this week (Auth.Abstractions + Auth.Core).  There's still one change that I'm not sure about, which is that Auth is now an optional service.  A call to `services.AddAuthenticationCore()` is required for `context.AuthenticateAsync` to work.  This is unlike the old HttpAbstractions.Authentication which always supported `context.Authentication.Authenticate`.

We could have hosting call AddAuthenticationCore similar to AddOptions, otherwise this is something that needs to be called when the host is being built, see https://github.com/aspnet/HttpSysServer/pull/325/commits/c57b98b1908a60fbbc35a12498eb93da4b372cb5#diff-2d7f0bf656ca8cab31fc9edcd837ee98R62

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationHttpContextExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+        public static Task SignInAsync(this HttpContext context, ClaimsPrincipal principal) =>
+            context.SignInAsync(scheme: null, principal: principal, properties: null);
+
+        public static Task SignInAsync(this HttpContext context, ClaimsPrincipal principal, AuthenticationProperties properties) =>
+            context.SignInAsync(scheme: null, principal: principal, properties: properties);
+
+        public static Task SignInAsync(this HttpContext context, string scheme, ClaimsPrincipal principal, AuthenticationProperties properties) =>
+            context.RequestServices.GetRequiredService<IAuthenticationService>().SignInAsync(context, scheme, principal, properties);
+
+
```
Tratcher: nit: extra line

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace Microsoft.AspNetCore.Authentication
+{
+    public class AuthenticationOptions
+    {
+        private readonly IList<AuthenticationSchemeBuilder> _schemes = new List<AuthenticationSchemeBuilder>();
+
+        /// <summary>
+        /// Returns the schemes in the order they were added (important for request handling priority)
+        /// </summary>
+        public IEnumerable<AuthenticationSchemeBuilder> Schemes
```
Tratcher: syntax
Schemes => _schemes;

src/Microsoft.AspNetCore.Http.Features/Authentication/SignInContext.cs
```diff
@@ -7,6 +7,7 @@
 
 namespace Microsoft.AspNetCore.Http.Features.Authentication
 {
+    [Obsolete("See https://go.microsoft.com/fwlink/?linkid=845470")]
```
Tratcher: I thought you weren't going to check these Obsolete's in until after you got all of the other repos converted?

Tratcher:
I think it's OK for AuthenticateAsync to not work until you add an auth middleware / services. We just need to make sure the caller (the AuthZ filter) gives you a good error message.

src/Microsoft.AspNetCore.Http.Features/Authentication/SignInContext.cs
HaoK: When I smush all the changes together for the merge, I won't check in anything outside of the 2 new assemblies, so you don't need to review anything outside of the Authentication packages in this PR

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationTicket.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// </summary>
+    public class AuthenticationTicket
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AuthenticationTicket"/> class
+        /// </summary>
+        /// <param name="principal">the <see cref="ClaimsPrincipal"/> that represents the authenticated user.</param>
+        /// <param name="properties">additional properties that can be consumed by the user or runtime.</param>
+        /// <param name="authenticationScheme">the authentication middleware that was responsible for this ticket.</param>
+        public AuthenticationTicket(ClaimsPrincipal principal, AuthenticationProperties properties, string authenticationScheme)
```
Tratcher: We're inconsistent about the type of things called authentication scheme. In some places it's a AuthenticationScheme, and in others it's just a string. Should the strings be called authenticationSchemeName? DefaultAuthenticationSchemeName? Or is that just overkill?

src/Microsoft.AspNetCore.Authentication.Abstractions/IAuthenticationSchemeProvider.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public interface IAuthenticationSchemeProvider
+    {
+        Task<IEnumerable<AuthenticationScheme>> GetAllSchemesAsync();
```
Tratcher: Time to fill in the comments?

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -0,0 +1,132 @@
...
+            if (_map.ContainsKey(name))
+            {
+                return Task.FromResult(_map[name]);
+            }
+            return Task.FromResult<AuthenticationScheme>(null);
+        }
+
+        public Task<IEnumerable<AuthenticationScheme>> GetRequestHandlerSchemesAsync(PathString requestPath)
+        {
+            return Task.FromResult<IEnumerable<AuthenticationScheme>>(_requestHandlers);
```
Tratcher: This isn't using the path, and it's not virtual. What's the plan?

src/Microsoft.AspNetCore.Authentication.Abstractions/Microsoft.AspNetCore.Authentication.Abstractions.csproj
```diff
@@ -0,0 +1,20 @@
...
+  <PropertyGroup>
+    <Description>ASP.NET Core common types used by the various authentication components.</Description>
+    <VersionPrefix>1.2.0</VersionPrefix>
+    <TargetFrameworks>net451;netstandard1.3</TargetFrameworks>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>aspnetcore;authentication;security</PackageTags>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="1.2.0-*" />
```
Tratcher: Versions go in dependencies.props

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationHandlerProvider.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        public AuthenticationHandlerProvider(IAuthenticationSchemeProvider schemes)
+        {
+            Schemes = schemes;
+        }
+
+        public IAuthenticationSchemeProvider Schemes { get; }
+
+        // handler instance cache, need to initialize once per request
+        private Dictionary<string, IAuthenticationHandler> _handlerMap = new Dictionary<string, IAuthenticationHandler>();
```
Tratcher: Ordinal

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -0,0 +1,132 @@
...
+            {
+                var scheme = builder.Build();
+                AddScheme(scheme);
+            }
+        }
+
+        private readonly AuthenticationOptions _options;
+        private readonly object _lock = new object();
+
+        private IDictionary<string, AuthenticationScheme> _map = new Dictionary<string, AuthenticationScheme>(StringComparer.Ordinal); // case sensitive?
```
Tratcher: Ordinal

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
```diff
@@ -0,0 +1,133 @@
...
+                throw new ArgumentNullException(nameof(principal));
+            }
+
+            if (authenticationScheme == null)
+            {
+                var defaultScheme = await Schemes.GetDefaultSignInSchemeAsync();
+                authenticationScheme = defaultScheme?.Name;
+                if (authenticationScheme == null)
+                {
+                    throw new InvalidOperationException($"No authenticationScheme was specified, and there was no DefaultAuthenticateScheme found.");
```
Tratcher: DefaultSignInScheme

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        public virtual async Task SignOutAsync(HttpContext httpContext, string authenticationScheme, AuthenticationProperties properties)
+        {
+            if (string.IsNullOrEmpty(authenticationScheme))
+            {
+                throw new ArgumentException(nameof(authenticationScheme));
+            }
+
+            var handler = await Handlers.GetHandlerAsync(httpContext, authenticationScheme);
```
Tratcher: default to sign-in scheme? Or define a default like the others?

src/Microsoft.AspNetCore.Authentication.Core/Microsoft.AspNetCore.Authentication.Core.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <Description>ASP.NET Core common types used by the various authentication middleware components.</Description>
+    <VersionPrefix>1.2.0</VersionPrefix>
```
Tratcher: extract versions

src/Microsoft.AspNetCore.Authentication.Core/Microsoft.AspNetCore.Authentication.Core.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <TargetFrameworks>net451;netstandard1.3</TargetFrameworks>
```
Tratcher: retarget

test/Microsoft.AspNetCore.Authentication.Core.Test/AuthenticationMiddlewareTests.cs
```diff
@@ -0,0 +1,181 @@
...
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.TestHost;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public class AuthenticationMiddlewareTests
```
Tratcher: AuthenticationMiddleware moved back to Security

test/Microsoft.AspNetCore.Authentication.Core.Test/TestExtensions.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using System.Xml;
+using System.Xml.Linq;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.TestHost;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public static class TestExtensions
```
Tratcher: Move back to Security?

test/Microsoft.AspNetCore.Authentication.Core.Test/TokenExtensionTests.cs
```diff
@@ -0,0 +1,181 @@
...
+            Assert.False(props.UpdateTokenValue("ONE", ".11"));
+            Assert.False(props.UpdateTokenValue("Jigglypuff", ".11"));
+
+            Assert.Null(props.GetTokenValue("ONE"));
+            Assert.Null(props.GetTokenValue("Jigglypuff"));
+            Assert.Equal(3, props.GetTokens().Count());
+
+        }
+
+        //public class TestAuthHandler : IAuthenticationHandler
```
Tratcher: ?

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
```diff
@@ -0,0 +1,69 @@
...
+                throw new InvalidOperationException("Scheme already exists: " + name);
+            }
+
+            var builder = new AuthenticationSchemeBuilder(name);
+            configureBuilder(builder);
+            _schemes.Add(builder);
+            SchemeMap[name] = builder;
+        }
+
+        public void ConfigureScheme(string name, Action<AuthenticationSchemeBuilder> configureBuilder)
```
HaoK: Yeah we don't need this anymore, it was needed before we had named options, as it was used to configure the options that lived in the scheme.settings.  Good catch I'll remove this

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationTicket.cs
HaoK: Well we called it authenticationScheme historically, so I think that should be familiar enough, do you think its confusing currently?  I think the type is enough to differentiate...

src/Microsoft.AspNetCore.Authentication.Abstractions/IAuthenticationSchemeProvider.cs
HaoK: Yeah, I'll fill in the comments today

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
HaoK: I'll remove requestPath for now, it was a leftover when we were trying to do the ghetto routing.  We can come back to this if we want later.

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
HaoK: I rebased, so I lost the context for this one, which method is lacking, SignIn should have a default now, are you referring to SignOut?

Tratcher: I was referring to SignOut. Should it default to SignInScheme, or should it have its own default?

HaoK: I don't really feel strongly either way, defaulting to SignIn seems reasonable, I guess the other way is to ask does it ever make sense for the defaults to be different??

HaoK:
Updated all doc comments, removed Forbid from IAuthenticationService, and addressed PR feedback, I think I got them all, but @Tratcher you should double check the important ones, since I fixed some that you noted in the Security PR.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
```diff
@@ -44,26 +46,19 @@ public void AddScheme(string name, Action<AuthenticationSchemeBuilder> configure
...
-            if (!SchemeMap.ContainsKey(name))
-            {
-                throw new InvalidOperationException("Scheme does not exists: " + name);
-            }
-
-            configureBuilder(SchemeMap[name]);
-        }
-
+        /// <summary>
+        /// Used by as the default scheme by <see cref="IAuthenticationService.AuthenticateAsync(Http.HttpContext, string)"/>.
```
Tratcher: Http.HttpContext?

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
Tratcher: If you're just using forms, social, or WsFed then there's one thing to sign in and out of (cookies).
If you're using Identity then there are several signouts and the app signs out of them all.
If you're using OIDC there are two signouts, and the app needs to choose which.
And if you're using Bearer then signout doesn't work anyways.

So having a default sign out scheme, or even defaulting to the sign in scheme would only help the first scenario which is less common. Forget I asked.

src/Microsoft.AspNetCore.Authentication.Core/Microsoft.AspNetCore.Authentication.Core.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>aspnetcore;authentication;security</PackageTags>
+    <EnableApiCheck>false</EnableApiCheck>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.AspNetCore.Authentication.Abstractions\Microsoft.AspNetCore.Authentication.Abstractions.csproj" />
+    <ProjectReference Include="..\Microsoft.AspNetCore.Http\Microsoft.AspNetCore.Http.csproj" />
+    <ProjectReference Include="..\Microsoft.AspNetCore.Http.Extensions\Microsoft.AspNetCore.Http.Extensions.csproj" />
+    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="1.2.0-*" PrivateAssets="All" />
```
Tratcher: Extract versions

test/Microsoft.AspNetCore.Authentication.Core.Test/Microsoft.AspNetCore.Authentication.Core.Test.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp1.1;net46</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netcoreapp1.1</TargetFrameworks>
+  </PropertyGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Authentication.Core\Microsoft.AspNetCore.Authentication.Core.csproj" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0-*" />
```
Tratcher: Extract versions

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
HaoK: Feels a little bit weird to add usings for doc comments, but whatever :)

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
```diff
@@ -28,7 +37,14 @@ public AuthenticationScheme(string name, Type handlerType)
         }
 
         // TODO: add display name?
```
HaoK: Yeah this one is already tracked so I can just remove

HaoK:
Good enough for initial merge @Tratcher ?

Tratcher (approved):

HaoK:
Merged 13925be91e0336a37e782af05a96507bb676122d the new packages, will open a new PR with obsolete's which will be checked in at the same time we switch all of security over (RIP build)

HaoK:
Obsolete PR https://github.com/aspnet/HttpAbstractions/pull/806

PR #803, opened 2017-03-24
Remove net451 as a cross-compile target

pranavkm:


src/Microsoft.AspNetCore.WebUtilities/BufferedReadStream.cs
```diff
@@ -211,7 +201,17 @@ public override int Read(byte[] buffer, int offset, int count)
 
             return await _inner.ReadAsync(buffer, offset, count, cancellationToken);
         }
-#if NET451
+#if NET46
```
Tratcher: why did you move these?

PR #806, opened 2017-03-31
Obsolete Auth 1.0

HaoK:
Will go in when Security switches over to Auth 2.0 (this will require all Auth usages to switch over to 2.0 as well)

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Authentication/AuthenticateInfo.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Authentication/AuthenticateInfo.cs
@@ -9,21 +9,25 @@ namespace Microsoft.AspNetCore.Http.Authentication
     /// <summary>
     /// Used to store the results of an Authenticate call.
     /// </summary>
+    [Obsolete("See https://go.microsoft.com/fwlink/?linkid=845470")]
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNetCore.Http.Abstractions/Authentication/AuthenticateInfo.cs
```diff
@@ -9,21 +9,25 @@ namespace Microsoft.AspNetCore.Http.Authentication
     /// <summary>
     /// Used to store the results of an Authenticate call.
     /// </summary>
+    [Obsolete("See https://go.microsoft.com/fwlink/?linkid=845470")]
```
natemcmaster: This is the pattern we usually use for obsolete messages. https://github.com/aspnet/Home/wiki/Engineering-guidelines#breaking-changes cc @Eilon

HaoK: Yeah this was mostly a placeholder for now, we have some time to craft the final message, since I can't check in this PR until I've updated all the downstream repos

HaoK:
Replaced by https://github.com/aspnet/HttpAbstractions/pull/866

PR #815, opened 2017-04-12
Add ChallengeOverload for MVC

HaoK:
@Tratcher needed for ChallengeResult which uses an overload that only passes in Properties today.


Tratcher (approved):

Tratcher:
:shipit:

HaoK:
f6cf5293a08be59005b42d896a19a7f70d9e4091

PR #818, opened 2017-04-15
Remove unneeded AuthenticateContext

HaoK:
@Tratcher


HaoK:
Also adds GetToken overload that uses default authenticate scheme

PR #821, opened 2017-04-20
Added feature to control http body behavior

davidfowl:
- Added a flag to allow synchronous IO

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Features/IHttpBodyControlFeature.cs
+++ b/src/Microsoft.AspNetCore.Http.Features/IHttpBodyControlFeature.cs
@@ -1,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Http.Features
+{
+    /// <summary>
+    /// Controls the IO behavior for the <see cref="IHttpRequestFeature.Body"/> and <see cref="IHttpResponseFeature.Body"/> 
+    /// </summary>
+    public interface IHttpBodyControlFeature
+    {
+        /// <summary>
+        /// Gets or sets a value that controls whether synchronous IO is allowed for the <see cref="IHttpRequestFeature.Body"/> and <see cref="IHttpResponseFeature.Body"/> 
+        /// </summary>
+        bool AllowSynchronousIO { get; set; }
+    }
+}
--- a/src/Microsoft.AspNetCore.Http.Features/IHttpBodyControlFeature.cs
+++ b/src/Microsoft.AspNetCore.Http.Features/IHttpBodyControlFeature.cs
@@ -1,6 +1,5 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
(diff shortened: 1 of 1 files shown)


JunTaoLuo:
I assume there is more to this PR than just this?

davidfowl:
> I was expecting a global flag on KestrelOptions, not a per request flag. Or do you plan to do both?

Both.

> I assume there is more to this PR than just this?

This is just adding the feature. I don't plan to push all at once. Feature first, then Kestrel change once it goes in.

src/Microsoft.AspNetCore.Http.Features/IHttpBodyControlFeature.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
```
JunTaoLuo: copyright

src/Microsoft.AspNetCore.Http.Features/IHttpBodyControlFeature.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Collections.Generic;
```
JunTaoLuo: What are you using from these namespaces?

src/Microsoft.AspNetCore.Http.Features/IHttpBodyControlFeature.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Text;
+
+namespace Microsoft.AspNetCore.Http.Features
+{
+    /// <summary>
+    /// Controls the IO behavior for the <see cref="IHttpRequestFeature.Body"/> and <see cref="IHttpResponseFeature.Body"/> 
+    /// </summary>
+    public interface IHttpBodyControlFeature
```
JunTaoLuo: Not sure about the naming BodyControl. Maybe IHttpBodyOptionsFeature, IHttpBodyIOFeature, IHttpBodyCapabilitiesFeature?

halter73:
@Tratcher Is HttpSysServer going to the same global flag?

It would be pretty simple to disallow sync IO using some middleware to wrap the request and response body streams. The perf hit won't be too bad if the wrapping stream is pooled and lazily initialized. It would also ensure that a server switch doesn't suddenly cause a bunch of Read/Write calls to fail.

JunTaoLuo:
Still don't like the naming, but functionally it's fine.

davidfowl:
@DamianEdwards do you have any naming suggestions?

src/Microsoft.AspNetCore.Http.Features/IHttpBodyControlFeature.cs
rynowak: Body Control sounds like the album name

DamianEdwards:
`IHttpBodyIOControlFeature`
`IHttpIOControlFeature`
`IHttpBodyIOFeature`
`IHttpBodySyncIOControlFeature`
`IHttpRequestResponseBodySynchronousIOControlFeature`
`IHttpIOIOIOItsOffToWorkWeGoFeature`

PR #840, opened 2017-05-17
Target netstandard2.0 TFM

JunTaoLuo:


test/Microsoft.AspNetCore.Http.Extensions.Tests/Microsoft.AspNetCore.Http.Extensions.Tests.csproj
```diff
@@ -14,6 +14,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(TestSdkVersion)" />
+    <PackageReference Include="NETStandard.Library.NETFramework" Version="$(NETStandardLibraryNETFrameworkVersion)" />
```
pranavkm: Could we add this to common.props? Easier to remove it once we figure out if we don't need it.

JunTaoLuo: Trying @natemcmaster 's trick with the Directory.Build.targets

natemcmaster: @JunTaoLuo that was meant to be temporary as I'm still under the impression that NS.Library.NETFramework should be automatic. If it won't be, we should put it in common.props

JunTaoLuo: Okay I'll wait for a verdict on this before merging.

test/Microsoft.Net.Http.Headers.Tests/Microsoft.Net.Http.Headers.Tests.csproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
```
pranavkm: We want to conditionally cross-compile these depending on the OS. Don't want desktop tests running on xplat (I think)?

JunTaoLuo: Ah yes

test/Microsoft.AspNetCore.Authentication.Core.Test/Microsoft.AspNetCore.Authentication.Core.Test.csproj
```diff
@@ -2,7 +2,8 @@
 
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net461</TargetFrameworks>
```
natemcmaster: Can you check that multi-TFM works in VS? In 15.0-15.2 the project system parses csproj as XML which means it IntelliSense would only recognize one TFM, not two. This is supposedly fixed in 15.3. cref https://github.com/dotnet/project-system/issues/547#issuecomment-277379863.

The workaround is to invert the logic to:
```xml
<TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
<TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.0</TargetFrameworks>
```

JunTaoLuo: Ah so that's why we used that order. I verified that this works in 15.3 (26510.0-Preview) though so it seems fixed.

natemcmaster: :+1:

JunTaoLuo:
:up::date:

build/common.props
```diff
@@ -20,4 +20,7 @@
     <PackageReference Include="NETStandard.Library" Version="$(BundledNETStandardPackageVersion)" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)'=='net461'">
```
natemcmaster: Are there any projects in src that target .NET Framework?

build/common.props
```diff
@@ -20,4 +20,7 @@
     <PackageReference Include="NETStandard.Library" Version="$(BundledNETStandardPackageVersion)" />
```
natemcmaster: In most repos, this item group is dead code now. It can be removed now unless there are projects in src that still target .NET Framework

build/common.props
JunTaoLuo: nope

build/common.props
JunTaoLuo: Removed

JunTaoLuo:
This can go in to unblock downstream repos since it doesn't depend on any upstream repos (testing, jsonpatch, ...)

samples/SampleApp/SampleApp.csproj
```diff
@@ -1,9 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <Import Project="..\..\build\dependencies.props" />
+  <Import Project="..\..\build\common.props" />
```
natemcmaster: 😬 ehh, we're trying to avoid this. This will bring in Internal.AspNetCore.Sdk, strong naming, etc. Can you just add NS.Library.NETFramework explicitly here instead?

Maybe we really should switch to directory.build.{props,targets}....

JunTaoLuo:
ping @pranavkm @natemcmaster @kichalla

Issue #758, opened 2017-01-19
Reduce allocations by using StringSegment in parsers


JunTaoLuo:
We can potentially reduce the amount of allocations if we expose some of the parsed types and intermediate tokens as StringSegment instead of strings by removing the need to call Substring everywhere.

Tratcher:
This could be done either as an internal optimization to the existing parsers (to limited effect) or by creating a new version that directly exposes StringSegment on the resulting object model. Thus far we have not exposed StringSegment in public APIs.

JunTaoLuo:
@BrennanConroy also suggest looking into using spans if available and possible. Could also see if that applies to StringSegment in Common as well.

Tratcher:
Note MVC has already done something like this. They cloned the existing parsers and reworked them in their own private library to be more efficient. We should try to reclaim those.

JunTaoLuo:
proposal: consolidate our custom header parsing. We should look at Mvc, Localization, HeaderUtilties, and StaticFiles.

Tratcher:
Localization just uses the TypedHeaders: https://github.com/aspnet/Localization/blob/861ae52de51f174e14701d6cdbb88e6543638b03/src/Microsoft.AspNetCore.Localization/AcceptLanguageHeaderRequestCultureProvider.cs#L33

Tratcher:
See https://github.com/aspnet/Mvc/blob/d19150ed227ed81f28680f1235339f77c9f0df86/src/Microsoft.AspNetCore.Mvc.Core/Formatters/MediaType.cs

Tratcher:
https://github.com/aspnet/Common/issues/213

Eilon:
@muratg please file a new bug for the consolidation work.

muratg:
Consolidation work: https://github.com/aspnet/HttpAbstractions/issues/825

Issue #812, opened 2017-04-10
Add Microbenchmarks for Microsoft.AspNetCore.WebUtilities


davidfowl:
- Multipart stream reading
- Header parsing and formatting
- Form parsing via FormReader
- Base64UrlTextEncoder

Tratcher:
And Microsoft.Net.Http.Headers

aspnet-hello:
This issue was moved to aspnet/Home#2707

PR #844, opened 2017-05-19
Convert the header parsers to use StringSegment

Tratcher:
#758 This primarily covers Microsoft.Net.Http.Headers, and reactions to those changes. As a rule I changed all strings to StringSegments where sensible, but did not expose it on the user friendly APIs like Form.

Working on the reactions across the Universe.

### DIFF
--- a/src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
+++ b/src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
@@ -6,7 +6,9 @@ using System.Buffers;
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Globalization;
+using System.Linq;
 using System.Text;
+using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.Net.Http.Headers
 {
@@ -20,26 +22,28 @@ namespace Microsoft.Net.Http.Headers
         private const string ModificationDateString = "modification-date";
         private const string ReadDateString = "read-date";
         private const string SizeString = "size";
+        private static readonly char[] QuestionMark = new char[] { '?' };
+        private static readonly char[] SingleQuote = new char[] { '\'' };
 
         private static readonly HttpHeaderParser<ContentDispositionHeaderValue> Parser
             = new GenericHeaderParser<ContentDispositionHeaderValue>(false, GetDispositionTypeLength);
 
         // Use list instead of dictionary since we may have multiple parameters with the same name.
         private ObjectCollection<NameValueHeaderValue> _parameters;
-        private string _dispositionType;
+        private StringSegment _dispositionType;
 
         private ContentDispositionHeaderValue()
         {
             // Used by the parser to create a new instance of this type.
         }
 
-        public ContentDispositionHeaderValue(string dispositionType)
+        public ContentDispositionHeaderValue(StringSegment dispositionType)
         {
             CheckDispositionTypeFormat(dispositionType, "dispositionType");
             _dispositionType = dispositionType;
         }
 
-        public string DispositionType
+        public StringSegment DispositionType
         {
             get { return _dispositionType; }
             set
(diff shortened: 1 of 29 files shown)


src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
```diff
@@ -479,15 +477,23 @@ private bool RequiresEncoding(string input)
...
+            fixed (char* chars = input.Buffer)
+            {
+                var byteCount = Encoding.UTF8.GetByteCount(chars + input.Offset, input.Length);
+                var buffer = new byte[byteCount];
+                fixed (byte* bytes = buffer)
+                {
+                    Encoding.UTF8.GetBytes(chars + input.Offset, input.Length, bytes, byteCount);
+                }
+                var encodedName = Convert.ToBase64String(buffer);
+                return string.Format(CultureInfo.InvariantCulture, "=?utf-8?B?{0}?=", encodedName);
```
davidfowl: Remove the string.Format and just use concat

src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
```diff
@@ -498,7 +504,8 @@ private bool TryDecodeMime(string input, out string output)
             {
                 return false;
             }
-            var parts = processedInput.Split('?');
+            // TODO: StringSegment.Split
```
davidfowl: When?

davidfowl:
Please add Microbenchmarks as part of this work.

src/Microsoft.Net.Http.Headers/SetCookieHeaderValue.cs
```diff
@@ -342,15 +341,15 @@ private static int GetSetCookieLength(string input, int startIndex, out SetCooki
                     }
                     var dateString = ReadToSemicolonOrEnd(input, ref offset);
                     DateTimeOffset expirationDate;
-                    if (!HttpRuleParser.TryStringToDate(dateString, out expirationDate))
+                    if (!HttpRuleParser.TryStringToDate(dateString.ToString(), out expirationDate))
```
davidfowl: Why not update the `HttpRuleParser` to take `StringSegment`

Tratcher: It's just a wrapper for DateTimeOffset.TryParseExact and I didn't want to hide the ToString.

davidfowl: It should stay a wrapper so that we can change the implementation in the future. To the ToString() on the inside not on the outside.

Tratcher: HttpRuleParser is all internal, we can change its signature whenever. Leaving it makes it more obvious that we should revisit it, not that I expect to be writing a date parser from scratch any time soon.

davidfowl: I don't see why you wouldn't change the signature to `StringSegment`, you literally changed everything else. What are you trying to preserve?

davidfowl: > not that I expect to be writing a date parser from scratch any time soon.

You don't have to the corefx team will.

Tratcher: I changed the ones that could actually utilize StringSegments and left the ones that had to use strings anyways. That way the caller can tell if it's going to be efficient or not.

davidfowl: I think we should just change it to be `StringSegment` and just call ToString() in the appropriate places.

Tratcher:
@davidfowl https://github.com/aspnet/HttpAbstractions/issues/812. Not enough time to do it now.

davidfowl:
@Tratcher not enough time to add a project and a few tests?

Tratcher:
Yes, not enough time. I've got less than three days left to address a bunch of functional issues, now is not the time to add non-essential work to that list.

davidfowl:
> Yes, not enough time. I've got less than three days left to address a bunch of functional issues, now is not the time to add non-essential work to that list.

That's funny because these changes are all for performance.

src/Microsoft.Net.Http.Headers/ObjectCollection.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Net.Http.Headers
     // type to throw if 'null' gets added. Collection<T> internally uses List<T> which comes at some cost. In addition
     // Collection<T>.Add() calls List<T>.InsertItem() which is an O(n) operation (compared to O(1) for List<T>.Add()).
     // This type is only used for very small collections (1-2 items) to keep the impact of using Collection<T> small.
-    internal class ObjectCollection<T> : Collection<T> where T : class
+    internal class ObjectCollection<T> : Collection<T>
```
JunTaoLuo: Naming might be a little inconsistent now that we allow value types.

PR #850, opened 2017-05-24
React to breaking change latest API Check detects

dougbu:
- see PR aspnet/BuildTools#259 and issue aspnet/BuildTools#146


dougbu:
@Tratcher was f5107596a5486f9531276af290ac32d2a62780ec an intentional breaking change? If not, we should revert it before I check aspnet/BuildTools#259 in.

Tratcher:
Yes it was.

dougbu:
5b9e129

Issue #848, opened 2017-05-24
PathString should have a registered TypeConverter


MarkPflug:
I ran into an issue while attempting to configure the OpenIdConnect provider using the Options/configuration APIs. The issue was around trying to assign a the OpenIdConnect.CallbackPath value which is of type PathString. This failed, with a conversion error trying to convert my string value from the configuration to the PathString property.
The original issue is here: https://github.com/aspnet/Security/issues/1230

This can resolved by reigistering a TypeConverter for PathString. However, it seems like such a TypeConverter would be best defined along side the PathString class itself.

HaoK:
Added for preview 2, but I need to harden the implementation + add a few more tests, so keeping this on my plate for now

HaoK:
12f89f66a687bdb1c1b683b0d7abd2732c05eb1a

Issue #690, opened 2016-08-16
PathString implicit string converters are asymmetric 


jkotalik:
In PathString.cs

```
public static implicit operator PathString(string s)
        {
            return new PathString(s);
        }
```

Will create the unescaped new PathString while

```
        public static implicit operator string(PathString path)
        {
            return path.ToString();
        }
```

Will create the escaped ToString.

davidfowl:
That's not great...

khellang:
I thought that was the whole point of `PathString`? Make it easier to do "the right thing" by expecting an unescaped and outputting an escaped string. This prevents someone accidentally `ToString`'ing a path in its unescaped form.

Tratcher:
@khellang We didn't specify which was wrong, just that it's asymmetric. If anything I think it's the string -> PathString converter that's wrong, it should use FromUriComponent to maintain symmetry. Unfortunately that's not how we normally to use it either.

muratg:
This would be a breaking change. We'll reconsider in 2.0

Tratcher:
@muratg we should re-triage this if we're ever going to fix it.

HaoK:
12f89f66a687bdb1c1b683b0d7abd2732c05eb1a

PR #857, opened 2017-06-06
Harden the PathString converter

HaoK:
Fixes https://github.com/aspnet/HttpAbstractions/issues/848 https://github.com/aspnet/HttpAbstractions/issues/690

@Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
@@ -443,9 +443,6 @@ namespace Microsoft.AspNetCore.Http
         /// </summary>
         /// <param name="s"></param>
         public static implicit operator PathString(string s)
-        {
-            return new PathString(s);
-        }
 
         /// <summary>
         /// Implicitly calls ToString().
@@ -455,6 +452,9 @@ namespace Microsoft.AspNetCore.Http
         {
             return path.ToString();
         }
+
+        internal static PathString ConvertFromString(string s)
+            => string.IsNullOrEmpty(s) ? new PathString(s) : FromUriComponent(s);
     }
 
     internal class PathStringConverter : TypeConverter
@@ -466,7 +466,7 @@ namespace Microsoft.AspNetCore.Http
 
         public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
             => value is string 
-            ? new PathString((string)value) 
+            ? PathString.ConvertFromString((string)value) 
             : base.ConvertFrom(context, culture, value);
 
         public override object ConvertTo(ITypeDescriptorContext context,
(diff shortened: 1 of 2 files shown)


davidfowl:
Isn't this in preview2?

HaoK:
Yeah this is just cleanup, but I can make this change in preview 2 if we want

src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs
```diff
@@ -459,9 +459,20 @@ public override int GetHashCode()
...
+            => sourceType == typeof(string) 
+            ? true 
+            : base.CanConvertFrom(context, sourceType);
+
         public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
-        {
-            return new PathString((string)value);
-        }
+            => value is string 
+            ? new PathString((string)value)
```
Tratcher: This replicates the asymmitry bug from the implicit converters:
https://github.com/aspnet/HttpAbstractions/issues/690

string -> PathString should use FromUriComponent(string), not new PathString(string)

We should change them both at the same time. @davidfowl

HaoK: Let me see if I can unify the code a bit to make that change easier

HaoK:
So I added a new 
```
internal static PathString ConvertFromString(string s) => string.IsNullOrEmpty(s) ? new PathString(s) : FromUriComponent(s);
```

That both the implicit op and ConvertFrom call, but it wasn't obvious to me exactly how to write a test verifying that the round tripping issue was fixed.

Tratcher:
The most obvious things that are broken by the asymmetry are the equality operators. E.g. Equals, StartsWithSegments, etc..
Here's a repro:
```
            PathString p1 = "/?";
            String s1 = p1; // Becomes "/%3F"
            PathString p2 = s1; // Stays "/%3F"
            Console.WriteLine(p1.Equals(p2)); // false
```

HaoK:
Added a test verifying the equality now work

HaoK:
Now also fixes: https://github.com/aspnet/HttpAbstractions/issues/690

test/Microsoft.AspNetCore.Http.Abstractions.Tests/PathStringTests.cs
```diff
@@ -208,11 +208,28 @@ public void ToUriComponentEscapeCorrectly(string category, string input, string
...
+            Assert.False(converter.CanConvertTo(typeof(int)));
+            Assert.False(converter.CanConvertTo(typeof(bool)));
         }
+
+        [Fact]
+        public void PathStringStaysEqualAfterAssignments()
+        {
+            PathString p1 = "/?";
+            string s1 = p1; // Becomes "/%3F"
+            PathString p2 = s1; // Stays "/%3F"
```
Tratcher: `Stays "/%3F"` is wrong after your fix

Tratcher:
*ping*

HaoK:
Yeah I'll merge this later today

HaoK:
12f89f66a687bdb1c1b683b0d7abd2732c05eb1a

Issue #861, opened 2017-06-07
Deprecate IHttpAuthenticationFeature.IAuthenticationHandler


davidfowl:
We deprecated the property on HttpContext but not on the feature interface

https://github.com/aspnet/HttpAbstractions/blob/c06ed8c29abe606ad7a36915d85d066e59699f27/src/Microsoft.AspNetCore.Http.Features/Authentication/IHttpAuthenticationFeature.cs#L12

/cc @Tratcher @HaoK

Tratcher:
We're still using .User

Tratcher:
Oh, I see. Nobody uses the feature interfaces directly

davidfowl:
So we should just outright delete the property then.

Tratcher:
There was some theory about not killing 1.1 middleware, but I don't know how realistic that is.

PR #863, opened 2017-06-07
Deprecate the IAuthenticationHandler property

davidfowl:
#861

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Features/Authentication/IHttpAuthenticationFeature.cs
+++ b/src/Microsoft.AspNetCore.Http.Features/Authentication/IHttpAuthenticationFeature.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Security.Claims;
 
 namespace Microsoft.AspNetCore.Http.Features.Authentication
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Http.Features/Authentication/IHttpAuthenticationFeature.cs
```diff
@@ -9,6 +10,7 @@ public interface IHttpAuthenticationFeature
     {
         ClaimsPrincipal User { get; set; }
 
+        [Obsolete("This is obsolete and will be removed in a future version. The recommended alternative is to use Microsoft.AspNetCore.Authentication.AuthenticationHttpContextExtensions. See https://go.microsoft.com/fwlink/?linkid=845470.")]
```
Tratcher: Drop "The recommended alternative is to use Microsoft.AspNetCore.Authentication.AuthenticationHttpContextExtensions."? It's not a direct replacement for this API.

davidfowl:
@HaoK Just the entry points.

Tratcher:
We should at least do AuthenticationManager. That would get any code that was accepting it.

HaoK:
@Tratcher just the class and not all of the methods right?

Tratcher:
At least the class. And at that point why not do the methods too?

HaoK:
How would they ever call these methods without going through context.Authentication?

Tratcher:
`public void DoAuth(AuthenticationManager manager);`

Yes someone would have to call context.Authentication before calling this method, but that may not even be in the same project as a method accepting AuthenticationManager.

HaoK:
So in that usage it would be enough to have it on the class then right?  They'd only have access to the methods via context.Authentication or an explicit type reference

Tratcher:
Right, but at that point why not just do all the methods? It's just some copy-paste?

HaoK:
I'm just trying to avoid the viral spread of obsolete/pragmas, if the methods don't spread pragamas sure...

PR #864, opened 2017-06-08
Don't throw for OnStarting or OnCompleted calls.

davidfowl:
I had to implement a ResponseFeature in SignalR that doesn't throw because we do by default. It *really* shouldn't throw.


davidfowl:
@JunTaoLuo yea the problem is tests not real apps

Issue #721, opened 2016-10-06
Implement MapRun


Tratcher:
Map and Run are a common combination. Add an extension that combines the two with MapRun.

Before:

```
            app.Map("/hello", subApp =>
            {
                subApp.Run(context =>
                {
                    return context.Response.WriteAsync("Hello world");
                });
            });
```

After:

```
            app.MapRun("/hello", context =>
            {
                 return context.Response.WriteAsync("Hello world");
            });
```

davidfowl:
There's some overlap here with routing. Also, we don't plan to build a child pipeline for this do we? That seems wasteful.

muratg:
It can be basic sugar method, no?

Tratcher:
Map has semantics about updating PathBase that I don't think would be needed here. `MatchRun` or `RunMatch` might be better. There's no need to build a sub-pipeline or update PathBase.

muratg:
cc @glennc

muratg:
cc @glennc

Tratcher:
Killed by @davidfowl

PR #869, opened 2017-06-13
Implement RunAt

Tratcher:
#721 
Intended usage: https://github.com/aspnet/Security/blob/9797d4bc5f3e54ffca3487c038044990585e278c/samples/SocialSample/Startup.cs#L214

I decided not to call it MapRun because the the mechanics are a bit different, Map adjust the PathBase.


test/Microsoft.AspNetCore.Http.Abstractions.Tests/RunExtensionsTests.cs
```diff
@@ -0,0 +1,140 @@
...
+namespace Microsoft.AspNetCore.Builder.Extensions
+{
+    public class RunExtensionsTests
+    {
+        [Fact]
+        public void Run_NullArguments_ArgumentNullException()
+        {
+            var builder = CreateBuilder();
+
+            Action nullConfiguration = () => builder.Run(null);
```
JunTaoLuo: var instead of Action? Or just place the action in the Assert.Throws statement, it's pretty clear what it's doing.

test/Microsoft.AspNetCore.Http.Abstractions.Tests/RunExtensionsTests.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            Assert.Throws<ArgumentNullException>(nullConfiguration);
+        }
+
+        [Fact]
+        public void RunAt_NullArguments_ArgumentNullException()
+        {
+            var builder = CreateBuilder();
+
+            Action nullPath = () => builder.RunAt(null, _ => Task.CompletedTask);
```
JunTaoLuo: Same here.

test/Microsoft.AspNetCore.Http.Abstractions.Tests/RunExtensionsTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            builder.RunAt(match, Increment("first"));
+            builder.Run(Increment("next"));
+
+            await builder.Build().Invoke(context);
+
+            Assert.Equal(0, Count(context, "first"));
+            Assert.Equal(1, Count(context, "next"));
+        }
+
+        private static HttpContext CreateContext()
```
JunTaoLuo: => new DefaultContext();

test/Microsoft.AspNetCore.Http.Abstractions.Tests/RunExtensionsTests.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            return new DefaultHttpContext();
+        }
+
+        private static ApplicationBuilder CreateBuilder()
```
JunTaoLuo: => new ApplicationBuilder(...);

davidfowl:
Why the name RunAt?

davidfowl:
Also, why not just use routing for this? It's works much better than these methods.

Tratcher:
Run the given action for requests at the given path. RunAtPath? RunForPath? RunFor?

Routing is definitely the advanced story, but I've never needed advanced for these scenarios. Also none of the places I want this can depend on routing due to repo layering.

davidfowl:
> Routing is definitely the advanced story, but I've never needed advanced for these scenarios. Also none of the places I want this can depend on routing due to repo layering.

It's really not that advanced. Why can't security depend on routing?

Tratcher:
Current layers:
```
10: IISIntegration
11: CORS, ResponseCaching, Routing, ServerTests, Session, StaticFiles
12: Diagnostics, Localization
13: Security, WebSockets
```
Looks like Security could use Routing without chagning layering, I hadn't realized it was that low already. The rest of level 11 couldn't without adding more layers though.
e.g. https://github.com/aspnet/Session/blob/dev/samples/SessionSample/Startup.cs#L47

I tried the Security samples with Routing and it's not much of an improvement over Map + Run.
https://github.com/aspnet/Security/commit/2e7b21d6c322dd17742830ae865768aa754ea239

The same samples get a bit simpler with RunAt.
https://github.com/aspnet/Security/commit/eebe484d19b80029388f9dbd23f315b008111830?w=1

davidfowl:
I care almost nothing about the security repository layering with respect to this extension method. I have plans post 2.0 to push routing further down the stack and this IMO doesn't make sense to add.

PR #870, opened 2017-06-13
DefaultForbid/SignOutSchemes + tests

HaoK:
Part of the fix for: https://github.com/aspnet/Mvc/issues/6392

Also added some tests + a sugar overload for AddScheme<THandler>

@Tratcher @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
@@ -48,6 +48,19 @@ namespace Microsoft.AspNetCore.Authentication
         }
 
         /// <summary>
+        /// Adds an <see cref="AuthenticationScheme"/>.
+        /// </summary>
+        /// <typeparam name="THandler">The <see cref="IAuthenticationHandler"/> responsible for the scheme.</typeparam>
+        /// <param name="name">The name of the scheme being added.</param>
+        /// <param name="displayName">The display name for the scheme.</param>
+        public void AddScheme<THandler>(string name, string displayName) where THandler : IAuthenticationHandler
+            => AddScheme(name, b =>
+            {
+                b.DisplayName = displayName;
+                b.HandlerType = typeof(THandler);
+            });
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
```diff
@@ -48,6 +48,19 @@ public void AddScheme(string name, Action<AuthenticationSchemeBuilder> configure
         }
 
         /// <summary>
+        /// Adds an <see cref="AuthenticationScheme"/>.
+        /// </summary>
+        /// <typeparam name="THandler">The <see cref="IAuthenticationHandler"/> responsible for the scheme.</typeparam>
+        /// <param name="name">The name of the scheme being added.</param>
+        /// <param name="displayName">The display name for the scheme.</param>
+        public void AddScheme<THandler>(string name, string displayName) where THandler : IAuthenticationHandler
```
Tratcher: Doesn't look like this will be used outside of tests. No?

HaoK: Sure it will, it's what end users could call to add handlers, its analogous to what hosting does: https://github.com/aspnet/IISIntegration/blob/dev/src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs#L53

except that's adding directly to the scheme provider rather than during startup/configure services.

Keep in mind, there's NOTHING on the builder, so the question I'm asking now is if we should nuke all of the `Action<AuthenticationSchemeBuilder>`...  It was more useful when we had a bag on the builder that could be configured.

HaoK: Its just Name + DisplayName + Type, I guess its slightly more future proof with the action.

HaoK: But mainly I just wanted the sugar, we had a bug for this that got punted to 2.1, but since I'm here and adding tests :) https://github.com/aspnet/Security/issues/1186

PR #873, opened 2017-06-16
Add defaults + split SignIn/SignOut

HaoK:
- Move SignIn/Signout into IAuthenticationHandlerSign[In\Out]Handler
- Add tests for AuthenticationService
- add overloads for SignOut/Forbid to use new default

@Tratcher


HaoK:
Fixes https://github.com/aspnet/Security/issues/1244 and https://github.com/aspnet/Security/issues/1264

Need to update Security/Hosts (to remove SignIn/Signout impls only) as a result of this PR as well

HaoK:
I'm going to combine the events changes we want for https://github.com/aspnet/Security/pull/1266 into this PR as well.  So the update to AuthenticationResult + rename of None => Ignore will go in this PR as well

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -93,7 +90,7 @@ public Task<AuthenticationScheme> GetDefaultForbidSchemeAsync()
         /// <summary>
         /// Returns the scheme that will be used by default for <see cref="IAuthenticationService.SignInAsync(HttpContext, string, System.Security.Claims.ClaimsPrincipal, AuthenticationProperties)"/>.
         /// This is typically specified via <see cref="AuthenticationOptions.DefaultSignInScheme"/>.
-        /// Otherwise, if only a single scheme exists, that will be used, if more than one exists, null will be returned.
+        /// Otherwise, this will fallback to <see cref="GetDefaultAuthenticateSchemeAsync"/>.
```
Tratcher: Update comment to include the single scenario?

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -121,6 +118,10 @@ public Task<AuthenticationScheme> GetDefaultSignOutSchemeAsync()
             {
                 return GetSchemeAsync(_options.DefaultSignOutScheme);
             }
+            if (_signOutHandlers.Count() == 1)
```
Tratcher: Add to doc comments?

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -121,6 +118,10 @@ public Task<AuthenticationScheme> GetDefaultSignOutSchemeAsync()
...
+            {
+                return Task.FromResult(_signOutHandlers[0]);
+            }
             return GetDefaultSignInSchemeAsync();
```
Tratcher: There's now no guarantee that a signin handler supports signout. Maybe we should change the hierarchy to ISignIn -> ISignOut -> IAuth

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -167,6 +168,14 @@ public void AddScheme(AuthenticationScheme scheme)
...
                 }
+                if (typeof(IAuthenticationSignInHandler).IsAssignableFrom(scheme.HandlerType))
+                {
+                    _signInHandlers.Add(scheme);
+                }
+                if (typeof(IAuthenticationSignOutHandler).IsAssignableFrom(scheme.HandlerType))
+                {
+                    _signOutHandlers.Add(scheme);
+                }
                 _map[scheme.Name] = scheme;
```
Tratcher: There's no check that the HandlerType actually implements IAuthenticationHandler

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -102,11 +99,11 @@ public Task<AuthenticationScheme> GetDefaultSignInSchemeAsync()
             {
                 return GetSchemeAsync(_options.DefaultSignInScheme);
             }
-            if (_map.Count == 1)
+            if (_signInHandlers.Count() == 1)
```
campersau: Why using Linq?

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
campersau: Why using Linq?

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
HaoK: I was just mirroring what we have in security today, where we have OIDC only implementing remote sign out...

HaoK: Maybe we just tweak this also check that the handler for sign in scheme implements sign out before returning it, we can have hit continue to fallback to sign in

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
HaoK: That's done in the scheme itself, https://github.com/aspnet/HttpAbstractions/blob/88a797cffddb43e586def38acc17c9435367a19a/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs#L31

src/Microsoft.AspNetCore.Authentication.Abstractions/IAuthenticationSignOutHandler.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    /// <summary>
+    /// Used to determine if a handler supports SignOut.
+    /// </summary>
+    public interface IAuthenticationSignOutHandler : IAuthenticationHandler
```
davidfowl: Why the derivation?

src/Microsoft.AspNetCore.Authentication.Abstractions/IAuthenticationSignInHandler.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System.Security.Claims;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    /// <summary>
+    /// Used to determine if a handler supports SignIn.
+    /// </summary>
+    public interface IAuthenticationSignInHandler : IAuthenticationHandler
```
davidfowl: Why the derivation?

src/Microsoft.AspNetCore.Authentication.Abstractions/IAuthenticationSignOutHandler.cs
HaoK: We require every handler to implement authenticate/challenge/forbid at a minimum, derivation makes this clearer, if we didn't require this, wouldn't that lead ppl to think they could only implement SignOut and use their handler?

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
HaoK: This problem also exists for the fallback for SignIn => Authenticate, so we need to add a similar check for that too

Tratcher: No, the SignIn interface derives from the Auth interface, but SignOut does not derive from SignIn.

HaoK: That's ok, I just mean:
```
 var signIn = GetDefaultSignInSchemeAsync();
 if (signIn != null && typeof(IAuthenticationSignOutHandler).IsAssignableFrom(signIn.HandlerType)) return signIn;
 else return null;
```

PinpointTownes:
So much sorcery in this PR... :cry: 

It would be much better to throw an `InvalidOperationException` when there's more than 1 scheme to encourage the developers to read the related documentation, so they can set the different schemes themselves.

PinpointTownes:
Related post: https://github.com/aspnet/Security/pull/1267#discussion_r122566994.

HaoK:
Updated the default sign/signout to enforce the handlers implement the interface, otherwise they return null (should this throw instead?)

HaoK:
We don't throw if the default scheme doesn't exist today, so it feels weird to throw

HaoK:
We'll also make SignIn derive from SignOut which dervies in Auth

HaoK:
Updated, SignInHandler is now the old uber AuthenticationHandler that has all methods implemented.  Renamed None() => NoResult() with the bool property Nothing => None (done for Security)

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -93,43 +90,49 @@ public Task<AuthenticationScheme> GetDefaultForbidSchemeAsync()
...
+        /// this will fallback to <see cref="GetDefaultSignInSchemeAsync"/> if that supoorts sign out.
         /// </summary>
         /// <returns>The scheme that will be used by default for <see cref="IAuthenticationService.SignOutAsync(HttpContext, string, AuthenticationProperties)"/>.</returns>
-        public Task<AuthenticationScheme> GetDefaultSignOutSchemeAsync()
+        public virtual async Task<AuthenticationScheme> GetDefaultSignOutSchemeAsync()
         {
             if (_options.DefaultSignOutScheme != null)
             {
-                return GetSchemeAsync(_options.DefaultSignOutScheme);
+                return EnsureAssignable<IAuthenticationSignOutHandler>(await GetSchemeAsync(_options.DefaultSignOutScheme));
```
Tratcher: The type check is duplicated in SignOutAsync

HaoK: Oh right, ok I can back out the extra complexity in the scheme provider then, and leave the type interface enforcement to the service

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticateResult.cs
```diff
@@ -64,22 +58,12 @@ public static AuthenticateResult Success(AuthenticationTicket ticket)
...
-        {
-            return new AuthenticateResult() { Handled = true };
-        }
-
-        /// <summary>
         /// Indicates that there was no information returned for this authentication scheme.
         /// </summary>
         /// <returns>The result.</returns>
-        public static AuthenticateResult None()
+        public static AuthenticateResult NoResult()
```
HaoK: @Tratcher how about we keep this None, but call it NoResult() on the ResultContext's that will get exposed to events?

Tratcher: That would leave you with a method and property that have the same name. NoResult is fine here.

HaoK: These are currently called Nothing and None(), I just don't see much benefit renaming None => NoResult as we do have usages in the code internally already today (i.e. Identity), we can keep NoResult on ResultContext so the public API for events is still the same.

Tratcher: NoResult() and None are better than None() and Nothing.

HaoK:
271faf11bbd5b05cd758f1c7e83eb59d45b6db59

PR #875, opened 2017-06-20
Proposal: Change default challenge and forbid behavior

davidfowl:
- ChallengeAsync with a scheme that has no handler will send a 401 with a
WWW-Authenticate header matching that scheme.
- ChallengeAsync without a scheme and
- ForbidAsync returns a 403 by default and continues to throw if there's no handler
for the specified scheme if any.

PS: This is a PR for discussion.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
@@ -5,6 +5,7 @@ using System;
 using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.Authentication
 {
@@ -89,14 +90,17 @@ namespace Microsoft.AspNetCore.Authentication
                 scheme = defaultChallengeScheme?.Name;
                 if (scheme == null)
                 {
-                    throw new InvalidOperationException($"No authenticationScheme was specified, and there was no DefaultChallengeScheme found.");
+                    context.Response.StatusCode = StatusCodes.Status401Unauthorized;
+                    return;
                 }
             }
 
             var handler = await Handlers.GetHandlerAsync(context, scheme);
             if (handler == null)
             {
-                throw new InvalidOperationException($"No authentication handler is configured to handle the scheme: {scheme}");
+                context.Response.StatusCode = StatusCodes.Status401Unauthorized;
+                context.Response.Headers[HeaderNames.WWWAuthenticate] = scheme;
+                return;
             }
 
             await handler.ChallengeAsync(properties);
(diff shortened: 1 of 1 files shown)


Tratcher (changes requested):
This is counter productive, it doesn't help apps actually work, it just leaves them broken in a different way.

HaoK:
Perhaps a cleaner way to do this would be to always add a HttpAuthenticationHandler that implements this, and set DefaultChallenge/Forbid to that

HaoK:
Or maybe we put that in a AddHttpAuthentication() that lives in AbstractionsCore, that adds the handler, sets the defaults.

HaoK:
I'm not sure we should bake this kind of stuff into the service, instead of just having a real "Http" scheme that does this

HaoK:
But I guess we will have to have support behavior in the service to enable fallback for unknown Challenges to into the header

davidfowl:
> This is counter productive, it doesn't help apps actually work, it just leaves them broken in a different way.

It means you can use the same patterns to do things that should just work by default. They actually aren't broken in a different way in some cases. The APIs we tell people to use for auth do *http* things.

> Perhaps a cleaner way to do this would be to always add a HttpAuthenticationHandler that implements this, and set DefaultChallenge/Forbid to that

That seems fine, I just want it on by default. It means you also need to be able to handle unknown schemes which isn't really a thing and why I think it actually needs to go here.

HaoK:
So basically we are saying unknown schemes for Challenge/Forbid are okay.  Unlike Authenticate/SignIn/SignOut

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
```diff
@@ -89,14 +90,17 @@ public AuthenticationService(IAuthenticationSchemeProvider schemes, IAuthenticat
                 scheme = defaultChallengeScheme?.Name;
                 if (scheme == null)
                 {
-                    throw new InvalidOperationException($"No authenticationScheme was specified, and there was no DefaultChallengeScheme found.");
+                    context.Response.StatusCode = StatusCodes.Status401Unauthorized;
```
PinpointTownes: Note that returning a 401 response without a `WWW-Authenticate` header is illegal:

> A server generating a 401 (Unauthorized) response MUST send a WWW-Authenticate header field containing at least one challenge. A server MAY generate a WWW-Authenticate header field in other response messages to indicate that supplying credentials (or different credentials) might affect the response.

https://tools.ietf.org/html/rfc7235#section-4.1

Tratcher:
I think @blowdart and I talked him out of it. He's going to add Unauthorized() [401] and Forbidden() [403] to MVC instead to disambiguate from the auth system's Challenge() and Forbid().

HaoK:
Cool, much cleaner :)

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
davidfowl: How illegal.

PinpointTownes: So illegal that many clients crash/throw when they don't get the mandatory `WWW-Authenticate` header :smile:

PinpointTownes:
@Tratcher why not encouraging folks to use `return StatusCode(401/403)`? It would be less confusing.

Tratcher:
@PinpointTownes Yes, that's what we kept telling him...

davidfowl:
Because I'm not tainted by how katana worked 😄

Issue #725, opened 2016-10-10
Consider increasing HttpResponseStreamWriter.DefaultBufferSize constant


halter73:
As https://github.com/aspnet/Mvc/issues/5364#issuecomment-252486546 demonstrates, increasing `HttpResponseStreamWriter.DefaultBufferSize` to something only marginally larger like 4096, could nearly double the RPS of a controller action writing a large JsonResult.

@ajaybhargavb @Tratcher @mikeharder

benaadams:
Its taken from `ArrayPool` so shouldn't cause any extra memory use. On thing to watch is the overall distribution of default sizes (which should be powers of 2) as if they are all the same that may increase contention on a particular bucket of the pool.

ajaybhargavb:
cc @rynowak

mikeharder:
@halter73: Were thoughput tests run on the scenario in https://github.com/aspnet/Mvc/issues/5364#issuecomment-252486546?  The data I see in that issue is for single-request latency rather than throughput.

halter73:
Yeah. I'm extrapolating a bit. It's really reducing the latency for requests made over a single connection. I could be wrong about the RPS improvement, but it stands to reason.

mikeharder:
Also, is this the correct place to improve performance, or should it be done at an even lower layer?  Why does the buffer size at this layer make such a large difference?

halter73:
My suspicion is after Kestrel's response buffer fills up and flushes (at least partially), it causes a 1KB write instead of a potentially larger (e.g. 4KB) write which would be more efficient. This could be mitigated by something like nagleing or making Kestrel's default response buffer larger, but these both have trade offs like increased write latency and higher memory usage respectively.

rynowak:
@mikeharder - we're still not done collecting data for #5364

muratg:
@halter73 Do we have recent measurements with the increased buffer size?  Do we still want to do this in 2.0.0?

halter73:
MVC already bumped [MemoryPoolHttpResponseStreamWriterFactory.DefaultBufferSize](https://github.com/aspnet/Mvc/pull/6192/files) buffer size from 1k to 16k. That's what's used to serialize JsonResults by default.

Based on https://github.com/aspnet/Mvc/pull/6192, `HttpResponseStreamWriter.DefaultBufferSize` isn't actually used anywhere other than tests, so maybe that const should just be removed.

[...]

PR #877, opened 2017-06-28
Change HttpResponseStreamWriter.DefaultBufferSize

natemcmaster:
Resolves https://github.com/aspnet/HttpAbstractions/issues/725

See discussion on that issue about the reasoning for this change.

cc @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.WebUtilities/HttpResponseStreamWriter.cs
+++ b/src/Microsoft.AspNetCore.WebUtilities/HttpResponseStreamWriter.cs
@@ -16,11 +16,7 @@ namespace Microsoft.AspNetCore.WebUtilities
     public class HttpResponseStreamWriter : TextWriter
     {
         private const int MinBufferSize = 128;
-
-        /// <summary>
-        /// Default buffer size.
-        /// </summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.WebUtilities/HttpResponseStreamWriter.cs
```diff
@@ -16,11 +16,7 @@ namespace Microsoft.AspNetCore.WebUtilities
     public class HttpResponseStreamWriter : TextWriter
     {
         private const int MinBufferSize = 128;
-
-        /// <summary>
-        /// Default buffer size.
-        /// </summary>
-        public const int DefaultBufferSize = 1024;
+        internal const int DefaultBufferSize = 16 * 1024;
```
halter73: I would consider making this private unless you think it's important to reference from the InlineData attribute.

natemcmaster: I would have, but its value is also used in the FlushesBuffer_ButNotStream_OnFlushAsync test method body and is apparently significant to how the test runs.

Issue #800, opened 2017-03-22
Track recent breaking changes in `exceptions.net*.json` files


dougbu:
This repo contains untracked breaking changes. To find and address them
1. Revert c9f12a2
2. Run `.\build.cmd /t:Pack` in projects c9f12a2 touched or `.\build.cmd /t:Package` in the repo root
3. Update `exceptions.net45.json` and `exceptions.netcore.json` to track the changes
4. If not already announced, suggest filing issues at https://github.com/aspnet/Announcements

PR #878, opened 2017-06-28
Re-enable API check

Tratcher:
#800


src/Microsoft.AspNetCore.Http.Features/Microsoft.AspNetCore.Http.Features.csproj
```diff
@@ -8,7 +8,6 @@
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>aspnetcore</PackageTags>
-    <EnableApiCheck>false</EnableApiCheck>
```
dougbu: Wondering whether this deletion and the one in should be commited separately to make it easier to see c9f12a2 was reverted. Thoughts?

Tratcher: *shrug* I don't think it makes much of a difference.

PR #879, opened 2017-06-28
Add CancellationToken to ISession Load/CommitAsync

Tratcher:
Part of https://github.com/aspnet/Session/issues/167


src/Microsoft.AspNetCore.Http.Features/ISession.cs
```diff
@@ -28,13 +29,13 @@ public interface ISession
         /// Load the session from the data store. This may throw if the data store is unavailable.
         /// </summary>
         /// <returns></returns>
-        Task LoadAsync();
+        Task LoadAsync(CancellationToken cancellationToken);
```
davidfowl: Add `CancellationToken cancellationToken = default(CancellationToken)`

Issue #853, opened 2017-06-02
Reduce duplication when configuring cookie options


JunTaoLuo:
As per discussion in https://github.com/aspnet/Antiforgery/pull/141. We should use the Action<..., CookieOptions> pattern on each components' options to configure the cookie options instead of duplicating them. Need to look through our code base for where we configure cookie options and update them where necessary.

Tratcher:
Review:
- Security
- Session
- MVC
- Localization
- Rewrite

Tratcher:
@muratg these would be breaking changes that should be considered for RTM

HaoK:
Alternatively, you could consider using `IOptionsSnapshot<CookieOptions>` and using named options for each cookie assuming cookie.Name is unique.  Then configuring a cookie would be the normal `services.Configure<CookieOptions>(cookieName, o => { })` That seems like a reasonable alternative to a mini options pattern for these...

muratg:
cc @javiercn

javiercn:
We went with the Action<HttpContext,CookieOptions> approach in the OpenIdConnect/RemoteAuthentication options as needed for https://github.com/aspnet/Security/pull/1262.

The twitter middleware is the last one that requires this on the security repo as far as I know. We shouldn't do this for the cookie policy middleware IMO as it would broaden the scope of the middleware and would require more design, and its something that can be revisited on 2.1

muratg:
cc @Eilon, @danroth27, @DamianEdwards do we want this in 2.0.0 (and obsolete the existing Options?)  Or we can do it in 2.1, but then we can't obsolete them until 3.0.

muratg:
We'll try to get this in because of the reason I mentioned above.

@natemcmaster please consult with @Tratcher on the ideas how to do this.

natemcmaster:
I exampled the repos Chris listed above. To be consistent with the pattern established by Antiforgery, I believe these are the changes we would make:

### Security
  - [Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationOptions](https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs)
    - Obsoleted properties
      - CookieDomain
      - CookiePath
      - CookieSameSite
      - CookieHttpOnly
      - CookieSecure
    - **New API**
      - `Action<HttpContext, CookieOptions> ConfigureCookieOptions`


### Session

[...]

PR #882, opened 2017-06-29
Add CookieBuilder

natemcmaster:
Addresses #853 

cc @HaoK @javiercn

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/CookieBuilder.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/CookieBuilder.cs
@@ -1,0 +1,93 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Http
+{
+    /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Http.Abstractions/CookieBuilder.cs
```diff
@@ -0,0 +1,94 @@
...
+        public virtual CookieOptions Build(HttpContext context)
+            => Build(context, DateTimeOffset.Now);
+
+        /// <summary>
+        /// Creates the cookie options from the given <paramref name="context"/> with an expiration based on <paramref name="now"/> and <see cref="Expiration"/>.
+        /// </summary>
+        /// <param name="context">The <see cref="HttpContext"/>.</param>
+        /// <param name="now">The time to use as the base for computing <seealso cref="CookieOptions.Expires" /></param>
+        /// <returns>The cookie options.</returns>
+        public virtual CookieOptions Build(HttpContext context, DateTimeOffset now)
```
HaoK: Can we call this expiresFrom or something similar instead of now?

natemcmaster: yeah, that's a little better. Updated.

PR #885, opened 2017-07-03
Add opt out flags for Auth defaults

HaoK:
Addresses https://github.com/aspnet/Security/issues/1287

New specific option that will UseAuthentication() will use to determine which scheme to authenticate

options.AutomaticAuthenticateScheme (falls back to DefaultAuthenticateScheme)

New flags to turn off fallback:

options.EnableDefaultFallback (turn off all fallback logic)
options.EnableDefaultSingleHandlerFallback (turns off just the single handler fallbacks)

cc @Tratcher @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
@@ -61,7 +61,7 @@ namespace Microsoft.AspNetCore.Authentication
             });
 
         /// <summary>
-        /// The scheme that will should be authenticated automatically used to set <see cref="HttpContext.User"/>"/>.
+        /// The scheme that should be authenticated automatically to set <see cref="HttpContext.User"/>"/>.
         /// </summary>
         public string AutomaticAuthenticateScheme { get; set; }
 
@@ -81,12 +81,12 @@ namespace Microsoft.AspNetCore.Authentication
         public string DefaultSignOutScheme { get; set; }
 
         /// <summary>
-        /// Used by as the default scheme by <see cref="IAuthenticationService.ChallengeAsync(HttpContext, string, AuthenticationProperties)"/>.
+        /// Used as the default scheme by <see cref="IAuthenticationService.ChallengeAsync(HttpContext, string, AuthenticationProperties)"/>.
         /// </summary>
         public string DefaultChallengeScheme { get; set; }
 
         /// <summary>
-        /// Used by as the default scheme by <see cref="IAuthenticationService.ForbidAsync(HttpContext, string, AuthenticationProperties)"/>.
+        /// Used as the default scheme by <see cref="IAuthenticationService.ForbidAsync(HttpContext, string, AuthenticationProperties)"/>.
         /// </summary>
         public string DefaultForbidScheme { get; set; }
 
--- a/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
@@ -61,17 +61,22 @@ namespace Microsoft.AspNetCore.Authentication
             });
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
```diff
@@ -61,28 +61,47 @@ public void AddScheme(string name, Action<AuthenticationSchemeBuilder> configure
...
+        /// This is enabled by default.
+        /// </summary>
+        public bool EnableDefaultFallback { get; set; } = true;
+
+        /// <summary>
+        /// If <see cref="EnableDefaultFallback"/> and this are both true, the default <see cref="IAuthenticationSchemeProvider"/> will use a single handler as the default for 
+        /// the Default[Authenticate/SignIn/SignOut/Challenge/Forbid]Schemes.
+        /// This is enabled by default.
+        /// </summary>
+        public bool EnableDefaultSingleHandlerFallback { get; set; } = true;
```
Tratcher: Switches everywhere

Tratcher:
Is the difference that AutomaticAuthenticateScheme would get triggered for every request, where DefaultAuthenticateScheme would only get triggered for [Authorize] requests? This new property doesn't help if it just falls back to DefaultAuthenticateScheme.

HaoK:
This is so you can turn off the automatic middleware without affecting the SignIn() that takes no arguments.

Right now there's no switch that directly affects what scheme the middleware authenticates automatically.

So now you can only disable the middleware from authenticating by turning off EnableDefaultFallback/not setting AutomaticAuthenticateScheme, without affecting how SignIn()/DefaultAuthenticateScheme behave.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
HaoK: Yeah well the tradeoff with adding fairly complex magic defaults, is we need flags to turn off all the magic when people don't want it...

HaoK:
Or more clearly AutomaticAuthenticateScheme is really AuthenticationMiddleware.AuthenticateScheme, but the options live in Abstractions, while the middleware lives in Security.

HaoK:
Alternatively, we could just have an overload `UseAuthentication(string automaticAuthenticateScheme)` instead

PinpointTownes:
Not sure there's a real point having two properties for the same concept :sweat_smile: 

`AuthenticationMiddleware` should probably just call `AuthenticateAsync(null)` and let `AuthenticationService` use `DefaultAuthenticateScheme`.

HaoK:
Well the problem is we overloaded DefaultAuthenticateScheme once we added fallback.  Originally it mapped only to UseAuthentication + SignIn(), now it has side effects of also setting all of the other defaults by default.  Hence more flags/need more control.  So we could have a stand alone DefaultScheme which all of them fallback to instead of having them all fall back to AuthenticateScheme, instead of adding the AutomaticAuthenticateScheme.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
davidfowl: I dunno how I feel about this. Can we talk about the scenario @PinpointTownes has tomorrow. I'd like to better understand why a handler should be able to opt out of the defaults.

HaoK: I think the min bar to unblock @PinpointTownes would be a way for him to opt out the single handler defaulting behavior, he only has a single ASOS handler but that throws on Authenticate, which UseAuthentication() will always call.

See https://github.com/aspnet/Security/issues/1287

davidfowl: What if we just didn't have any defaults unless one was set. That would also fix the bizzare behavior where adding IIS can change your default because another scheme was silently added.

HaoK:
Will replace with new iteration

Issue #886, opened 2017-07-05
Returning a null AuthenticateResult causes a NullReferenceException in AuthenticationService.AuthenticateAsync()


PinpointTownes:
Discovered when porting the aspnet-contrib validation/introspection middleware to the new stack.

https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs#L69

/cc @HaoK @Tratcher

Tratcher:
There's no defined semantic for a null result though, right? Isn't that what AuthenticationResult.NoResult() is for?

PinpointTownes:
Right. FWIW, I reported a similar bug in `AuthenticationHandler` for 1.x a while ago: https://github.com/aspnet/Security/issues/760. You might want to adopt the same fix.

Alternatively, making `AuthenticateResult` a struct would prevent returning `null` and save an allocation everywhere we use it :sweat_smile:

HaoK:
I think our results are slightly too big to be a good value type/struct, with the 4 fields: Exception (8 bytes) + Ticket (8 bytes) + 2 bools (2 bytes) = 18 bytes.  

We can just fix it

Tratcher:
What does size have to do with it? The most significant qualification I see is that it needs to be immutable, which it is.

HaoK:
It will be copied instead of passed by reference everywhere https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/choosing-between-class-and-struct

PinpointTownes:
@HaoK I totally agree that we have to be careful when opting for structs but AFAICT, there are only 3 backing fields in `AuthenticateResult` (corresponding to the `Failure`, `Ticket` and `None` properties).

The last one is basically useless, as the property could be replaced by something like

```csharp
public bool None => Ticket == null && Failure == null;
```

So we'd only have `Failure` + `Ticket` (8 x 2 = 16 bytes on x64 platforms).

PinpointTownes:
Actually, the only problematic thing is that `HandleRequestResult` currently inherits from `AuthenticateResult`, which wouldn't be possible with a struct.

PR #887, opened 2017-07-05
Guard against null

HaoK:
Fixes https://github.com/aspnet/HttpAbstractions/issues/886

@Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
@@ -66,7 +66,7 @@ namespace Microsoft.AspNetCore.Authentication
             }
 
             var result = await handler.AuthenticateAsync();
-            if (result.Succeeded)
+            if (result?.Succeeded)
             {
                 var transformed = await Transform.TransformAsync(result.Principal);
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
```diff
@@ -66,7 +66,7 @@ public AuthenticationService(IAuthenticationSchemeProvider schemes, IAuthenticat
             }
 
             var result = await handler.AuthenticateAsync();
-            if (result.Succeeded)
+            if (result?.Succeeded)
```
PinpointTownes: I bet you $100 it will never compile (`result?.Succeeded` is a `bool?`).

You probably want to use `result != null && result.Succeeded` here.

HaoK: yea I already fixed that :)

PinpointTownes:
BTW, maybe it would have been better to return `AuthenticateResult.NoResult()` when `result` is null. This way, the caller would always get a non-null value.

HaoK:
Do you have a handler that's returning a null AuthenticateResult?

PinpointTownes:
I used to, but I moved to `AuthenticateResult.Skip()` (now `AuthenticateResult.NoResult())` 'cause it was already broken in 1.0: https://github.com/aspnet/Security/issues/760 :sweat_smile:

PR #889, opened 2017-07-06
Add AuthenticationProperties to AuthenticateResult for failures.

Tratcher:
https://github.com/aspnet/Security/issues/1188
See the matching Security PR.

@HaoK how crazy is this? I know you just finished cleaning these up.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticateResult.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticateResult.cs
@@ -31,7 +31,7 @@ namespace Microsoft.AspNetCore.Authentication
         /// <summary>
         /// Additional state values for the authentication session.
         /// </summary>
-        public AuthenticationProperties Properties => Ticket?.Properties;
+        public AuthenticationProperties Properties { get; protected set; }
 
         /// <summary>
         /// Holds failure information from the authentication.
@@ -54,7 +54,7 @@ namespace Microsoft.AspNetCore.Authentication
             {
                 throw new ArgumentNullException(nameof(ticket));
             }
-            return new AuthenticateResult() { Ticket = ticket };
+            return new AuthenticateResult() { Ticket = ticket, Properties = ticket.Properties };
         }
 
         /// <summary>
@@ -79,11 +79,29 @@ namespace Microsoft.AspNetCore.Authentication
         /// <summary>
         /// Indicates that there was a failure during authentication.
         /// </summary>
+        /// <param name="failure">The failure exception.</param>
+        /// <param name="properties">Additional state values for the authentication session.</param>
+        /// <returns>The result.</returns>
+        public static AuthenticateResult Fail(Exception failure, AuthenticationProperties properties)
+        {
+            return new AuthenticateResult() { Failure = failure, Properties = properties };
+        }
+
+        /// <summary>
+        /// Indicates that there was a failure during authentication.
+        /// </summary>
         /// <param name="failureMessage">The failure message.</param>
         /// <returns>The result.</returns>
         public static AuthenticateResult Fail(string failureMessage)
-        {
-            return new AuthenticateResult() { Failure = new Exception(failureMessage) };
-        }
+            => Fail(new Exception(failureMessage));
+
+        /// <summary>
+        /// Indicates that there was a failure during authentication.
+        /// </summary>
+        /// <param name="failureMessage">The failure message.</param>
+        /// <param name="properties">Additional state values for the authentication session.</param>
+        /// <returns>The result.</returns>
+        public static AuthenticateResult Fail(string failureMessage, AuthenticationProperties properties)
+            => Fail(new Exception(failureMessage), properties);
     }
 }


HaoK:
Bleh why don't you create a real first class AuthenticationError class with the things you want, that can contain a Properties if you want, but not just a top level properties

Tratcher:
How's this?

HaoK:
Do we really need properties, or just a dictionary bag?  Or are we storing original properties?

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticateResult.cs
```diff
@@ -31,12 +31,17 @@ public class AuthenticateResult
         /// <summary>
         /// Additional state values for the authentication session.
         /// </summary>
-        public AuthenticationProperties Properties => Ticket?.Properties;
+        public AuthenticationProperties Properties => Ticket?.Properties ?? Error?.Properties;
```
davidfowl: Will both of these ever be set?

HaoK: Ugh, get rid of this fallback, I think Properties should only be Ticket.Properties, make them go Error.Properties if they want the error ones.

HaoK: Or we should just get rid of this property since there's potentially two Properties now

davidfowl: Yea, this looks like a disaster waiting to happen.

Tratcher:
In this case it's the original auth properties from the user's challenge.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticateResult.cs
Tratcher: There will never be both a Ticket and an Error, and the Properties collection is actually the same either way.

HaoK: Then maybe instead of an Error property, we should have ErrorResult derive from AuthenticateResult, have Ticket return null, Properties return the error properties.

HaoK: So basically instead of a new standalone AuthenticationError class, have an ErrorResult : AuthenticateResult instead which sets the Result up properly (no ticket, properties only)

Tratcher: The AuthError gives me a mechanic to solve the next problem (#1178) flowing error details.

HaoK: Sure but the Error is just an exception + properties, can't you do the same with an ErrorResult (which also has Failure exception + Properties)?  Where's the relevant code in the associated PR today?

Tratcher:
*bump* rebased and revived for 2.1. We now have to be more careful that the changes are strictly additive.

HaoK:
I think I'd prefer we added something like a generic bag to AuthenticateResult:
```
   public Dictionary<string, object> CustomData { get } = new Dictionary<string, object>();
```

And then just have Security/RemoteAuthHandler put whatever it needs into the CustomData to flow the extra error info you need.  AuthenticateError looks way too similar to AuthenticateResult...

Tratcher:
AuthProperties aren't custom data, they're auth flow state that we surface for success but not for failures.

Your suggestion applies more to https://github.com/aspnet/Security/issues/1178#issuecomment-313448512 than the current issue.

HaoK:
There's already a Properties on AuthenticateResult

HaoK:
Why can't security just manufacture an error AuthenticateResult with the correct Properties inside?

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticateResult.cs
```diff
@@ -73,17 +78,35 @@ public static AuthenticateResult NoResult()
...
         /// </summary>
-        /// <param name="failureMessage">The failure message.</param>
+        /// <param name="failure">The failure exception.</param>
+        /// <param name="properties">Additional state values for the authentication session.</param>
         /// <returns>The result.</returns>
-        public static AuthenticateResult Fail(string failureMessage)
+        public static AuthenticateResult Fail(Exception failure, AuthenticationProperties properties)
         {
-            return new AuthenticateResult() { Failure = new Exception(failureMessage) };
+            return new AuthenticateResult() { Error = new AuthenticationError(failure, properties) };
```
HaoK: Basically instead of adding an Error, just add a new _errorProperties field to use here, and return that instead of _Error.Properties

Tratcher: They are the same properties. We just don't have a way to surface them for failure scenarios.

Tratcher:
AuthenticateResult.Properties requires an AuthTicket, which requires a ClaimsPrincipal.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticateResult.cs
HaoK: A field can still work, just copy the properties in both Success and Failure instead of calling through to the Ticket/Error.

My main pushback is AuthenticateResult is already meant to handle errors.  We don't need another AuthenticationError class

Tratcher:
Updated, this version's a little simpler.

PR #891, opened 2017-07-07
Add DefaultScheme, remove single fallback

HaoK:
Fixes https://github.com/aspnet/Security/issues/1287

@Tratcher @davidfowl @danroth27

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -170,14 +130,6 @@ public virtual void AddScheme(AuthenticationScheme scheme)
                 {
                     _requestHandlers.Add(scheme);
                 }
-                if (typeof(IAuthenticationSignInHandler).IsAssignableFrom(scheme.HandlerType))
```
PinpointTownes: > Its consistent with IAuthenticationRequestHandler, all AuthenticationHandlers implement Authenticate/Challenge/Forbid. We have one handler that implements SignIn, and two that implement SignOut.

The real difference is that `IAuthenticationRequestHandler` is still used and can be justified by performance reasons (`AuthenticationMiddleware.Invoke` avoids instantiating new handlers that don't implement `IAuthenticationRequestHandler`).

`IAuthenticationSignInHandler` and `IAuthenticationSignOutHandler` are now dead code with no use. I can't find any technical reason to keep them. It makes logging a PITA for implementers, it requires a cast for every request and offers 0 benefit.

PR #898, opened 2017-07-13
Add boolean dispose flag in HttpRequestStreamReader/ResponseWriter (#540)

### LINKED ISSUE #540: Don't make things null in Dispose
src/Microsoft.AspNet.WebUtilities/HttpRequestStreamReader.Dispose nulls out properties. According to @Tratcher https://github.com/aspnet/HttpAbstractions/pull/539#discussion_r49917289 we should instead set up a disposed bool so we don't accidentally get nullrefs.

jkotalik:
Remove nulling out fields in favor of a dispose boolean.
Fixes #540

### DIFF
--- a/src/Microsoft.AspNetCore.WebUtilities/HttpRequestStreamReader.cs
+++ b/src/Microsoft.AspNetCore.WebUtilities/HttpRequestStreamReader.cs
@@ -32,6 +32,7 @@ namespace Microsoft.AspNetCore.WebUtilities
         private int _bytesRead;
 
         private bool _isBlocked;
+        private bool _disposed;
 
         public HttpRequestStreamReader(Stream stream, Encoding encoding)
             : this(stream, encoding, DefaultBufferSize, ArrayPool<byte>.Shared, ArrayPool<char>.Shared)
@@ -50,44 +51,23 @@ namespace Microsoft.AspNetCore.WebUtilities
             ArrayPool<byte> bytePool,
             ArrayPool<char> charPool)
         {
-            if (stream == null)
-            {
-                throw new ArgumentNullException(nameof(stream));
-            }
-
-            if (!stream.CanRead)
-            {
-                throw new ArgumentException(Resources.HttpRequestStreamReader_StreamNotReadable, nameof(stream));
-            }
-
-            if (encoding == null)
-            {
-                throw new ArgumentNullException(nameof(encoding));
-            }
-
-            if (bytePool == null)
-            {
-                throw new ArgumentNullException(nameof(bytePool));
-            }
-
-            if (charPool == null)
-            {
-                throw new ArgumentNullException(nameof(charPool));
-            }
+            _stream = stream ?? throw new ArgumentNullException(nameof(stream));
+            _encoding = encoding ?? throw new ArgumentNullException(nameof(encoding));
+            _bytePool = bytePool ?? throw new ArgumentNullException(nameof(bytePool));
+            _charPool = charPool ?? throw new ArgumentNullException(nameof(charPool));
 
             if (bufferSize <= 0)
             {
                 throw new ArgumentOutOfRangeException(nameof(bufferSize));
             }
+            if (!stream.CanRead)
+            {
+                throw new ArgumentException(Resources.HttpRequestStreamReader_StreamNotReadable, nameof(stream));
+            }
 
-            _stream = stream;
-            _encoding = encoding;
             _byteBufferSize = bufferSize;
-            _bytePool = bytePool;
-            _charPool = charPool;
 
             _decoder = encoding.GetDecoder();
-
             _byteBuffer = _bytePool.Rent(bufferSize);
 
             try
@@ -98,33 +78,24 @@ namespace Microsoft.AspNetCore.WebUtilities
             catch
             {
                 _bytePool.Return(_byteBuffer);
-                _byteBuffer = null;
 
                 if (_charBuffer != null)
                 {
                     _charPool.Return(_charBuffer);
-                    _charBuffer = null;
                 }
+
+                throw;
             }
         }
 
         protected override void Dispose(bool disposing)
         {
-            if (disposing && _stream != null)
+            if (disposing && !_disposed)
             {
-                _stream = null;
+                _disposed = true;
 
-                if (_bytePool != null)
-                {
-                    _bytePool.Return(_byteBuffer);
-                    _byteBuffer = null;
-                }
-
-                if (_charPool != null)
-                {
-                    _charPool.Return(_charBuffer);
-                    _charBuffer = null;
-                }
+                _bytePool.Return(_byteBuffer);
+                _charPool.Return(_charBuffer);
             }
 
             base.Dispose(disposing);
@@ -132,9 +103,9 @@ namespace Microsoft.AspNetCore.WebUtilities
 
         public override int Peek()
         {
-            if (_stream == null)
+            if (_disposed)
             {
-                throw new ObjectDisposedException("stream");
+                throw new ObjectDisposedException(nameof(HttpRequestStreamReader));
             }
 
             if (_charBufferIndex == _charsRead)
@@ -150,9 +121,9 @@ namespace Microsoft.AspNetCore.WebUtilities
 
         public override int Read()
         {
-            if (_stream == null)
+            if (_disposed)
             {
-                throw new ObjectDisposedException("stream");
+                throw new ObjectDisposedException(nameof(HttpRequestStreamReader));
             }
 
             if (_charBufferIndex == _charsRead)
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.WebUtilities/HttpRequestStreamReader.cs
```diff
@@ -107,23 +108,20 @@ public HttpRequestStreamReader(Stream stream, Encoding encoding, int bufferSize)
                 }
             }
         }
-
+
```
moozzyk: (shameless plug) - install this: https://visualstudiogallery.msdn.microsoft.com/bcb3ef8e-a3a4-401e-9e63-fdf00b45268f/view/Reviews and you won't make this mistake again.

Tratcher:
Titles need a description.

Tratcher:
let's do the response stream as well, and check if any other near by components have the same problem.

jkotalik:
🆙  📅  based on what @Tratcher mentioned. Also added tests and modified both files to be consistent.

src/Microsoft.AspNetCore.WebUtilities/HttpRequestStreamReader.cs
```diff
@@ -105,25 +85,23 @@ public HttpRequestStreamReader(Stream stream, Encoding encoding, int bufferSize)
                     _charPool.Return(_charBuffer);
                     _charBuffer = null;
```
Tratcher: remove these nulls as well.

src/Microsoft.AspNetCore.WebUtilities/HttpRequestStreamReader.cs
```diff
@@ -105,25 +85,23 @@ public HttpRequestStreamReader(Stream stream, Encoding encoding, int bufferSize)
...
 
         protected override void Dispose(bool disposing)
         {
-            if (disposing && _stream != null)
+            if (disposing && !_disposed)
             {
-                _stream = null;
-
+                _disposed = true;
                 if (_bytePool != null)
```
Tratcher: why are there null checks on _bytePool and _charPool here? The constructor prevents nulls.

test/Microsoft.AspNetCore.WebUtilities.Tests/HttpRequestStreamReaderTest.cs
```diff
@@ -191,6 +195,43 @@ public static void ReadLine_ReadWithNoNewlines()
...
+        private static IEnumerable<object[]> HttpRequestStreamReaderData()
+        {
+            yield return new object[] { null, Encoding.UTF8, ArrayPool<byte>.Shared, ArrayPool<char>.Shared };
+            yield return new object[] { new MemoryStream(), null, ArrayPool<byte>.Shared, ArrayPool<char>.Shared };
+            yield return new object[] { new MemoryStream(), Encoding.UTF8, null, ArrayPool<char>.Shared };
+            yield return new object[] { new MemoryStream(), Encoding.UTF8, ArrayPool<byte>.Shared, null };
+        }
+
+        [Fact]
+        public static void NegativeBufferSize_ExpectArgumentOutOfRangeException()
```
Tratcher: zero

src/Microsoft.AspNetCore.WebUtilities/HttpRequestStreamReader.cs
jkotalik: I think they were from before when we deliberately null out these streams.

Tratcher: Except _bytePool wasn't being nulled out before.

src/Microsoft.AspNetCore.WebUtilities/HttpRequestStreamReader.cs
```diff
@@ -132,9 +110,9 @@ protected override void Dispose(bool disposing)
 
         public override int Peek()
         {
-            if (_stream == null)
+            if (_disposed)
             {
-                throw new ObjectDisposedException("stream");
+                throw new ObjectDisposedException(nameof(HttpRequestStreamReader));
```
Eilon: I wouldn't pass anything into the ctor here. The type name is not the "object name" - it's pretty useless here and generates noise.

test/Microsoft.AspNetCore.WebUtilities.Tests/HttpRequestStreamReaderTest.cs
```diff
@@ -191,6 +195,43 @@ public static void ReadLine_ReadWithNoNewlines()
             Assert.Null(eol);
         }
 
+        [Theory, MemberData(nameof(HttpRequestStreamReaderData))]
```
Eilon: Consider putting the attributes on separate lines.

src/Microsoft.AspNetCore.WebUtilities/HttpRequestStreamReader.cs
jkotalik: ObjectDisposedException doesn't actually have a ctor with no arguments. We seem to do the nameof(classname) across our repos.

Eilon: Eh, that's fine. I looked through the CoreFX repo and in there it's a mix of `null`, `GetType().Name`, and other stuff like that. So `nameof(this_class_name)` seems fine.

Eilon:
Would it make sense to add tests that check for `ObjectDisposedException`? If that's a "feature" of this type, should we not test it?

jkotalik:
Sounds reasonable, I'll add some in.

jkotalik:
@Tratcher @Eilon 🆙 📅 . I don't believe the async version can be tested in Assert.Throws, but please correct me if I'm wrong.

Tratcher:
There's an Assert.ThrowsAsync

test/Microsoft.AspNetCore.WebUtilities.Tests/HttpRequestStreamReaderTest.cs
```diff
@@ -242,6 +242,17 @@ public static void StreamDisposed_ExpectedObjectDisposedException(Action<HttpReq
             });
         }
 
+        [Fact]
+        public static void StreamDisposed_ExpectObjectDisposedExceptionAsync()
+        {
+            var httpRequestStreamReader = new HttpRequestStreamReader(new MemoryStream(), Encoding.UTF8, 10, ArrayPool<byte>.Shared, ArrayPool<char>.Shared);
+            httpRequestStreamReader.Dispose();
+
+            Assert.ThrowsAsync<ObjectDisposedException>(async () =>
```
Tratcher: await this (and make the method public async Task

jkotalik: The functions below are already async. I don't understand the value in doing this.

jkotalik: `yield return new object[] { new Func<HttpResponseStreamWriter, Task>(async (httpResponseStreamWriter) =>
            {
                await httpResponseStreamWriter.WriteAsync('a');
            })};`

test/Microsoft.AspNetCore.WebUtilities.Tests/HttpRequestStreamReaderTest.cs
```diff
@@ -191,6 +195,64 @@ public static void ReadLine_ReadWithNoNewlines()
             Assert.Null(eol);
         }
 
+        [Theory]
+        [MemberData(nameof(HttpRequestNullData))]
+        public static void NullInputsInConstructor_ExpectArgumentNullException(Stream stream, Encoding encoding, ArrayPool<byte> bytePool, ArrayPool<char> charPool)
```
Tratcher: Why are your test methods static?

test/Microsoft.AspNetCore.WebUtilities.Tests/HttpRequestStreamReaderTest.cs
```diff
@@ -191,6 +195,64 @@ public static void ReadLine_ReadWithNoNewlines()
...
+            httpRequestStreamReader.Dispose();
+
+            Assert.Throws<ObjectDisposedException>(() =>
+            {
+                action(httpRequestStreamReader);
+            });
+        }
+
+        [Fact]
+        public static async void StreamDisposed_ExpectObjectDisposedExceptionAsync()
```
Tratcher: async Task, never async void.

test/Microsoft.AspNetCore.WebUtilities.Tests/HttpRequestStreamReaderTest.cs
jkotalik: All of them were static to begin with, should I change it in another PR?

Tratcher: you can leave it for now, it's just odd.

Eilon: `static` means the test class's ctor won't run. Of course, if there isn't one, that's no problem. As long as the test *runs* and it *passes*, it's probably good 😄

jkotalik:
@Eilon this good to go 😄

Issue #682, opened 2016-08-04
UriHelpers has no null checks, consider throwing if null


jkotalik:
If a param is null, we null ref out rather than throwing an argument exception.
One example:

```
public static string BuildRelative(
            PathString pathBase = new PathString(),
            PathString path = new PathString(),
            QueryString query = new QueryString(),
            FragmentString fragment = new FragmentString())
        {
            string combinePath = (pathBase.HasValue || path.HasValue) ? (pathBase + path).ToString() : "/";
            return combinePath + query.ToString() + fragment.ToString();
        }
```

Though these have defaults, if a client passes in null, it will null ref out.

dougbu:
Why do we have optional parameters in a public API? [Engineering guidelines](https://github.com/aspnet/Home/wiki/Engineering-guidelines#async-method-patterns) mention `default(CancellationToken)` is "one of the few places that we use optional parameters".

Probably can't fix this in 1.1.0 (because back-compat) but do we need a 2.0 issue about splitting this method into multiple overloads?

jkotalik:
Too late to fix the splitting into multiple overloads for 2.0.0, but will add null checks for 2.1

PR #899, opened 2017-07-14
Adds null checks and fixes typo (#682)

jkotalik:
#682 
@Tratcher 
I can factor out some of the null checks into a separate method, however it seems inappropriate.


jkotalik:
@Tratcher Seems to be an issue here for null checks: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Abstractions/PathString.cs#L341
We assume that if both of them have no value, they are equal, which means that a null pathString and an empty path string are the same thing. Thoughts?

davidfowl:
@jkotalik add some tests

Tratcher:
Oh hay, Host/Path/Query/FragmentStrings are all structs. I guess those don't need null checks after all.

Side note: Put the bug number in the description (not the title) so it gets auto linked.

Tratcher:
Note: delete your branch. Also, if your branch is up to date when you push to dev, then the PR is aromatically closed as Merged.

PR #901, opened 2017-07-15
Removes quotes from ContentDispositonHeaderValue Name and FileName

### LINKED ISSUE #446: ContentDispositionHeaderValue doesnt parse FileName and Name correctly

jkotalik:
Fixes #446  using part of #533

### DIFF
--- a/src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
+++ b/src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
@@ -73,6 +73,7 @@ namespace Microsoft.Net.Http.Headers
             set { SetName(NameString, value); }
         }
 
+
         public StringSegment FileName
         {
             get { return GetName(FileNameString); }
@@ -337,7 +338,7 @@ namespace Microsoft.Net.Http.Headers
 
         // Gets a parameter of the given name and attempts to decode it if necessary.
         // Returns null if the parameter is not present or the raw value if the encoding is incorrect.
-        private string GetName(string parameter)
+        private StringSegment GetName(string parameter)
         {
             var nameParameter = NameValueHeaderValue.Find(_parameters, parameter);
             if (nameParameter != null)
@@ -359,7 +360,7 @@ namespace Microsoft.Net.Http.Headers
                     return result;
                 }
                 // May not have been encoded
-                return nameParameter.Value.ToString();
+                return HeaderUtilities.RemoveQuotes(nameParameter.Value);
             }
             return null;
         }
--- a/test/Microsoft.Net.Http.Headers.Tests/ContentDispositionHeaderValueTest.cs
+++ b/test/Microsoft.Net.Http.Headers.Tests/ContentDispositionHeaderValueTest.cs
@@ -152,7 +152,7 @@ namespace Microsoft.Net.Http.Headers
             Assert.Equal(1, contentDisposition.Parameters.Count);
             Assert.Equal("FILENAME", contentDisposition.Parameters.First().Name);
             Assert.Equal("\"=?utf-99?Q?R=mlsZcODTmFtZS5iYXQ=?=\"", contentDisposition.Parameters.First().Value);
-            Assert.Equal("\"=?utf-99?Q?R=mlsZcODTmFtZS5iYXQ=?=\"", contentDisposition.FileName);
+            Assert.Equal("=?utf-99?Q?R=mlsZcODTmFtZS5iYXQ=?=", contentDisposition.FileName);
 
             contentDisposition.FileName = "new_name";
             Assert.Equal("new_name", contentDisposition.FileName);
@@ -558,6 +558,19 @@ namespace Microsoft.Net.Http.Headers
         public void ContentDispositionHeaderValue_ParseInvalid_Throws(string input)
         {
             Assert.Throws<FormatException>(() => ContentDispositionHeaderValue.Parse(input));
+        }
+
+        [Fact]
+        public void HeaderNamesWithQuotes_ExpectNamesToNotHaveQuotes()
+        {
+            var contentDispositionLine = "form-data; name =\"dotnet\"; filename=\"example.png\"";
+            var expectedName = "dotnet";
+            var expectedFileName = "example.png";
+
+            var result = ContentDispositionHeaderValue.Parse(contentDispositionLine);
+
+            Assert.Equal(expectedName, result.Name);
+            Assert.Equal(expectedFileName, result.FileName);
         }
 
         public class ContentDispositionValue


Tratcher:
Do a quick sanity test on the setters while you're here.
https://github.com/aspnet/HttpAbstractions/issues/446#issuecomment-149391050
https://github.com/aspnet/Mvc/blob/9342cb0ab78add8c527edcbc4c0988edc96f01e3/src/Microsoft.AspNet.Mvc.Core/FileResult.cs#L81

See if it should be adding quotes.

Tratcher:
File an MVC issue to remove https://github.com/aspnet/Mvc/blob/eef6c3883a7e27b8387b0925f0b6a88df0a484c5/src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormFileModelBinder.cs#L108

test/Microsoft.Net.Http.Headers.Tests/ContentDispositionHeaderValueTest.cs
```diff
@@ -573,6 +573,21 @@ public void HeaderNamesWithQuotes_ExpectNamesToNotHaveQuote()
...
+        [Fact]
+        public void HeaderNamesWithQuotes_CheckSettersToNotHaveQuotes()
+        {
+            var contentDispositionLine = "form-data; name =\"dotnet\"; filename=\"example.png\"";
+            var expectedName = "hello";
+            var expectedFileName = "example.png";
+
+            var result = ContentDispositionHeaderValue.Parse(contentDispositionLine);
+            result.Name = "hello";
+            result.FileName = "\"example.png\"";
```
Tratcher: I meant what happens when you set a value that requires quotes, like "my file name with spaces.txt"

test/Microsoft.Net.Http.Headers.Tests/ContentDispositionHeaderValueTest.cs
```diff
@@ -560,6 +560,34 @@ public void ContentDispositionHeaderValue_ParseInvalid_Throws(string input)
...
+            var expectedFileName = "example.png";
+
+            var result = ContentDispositionHeaderValue.Parse(contentDispositionLine);
+
+            Assert.Equal(expectedName, result.Name);
+            Assert.Equal(expectedFileName, result.FileName);
+        }
+
+        [Fact]
+        public void HeaderNamesWithQuotes_CheckSettersToNotHaveQuotes()
```
Tratcher: I found that the test I wanted already exists:
https://github.com/aspnet/HttpAbstractions/blob/98da5378bd3383d8590b71e36e4f7a8f76d0e41f/test/Microsoft.Net.Http.Headers.Tests/ContentDispositionHeaderValueTest.cs#L130

jkotalik:
@Tratcher This good to go?

test/Microsoft.Net.Http.Headers.Tests/ContentDispositionHeaderValueTest.cs
Tratcher: You can remove this, it's redundant

src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
```diff
@@ -69,13 +69,14 @@ public IList<NameValueHeaderValue> Parameters
 
         public StringSegment Name
         {
-            get { return GetName(NameString); }
+            get { return HeaderUtilities.RemoveQuotes(GetName(NameString)); }
```
davidfowl: What's the reason we do this every time?

jkotalik: How often will the getter be called @Tratcher ?We can store a local copy after RemoveQuotes has been called once, but it would cause a bit of clutter because only Name and FileName would have these local copies.

Tratcher: The actual data is stored in the Parameters collection and is directly modifiable. This is just a convenience accessor. Any attempt at caching would have to check that the underlying version hadn't changed.

Tratcher: Not often. Don't optimize it prematurely.

davidfowl: This is also turning the StringSegment into a string and back into a StringSegment negating the benefits. GetName should take a StringSegment.

Tratcher: You mean GetName should return a StringSegment. I guess RemoveQuotes could be moved into GetName.

src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
```diff
@@ -337,7 +338,7 @@ private void SetDate(string parameter, DateTimeOffset? date)
 
         // Gets a parameter of the given name and attempts to decode it if necessary.
         // Returns null if the parameter is not present or the raw value if the encoding is incorrect.
-        private string GetName(string parameter)
```
Tratcher: line 363 can remove the ToString() now

src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs
jkotalik: We would need to make GetName accept a StringSegment instead.

davidfowl: It should accept and return a `StringSegment`. At least there'd be no allocations then.

Tratcher: GetName takes a constant key. It doesn't manipulate that value, it uses it for a lookup.

jkotalik: Why aren't they leading with a _?

jkotalik: They are all private const

Tratcher: Constants, not variables

Issue #541, opened 2016-01-20
ResponseCookies.Delete() does not honor CookieOptions such as Secure or HttpOnly


vladimirlapacek:
`Delete()` basically creates new `CookieOptions` that mutes the `Secure` and `HttpOnly` options. We have an HP security scan indicating that the cookie expiration call is insecure as it's not marking the OWIN OIDC nonce cookies expiration as `Secure`. I understand that this is not a real threat, but making the `Delete()` call consistent with the input `CookieOptions` would help people get rid of such problems.

Tratcher:
~~OWIN~~
I'd be curious if any clients key on those fields, but I assume not or we'd have gotten more complaints by now. We can try passing those values through and see what happens with common browsers.

vladimirlapacek:
Thanks for the quick response. Sounds great!

muratg:
Moving this to Backlog as we will be in RC2 ask mode very soon.  If you feel strongly about this issue, please ping me.

PR #905, opened 2017-07-17
Preserve secure and httponly headers in ResponseCookies

jkotalik:
#541 @Tratcher 
I would like to manually verify the before and after of this change in the response cookie, @Tratcher can you help me set that up?


test/Microsoft.AspNetCore.Http.Tests/ResponseCookiesTest.cs
```diff
@@ -6,6 +6,7 @@
 using Microsoft.Extensions.ObjectPool;
 using Microsoft.Net.Http.Headers;
 using Xunit;
+using System;
```
Tratcher: sort System goes first. There's a VS setting for this.

jkotalik:
@Tratcher Tested on a real application and the secure and httponly headers are preserved.

src/Microsoft.AspNetCore.Http/Internal/ResponseCookies.cs
```diff
@@ -129,6 +129,8 @@ public void Delete(string key, CookieOptions options)
                 Path = options.Path,
                 Domain = options.Domain,
                 Expires = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc),
+                Secure = options.Secure,
+                HttpOnly = options.HttpOnly
```
Tratcher: Add the new SameSite option as well.

Issue #732, opened 2016-10-25
`NameValueHeaderValue` and `MediaTypeHeaderValue` mishandle quoted parameter values


dougbu:
Parsing `text/json; version="23"` using `MediaTypeHeaderValue` creates a `NameValueHeaderValue` for `version` with `Value=="\"23\""`. This is incorrect because `text/json; version=23` and `text/json; version="23"` should be 💯% semantically identical.

See aspnet/Mvc#5349 for a similar issue we recently fixed in MVC.

dougbu:
FYI @Tratcher @Eilon

Eilon:
Do you mean it creates Value = `"23"`? It's not clear to me from your formatting what the problem is.

dougbu:
I mean the `Value` includes the surrounding quotes for a quoted parameter value in a media type string. These quotation marks should be ignored when creating the `Value`.

Put another way, the `NameValueHeaderValue` for a media type containing `version=23` will have no quotes in the `Value` while one for a media type containing `version="23"` will. The two `Value`s should be the same.

Eilon:
Got it.

Tratcher:
@jkotalik update: it's ok for NameValueHeaderValue to remove the quotes from Value. But however you do it, you also need to make sure quotes get added to the final wire read ToString output when required. The simplest out for this is to strip the quotes only from the .Value getter. The parser and setter will make sure any values that require quotes will have them (though it won't add them for you, maybe it should to make the round trip scenario easier).

jkotalik:
@Tratcher @dougbu @Eilon I believe this would be a breaking change for 2.1. People may only be checking for the Value that contains quotes. Maybe consider for 3.0?

jkotalik:
I've got a WIP on https://github.com/aspnet/HttpAbstractions/commits/jkotalik/MTHV

I do want to mention that this would be a breaking change, @davidfowl @Eilon @muratg may want to confirm that we want to do this change.

Eilon:
@jkotalik can you describe exactly what the breaking change is w/ an example?

jkotalik:
With the changes I made, every time we access the getter of Value, the returned value would not include any quotes around it. One small problem is that if a customer were to use the properties of NameValueHeaderValue instead of ToString(), like:
```
var nameValue = new NameValueHeaderValue("name", "\"value\"");
var str = nameValue.Name + "=" + nameValue.Value;
```
[...]

PR #913, opened 2017-08-06
NameValueHeaderValue Escaping and Unescaping quoted-string

jkotalik:
Fix for #732 
This is a barebones version on how I wanted to implement GetEncodedValue and SetAndEncodeValue. The general idea is if you construct a NameValueHeaderValue with quotes around it (either through the parser or the constructor), we would remember that for when someone calls SetAndEncodeValue such that round tripping works.

For example:
```C#
var nameValueHeader= new NameValueHeaderValue("test", "\"value\"");
var decodedValue =  nameValueHeader.GetDecodedValue();
nameValueHeader.SetAndEncode(decodedValue);
var result = header.Value // result = "\"value\""
// Should this be expected behavior>
nameValueHeader.SetAndEncode("hello");
result = header.Value // result = "\"hello\""
```

There are still many things that need to be defined:
- Besides removing quotes, what other encoding/decoding should be done?
- If nameValueHeader.SetAndEncode() is called on a new value, should SetAndEncode expect the user to have the string unquoted?
- Are we still considering deprecating HeaderUtilities.RemoveQuotes.
- Perf with all of these new strings?

I want to certain this is the direction we would like to go here, such that it is usable by MVC and does not cause any breaking changes.


src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -142,6 +147,20 @@ public override bool Equals(object obj)
...
+        public StringSegment GetDecodedValue()
+        {
+            return HeaderUtilities.RemoveQuotes(_value);
+        }
+
+        public void SetAndEncodeValue(StringSegment value)
+        {
+            if (_quotedValue)
+            {
+                value = $"\"{value}\"";
```
dougbu: 1. Why check `_quotedValue`? That's about the previous `Value`.
2. This doesn't "encode" anything, especially double-quote characters _within_ the `value`.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -21,6 +21,9 @@ public class NameValueHeaderValue
 
         private StringSegment _name;
         private StringSegment _value;
+        private bool _quotedValue;
```
dougbu: Why save this as a field? `HeaderUtilities.IsQuoted()` doesn't look that expensive.

jkotalik: It a decision between whether we should **always add quotes** around the value in SetAndEncodeValue or do it based on whether the value **originally had quotes** around it when parsed or constructed. I decided on the latter.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
jkotalik: 1. See above comment.
2. I am not sure what other utilities are necessary besides unquoting. I will fix this to handle the double quote scenario.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -601,6 +601,11 @@ public static StringSegment RemoveQuotes(StringSegment input)
             return input;
         }
 
+        public static bool IsQuoted(StringSegment input)
```
Tratcher: Put this in HeaderUtilities and call it from RemoveQuotes.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
Tratcher: don't save it.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -142,6 +147,20 @@ public override bool Equals(object obj)
             }
         }
 
+        public StringSegment GetDecodedValue()
+        {
+            return HeaderUtilities.RemoveQuotes(_value);
```
Tratcher: if !IsQuoted, return _value;
Remove quotes.
if has escaped values, unescape them.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
Tratcher: validate as token
if not token, validate as quoted string
if not quoted string, add quotes and escape reserved chars.
validate as quoted string, throw if still not valid (not everything can be escaped).

test/Microsoft.Net.Http.Headers.Tests/NameValueHeaderValueTest.cs
```diff
@@ -575,6 +575,49 @@ public void TryParseStrictList_WithSomeInvlaidValues_ReturnsFalse()
             Assert.False(NameValueHeaderValue.TryParseStrictList(inputs, out results));
         }
 
+        [Theory]
+        [InlineData("value", "value")]
+        [InlineData("\"value\"", "value")]
+        [InlineData("\"quoted value\"", "quoted value")]
+        public void TestGetDecodedValueThroughConstructor_ReturnTrue(string input, string expected)
```
Eilon: BTW these test method names seem odd to me. Nothing here is returning "True".

Eilon: (Same for several other tests.)

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -21,6 +21,9 @@ public class NameValueHeaderValue
 
         private StringSegment _name;
         private StringSegment _value;
+        private bool _quotedValue;
+        // private StringSegment _encodedValue;
```
Eilon: `//remove`

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
Eilon: BTW even if we were to save it, a better name would be `_isQuotedValue`. Before I noticed this was a `bool`, I thought it stored the quoted value, not whether the value was quoted.

jkotalik:
Will reopen once ready.

test/Microsoft.Net.Http.Headers.Tests/NameValueHeaderValueTest.cs
```diff
@@ -575,6 +575,52 @@ public void TryParseStrictList_WithSomeInvlaidValues_ReturnsFalse()
             Assert.False(NameValueHeaderValue.TryParseStrictList(inputs, out results));
         }
 
+        [Theory]
+        [InlineData("value", "value")]
+        [InlineData("\"value\"", "value")]
+        [InlineData("\"quoted value\"", "quoted value")]
+        [InlineData("\"quoted\\\"valuewithquote\"", "quoted\"valuewithquote")]
+        public void TestGetDecodedValue_ReturnsExpectedValue(string input, string expected)
```
Tratcher: Test?

test/Microsoft.Net.Http.Headers.Tests/NameValueHeaderValueTest.cs
```diff
@@ -575,6 +575,52 @@ public void TryParseStrictList_WithSomeInvlaidValues_ReturnsFalse()
...
+            Assert.Equal(expected, actual);
+        }
+
+        [Theory]
+        [InlineData("value", "value")]
+        [InlineData("\"value\"", "\"value\"")]
+        [InlineData("\"assumes already encoded \\\"\"", "\"assumes already encoded \\\"\"")]
+        [InlineData("unquoted \"value", "\"unquoted \\\"value\"")]
+        [InlineData("value\\morevalues\\evenmorevalues", "\"value\\\\morevalues\\\\evenmorevalues\"")]
+        public void TestSetAndEncodeValue_ReturnsExpectedValue(string input, string expected)
```
Tratcher: EncodesIfRequired

test/Microsoft.Net.Http.Headers.Tests/NameValueHeaderValueTest.cs
```diff
@@ -575,6 +575,52 @@ public void TryParseStrictList_WithSomeInvlaidValues_ReturnsFalse()
...
+        {
+            var header = new NameValueHeaderValue("test");
+            header.SetAndEncodeValue(input);
+
+            var actual = header.Value;
+
+            Assert.Equal(expected, actual);
+        }
+
```
Tratcher: Add a test for setting invalid values?

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -594,13 +595,81 @@ public static string FormatDate(DateTimeOffset dateTime, bool quoted)
...
             }
             return input;
         }
 
+        public static bool IsQuoted(StringSegment input)
+        {
+            return !StringSegment.IsNullOrEmpty(input) && input.Length >= 2 && input[0] == '"' && input[input.Length - 1] == '"';
+        }
+
+        public static StringSegment DecodeEscapeCharacters(StringSegment input)
```
jkotalik: Yeah I was thinking we needed a rename considering "Encode" and "Decode" are too broad.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -594,13 +595,81 @@ public static string FormatDate(DateTimeOffset dateTime, bool quoted)
...
+                    {
+                        throw new FormatException(string.Format(CultureInfo.InvariantCulture, "Trailing slash at the end of string {0}.", input));
+                    }
+
+                    if (input[i + 1] != '\\' && input[i + 1] != '\"')
+                    {
+                        continue;
+                    }
+
+                    stringBuilder.Append(input.Substring(previousSubstring, i - previousSubstring));
```
jkotalik: I wasn't sure of the perf difference, but I'll stick by your word.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -594,13 +595,81 @@ public static string FormatDate(DateTimeOffset dateTime, bool quoted)
...
+                stringBuilder.Append(input.Substring(previousSubstring, input.Length - previousSubstring));
+                return stringBuilder.ToString();
+            }
+            else
+            {
+                return input;
+            }
+        }
+
+        public static StringSegment EncodeEscapeCharacters(StringSegment input)
```
jkotalik: The RFC specifies a quoted-string can have any character after '\', will this need to be updated?

Eilon: If we plan on being spec-compliant, which is a good idea, then we should follow the entire spec (unless there's something totally 🍌🍌🍌 in there ).

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -594,13 +595,64 @@ public static string FormatDate(DateTimeOffset dateTime, bool quoted)
...
+                stringBuilder.Append(input[i]);
+            }
+
+            return stringBuilder.ToString();
+        }
+
+        // See https://tools.ietf.org/html/rfc2616#section-2.2
+        public static StringSegment EscapeAsQuotedString(StringSegment input)
+        {
+            var stringBuilder = new StringBuilder(input.Length * 2);  // worst case the string size doubles as every character must be escaped.
```
dougbu: - Do not allocate if `input` contains only token characters i.e. if the quotes aren't needed.
- Should the StringBuilder instances be pooled somewhere? They can be large and are reusable.
- If not pooling, make the `StringBuilder` as small as possible by calculating how many backslashes must be added before creating the instance.
- `input.Length*2` is two characters too short for the worst case. Don't forget about the surrounding quotes.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -594,13 +595,64 @@ public static string FormatDate(DateTimeOffset dateTime, bool quoted)
...
+            stringBuilder.Append('\"');
+            for (var i = 0; i < input.Length; i++)
+            {
+                if (input[i] == '\\' || input[i] == '\"')
```
dougbu: This matches what paring method do. Does the RFC imply any other characters in a quoted string must be escape?

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -594,13 +595,64 @@ public static string FormatDate(DateTimeOffset dateTime, bool quoted)
...
+                    }
+                }
+                stringBuilder.Append(input[i]);
+            }
+
+            return stringBuilder.ToString();
+        }
+
+        // See https://tools.ietf.org/html/rfc2616#section-2.2
+        public static StringSegment EscapeAsQuotedString(StringSegment input)
```
dougbu: Shouldn't `DecodeQuotedString()` and this method have aligned names e.g. `EncodeAsQuotedString()` here?

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -142,6 +143,32 @@ public override bool Equals(object obj)
...
+        public void SetAndEncodeValue(StringSegment value)
+        {
+            if (StringSegment.IsNullOrEmpty(value) || (GetValueLength(value, 0) == value.Length))
+            {
+                _value = value;
+            }
+            else
+            {
+                var encodedValue = HeaderUtilities.EscapeAsQuotedString(value);
+                if (GetValueLength(encodedValue, 0) != encodedValue.Length)
```
dougbu: This isn't strong enough because `GetValueLength()` does not confirm quoted strings contain only valid characters. Further, the parsing routines explicitly ignore invalid quoted pairs.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -594,13 +595,64 @@ public static string FormatDate(DateTimeOffset dateTime, bool quoted)
...
             return input;
         }
 
+        public static bool IsQuoted(StringSegment input)
+        {
+            return !StringSegment.IsNullOrEmpty(input) && input.Length >= 2 && input[0] == '"' && input[input.Length - 1] == '"';
+        }
+
+        // See https://tools.ietf.org/html/rfc2616#section-2.2
+        public static StringSegment DecodeQuotedString(StringSegment input)
```
dougbu: Oops, that depends on making parsing methods more strict than they are today.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -594,13 +595,64 @@ public static string FormatDate(DateTimeOffset dateTime, bool quoted)
...
+            for (var i = 0; i < input.Length; i++)
+            {
+                if (input[i] == '\\')
+                {
+                    if (input[i + 1] == '\\' || input[i + 1] == '\"')
+                    {
+                        stringBuilder.Append(input[i + 1]);
+                        i++;
+                        continue;
+                    }
```
jkotalik: I don't believe so. If we have four slashes ("\\\\") it will not add any to the string.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
jkotalik: Agreed. @Tratcher wanted EscapeAsQuotedString, I'll change the other method to reflect this.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
jkotalik: Going to discuss with @Tratcher tomorrow.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
jkotalik: We most likely should be sharing this StringBuilders, however I am not sure where to store them. I can make a separate issue for this if necessary.

davidfowl: Look at the InPlaceStringBuilder, example here https://github.com/aspnet/HttpAbstractions/blob/aa158f5d2522949cc9566b060411ead324f6a943/src/Microsoft.AspNetCore.WebUtilities/Base64UrlTextEncoder.cs#L49

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
dougbu: Agreed, I wasn't thinking. That said, you only need to special-case backslashes. Double-quotes shouldn't be mentioned in the condition.

test/Microsoft.Net.Http.Headers.Tests/HeaderUtilitiesTest.cs
```diff
@@ -210,5 +210,29 @@ public void TryParseNonNegativeInt32_Succeeds(string valueString, long expected)
             Assert.True(HeaderUtilities.TryParseNonNegativeInt32(valueString, out value));
             Assert.Equal(expected, value);
         }
+
+        [Theory]
+        [InlineData("\"hello\\\\\"", "hello\\")]
+        [InlineData("\"hello\\\"\"", "hello\"")]
+        [InlineData("\"hello\\\\what\\\\\"", "hello\\what\\")]
+        [InlineData("hello\\\"foo\\\\bar\\\\baz\\\\", "hello\"foo\\bar\\baz\\")]
+        [InlineData("hello\\nfoo", "hello\\nfoo")]
```
dougbu: These two are unquoted strings containing non-token characters. Method should `throw` given such a value.

test/Microsoft.Net.Http.Headers.Tests/HeaderUtilitiesTest.cs
```diff
@@ -210,5 +210,29 @@ public void TryParseNonNegativeInt32_Succeeds(string valueString, long expected)
             Assert.True(HeaderUtilities.TryParseNonNegativeInt32(valueString, out value));
             Assert.Equal(expected, value);
         }
+
+        [Theory]
+        [InlineData("\"hello\\\\\"", "hello\\")]
```
dougbu: Also test invalid quoted pairs e.g. `"\"hello\\\""`. Suspect that'll throw an `IndexOutOfRangeException` right now.

jkotalik:
There is another underlying issue with quoted string parsing (see #923 ) that allows invalid characters in quoted strings. After deciding on that, I will either add illegal character checking here or in HttpRuleParser.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -142,6 +143,30 @@ public override bool Equals(object obj)
...
+        {
+            HeaderUtilities.ThrowIfReadOnly(IsReadOnly);
+            if (StringSegment.IsNullOrEmpty(value) || (GetValueLength(value, 0) == value.Length))
+            {
+                _value = value;
+            }
+            else
+            {
+                var encodedValue = EscapeAsQuotedString(value);
+                CheckValueFormat(encodedValue);
```
jkotalik: Relates to comment, the call to CheckValueFormat should probably throw (which calls HttpRuleParser) should probably throw if there is an invalid character in the string (like '\n').

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -390,13 +415,109 @@ private static void CheckValueFormat(StringSegment value)
...
             }
         }
 
         private static NameValueHeaderValue CreateNameValue()
         {
             return new NameValueHeaderValue();
         }
+
+        // See https://tools.ietf.org/html/rfc7230#section-3.2.6
+        private static StringSegment UnescapeAsQuotedString(StringSegment input)
```
Tratcher: Or did you leave it here because you didn't want to make assumptions about the input? Some assumptions are ok so long as they are well documented.

jkotalik: Yeah I moved it here because of that.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -142,6 +143,30 @@ public override bool Equals(object obj)
...
+            }
+            return UnescapeAsQuotedString(_value);
+        }
+
+        public void SetAndEscapeValue(StringSegment value)
+        {
+            HeaderUtilities.ThrowIfReadOnly(IsReadOnly);
+            if (StringSegment.IsNullOrEmpty(value) || (GetValueLength(value, 0) == value.Length))
+            {
+                _value = value;
```
jkotalik: Setting the Value will throw if there is an error. If the format is invalid, we still need to try calling EscapeAsQuotedString before throwing.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -390,13 +413,109 @@ private static void CheckValueFormat(StringSegment value)
...
+            var stringBuilder = new InplaceStringBuilder(input.Length - backSlashCount);
+
+            for (var i = 0; i < input.Length; i++)
+            {
+                if (input[i] == '\\')
+                {
+                    if (input.Length == i + 1)
+                    {
+                        throw new FormatException(string.Format(CultureInfo.InvariantCulture,
+                            "The header value is invalid, illegal escape character in input: '{0}', position: '{1}'", input, i + 1));
```
Eilon: Change the first comma to a period, a colon, or a semicolon; the comma is incorrect.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -390,13 +413,109 @@ private static void CheckValueFormat(StringSegment value)
...
+        {
+            var numBackSlashes = 0;
+            for (var i = 0; i < input.Length; i++)
+            {
+                if (input[i] == '\\')
+                {
+                    if (input.Length == i + 1)
+                    {
+                        throw new FormatException(string.Format(CultureInfo.InvariantCulture,
+                          "The header value is invalid, illegal escape character in input: '{0}', position: '{1}'", input, i + 1));
```
Eilon: See previous comment.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -6,6 +6,7 @@
 using System.Diagnostics;
 using System.Diagnostics.Contracts;
 using System.Globalization;
+using System.Text;
```
dougbu: No need for this.

src/Microsoft.Net.Http.Headers/NameValueHeaderValue.cs
```diff
@@ -390,13 +413,109 @@ private static void CheckValueFormat(StringSegment value)
...
+                        // Only count escaped backslashes once
+                        i++;
+                    }
+                    numBackSlashes++;
+                }
+            }
+            return numBackSlashes;
+        }
+
+        private static int CountBackslashesForEncodingQuotedString(StringSegment input)
```
dougbu: Name doesn't describe what the method does; actually counts backslashes and double-quotes. Suggest `CountCharactersNeedingBackslashesWhenEncoding()` or some such.

test/Microsoft.Net.Http.Headers.Tests/HeaderUtilitiesTest.cs
```diff
@@ -210,5 +210,29 @@ public void TryParseNonNegativeInt32_Succeeds(string valueString, long expected)
             Assert.True(HeaderUtilities.TryParseNonNegativeInt32(valueString, out value));
             Assert.Equal(expected, value);
         }
+
```
dougbu: thx

test/Microsoft.Net.Http.Headers.Tests/NameValueHeaderValueTest.cs
```diff
@@ -575,6 +575,83 @@ public void TryParseStrictList_WithSomeInvlaidValues_ReturnsFalse()
...
+            var header = new NameValueHeaderValue("test", input);
+
+            var actual = header.GetUnescapedValue();
+
+            Assert.Equal(expected, actual);
+        }
+
+        [Theory]
+        [InlineData("\"hello\\\"")]
+        // See https://github.com/aspnet/HttpAbstractions/issues/923 for why invalid escape characters do not throw.
```
dougbu: Suggest clarifying the comment. Not clear you're talking about the `NameValueHeaderValue` constructor accepting invalid arguments.

test/Microsoft.Net.Http.Headers.Tests/NameValueHeaderValueTest.cs
```diff
@@ -575,6 +575,83 @@ public void TryParseStrictList_WithSomeInvlaidValues_ReturnsFalse()
...
         }
 
+        [Theory]
+        [InlineData("value", "value")]
+        [InlineData("\"value\"", "value")]
+        [InlineData("\"hello\\\\\"", "hello\\")]
+        [InlineData("\"hello\\\"\"", "hello\"")]
+        [InlineData("\"hello\\\"foo\\\\bar\\\\baz\\\\\"", "hello\"foo\\bar\\baz\\")]
+        [InlineData("\"quoted value\"", "quoted value")]
+        [InlineData("\"quoted\\\"valuewithquote\"", "quoted\"valuewithquote")]
```
dougbu: Suggest verbatim strings to make the backslashes easier to see.

jkotalik: It gets ugly with quotes (need to do ""), so I personally prefer these.

test/Microsoft.Net.Http.Headers.Tests/NameValueHeaderValueTest.cs
```diff
@@ -575,6 +575,86 @@ public void TryParseStrictList_WithSomeInvlaidValues_ReturnsFalse()
...
+            header.SetAndEscapeValue(valueHeader);
+
+            var actual = header.Value;
+
+            Assert.Equal(input, actual);
+        }
+        [Theory]
+        [InlineData("val\\nue")]
+        [InlineData("val\\bue")]
+        public void GetAndSetEncodeValueRoundTrip_Fails(string input)
```
Tratcher: "Overescaped values do not round trip"

test/Microsoft.Net.Http.Headers.Tests/NameValueHeaderValueTest.cs
```diff
@@ -575,6 +575,86 @@ public void TryParseStrictList_WithSomeInvlaidValues_ReturnsFalse()
...
+        [InlineData("\"hello\\\"")]
+        // See https://github.com/aspnet/HttpAbstractions/issues/923
+        // Invalid escape characters in the name value do not throw a FormatException
+        // because a quoted-string isn't checked for invalid characters.
+        // Calling GetUnescapedValue() will therefore succeed.
+        public void GetUnescapedValue_ThrowsFormatException(string input)
+        {
+            var header = new NameValueHeaderValue("test", input);
+
+            Assert.Throws<FormatException>(() => header.GetUnescapedValue());
```
Tratcher: This isn't good. Anything that gets past the parser should return something here, even if you have to approximate it because the parser was too lax. Only the setter should throw.

Tratcher:
Let's talk about moving UnescapeAsQuotedString back to HeaderUtilities.

test/Microsoft.Net.Http.Headers.Tests/NameValueHeaderValueTest.cs
jkotalik: I'm assuming we can't change the parser at this point. Should I just return "hello" here as the last escape character was invalid?

Tratcher: Invalid inputs should generally be left alone rather than filtered out. E.g. return the chars `hello\`

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -606,6 +606,109 @@ public static bool IsQuoted(StringSegment input)
...
+                return input;
+            }
+
+            var stringBuilder = new InplaceStringBuilder(input.Length - backSlashCount);
+
+            for (var i = 0; i < input.Length; i++)
+            {
+                if (input[i] == '\\' && i + 1 < input.Length)
+                {
+                    // If there is an escape character as the second to last character in the string,
```
dougbu: If this happens, then what?

jkotalik: The world explodes via dank memes.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -594,13 +594,121 @@ public static string FormatDate(DateTimeOffset dateTime, bool quoted)
...
+            }
+            return numBackSlashes;
+        }
+
+        private static int CountBackslashesForDecodingQuotedString(StringSegment input)
+        {
+            var numBackSlashes = 0;
+            for (var i = 0; i < input.Length; i++)
+            {
+                if (input[i] == '\\' && i + 1 < input.Length)
```
muratg: I recommend changing this to the following to make it clearer (order of the clauses, and moving 1 to the other side of inequality):

`if (i < input.Length - 1 && input[i] == '\\')`

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -594,13 +594,124 @@ public static string FormatDate(DateTimeOffset dateTime, bool quoted)
...
+
+                stringBuilder.Append(input[i]);
+            }
+            stringBuilder.Append('\"');
+            return stringBuilder.ToString();
+        }
+
+        private static int CountCharactersNeedingBackslashesWhenEncoding(StringSegment input)
+        {
+            var numBackSlashes = 0;
```
muratg: nit: name of the var. (it doesn't count number of back slashes)

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -594,13 +594,127 @@ public static string FormatDate(DateTimeOffset dateTime, bool quoted)
...
+        }
+
+        /// <summary>
+        /// Escapes a <see cref="StringSegment"/> as a quoted-string, which is defined by
+        /// <see href="https://tools.ietf.org/html/rfc7230#section-3.2.6">the RFC specification</see>.
+        /// </summary>
+        /// <remarks>
+        /// This will add a backslash before each backslash and quote and add quotes
+        /// around the input. Assumes that the input can be escaped as a quoted-string. Assumes
+        /// that the input does not have quotes around it, as this moethd will add them.
```
Tratcher: method

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -594,13 +594,127 @@ public static string FormatDate(DateTimeOffset dateTime, bool quoted)
...
+                stringBuilder.Append(input[i]);
+            }
+            stringBuilder.Append('\"');
+            return stringBuilder.ToString();
+        }
+
+        private static int CountCharactersNeedingBackslashesWhenEncoding(StringSegment input)
+        {
+            var numberOfCharactersNeedingEscaping = 0;
+            for (var i = 0; i < input.Length; i++)
```
Tratcher: As this is a new code path, we could be stricter here about throwing for control characters 0x00-0x1F and 0x7F, except 0x09 (TAB).

test/Microsoft.Net.Http.Headers.Tests/NameValueHeaderValueTest.cs
Tratcher: Still no tests for setting invalid values.

jkotalik: Right now set accepts all values. Depending on if we add checks for incorrect values, I will add in tests.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
jkotalik: Do we want to have this be responsible for throwing if there are invalid characters or the quoted-string checking? I'm fine with either, but it seems more appropriate during parsing.

jkotalik: Aka https://github.com/aspnet/HttpAbstractions/issues/923 vs checking here.

Tratcher: which check? We can only add validation on the new code path.

Tratcher: I meant to put this comment in EscapeAsQuotedString

Tratcher: #923 is a breaking change we are unlikely to take soon, but we can fix it at least for this new code path.

src/Microsoft.Net.Http.Headers/HeaderUtilities.cs
```diff
@@ -594,13 +594,134 @@ public static string FormatDate(DateTimeOffset dateTime, bool quoted)
...
+        private static int CountAndCheckCharactersNeedingBackslashesWhenEncoding(StringSegment input)
+        {
+            var numberOfCharactersNeedingEscaping = 0;
+            for (var i = 0; i < input.Length; i++)
+            {
+                if (input[i] == '\\' || input[i] == '\"')
+                {
+                    numberOfCharactersNeedingEscaping++;
+                }
+                else if ((input[i] <= 0x1F || input[i] == 0x7F) && input[i] != 0x09)
```
Tratcher: Move this up to EscapeAsQuotedString for visibility. This method only has to count chars that it will escape.

test/Microsoft.Net.Http.Headers.Tests/HeaderUtilitiesTest.cs
```diff
@@ -210,5 +210,76 @@ public void TryParseNonNegativeInt32_Succeeds(string valueString, long expected)
...
+            var actual = HeaderUtilities.EscapeAsQuotedString(input);
+
+            Assert.Equal(expected, actual);
+        }
+
+        [Theory]
+        [InlineData("\n")]
+        [InlineData("\b")]
+        [InlineData("\r")]
+        public void SetAndEscapeValue_ThrowsFormatException(string input)
```
Tratcher: ControlCharacters

Issue #820, opened 2017-04-20
Forward slash '/' character in multipart/form-data boundary breaks header parsing


drayfar:
Hey Team,

We have a project that uploads data from a mobile device to our ASP.NET Core 2.0, MVC 1.1.1 server. We noticed requests seemingly randomly failing, and have narrowed it down to requests that have a forward slash '/' character in the multipart form data boundary. This character is legal in a boundary according to [RFC2046](https://tools.ietf.org/html/rfc2046#page-22). I have been able to repro the issue by changing a unit test in [this PR](https://github.com/drayfar/HttpAbstractions/pull/1). It appears that a token is incorrectly considered invalid in [HttpRuleParser.cs#69](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.Net.Http.Headers/HttpRuleParser.cs#L69). I am not familiar with this codebase, so I am not comfortable making this change myself without some direction. Could you please take a look and advise?

Thanks!
Danny Farmer
@drayfar
Software Engineer 2
Bing Experiences
Microsoft

Tratcher:
While some chars are allowed in the boundary value, they may not be allowed as is in the header. See the note about quotes here : https://tools.ietf.org/html/rfc2046#section-5.1.1. Can you capture a full request in a fiddler trace for context?

drayfar:
[Here is a contrived example](https://github.com/drayfar/HttpAbstractions/blob/dev/boundary.saz) I used in testing. I'll work with our client team to get a real one.

drayfar:
I have verified that this same request works on our older traditional ASP.NET server.

muratg:
@mikaelm12 could you attempt a repro?

jkotalik:
@muratg Feel free to pass this to me, I can look into the repro

jkotalik:
According to @Tratcher comment (https://tools.ietf.org/html/rfc2046#section-5.1.1), it is required to have quotes around the boundary header if there is a '/' in it (or characters like ':', '"', etc. I just tried adding the 
'/' character to the header and it passed all tests. Hopefully adding in quotes around the string will fix your issue.

I'm going to add/modify a few tests for regression.

jkotalik:
https://github.com/aspnet/HttpAbstractions/pull/914

PR #914, opened 2017-08-07
Modifies tests to include characters in quoted strings

jkotalik:
Regression tests for #820 
I didn't see much reason to add separate tests for this change, however I can do so if preferred.


Tratcher:
The unquoted version is more common, that should be the primary test case. Add one or more independent tests for the quoted version.

jkotalik:
🙃 📅

test/Microsoft.AspNetCore.Http.Tests/Features/FormFeatureTests.cs
```diff
@@ -211,6 +225,43 @@ public class FormFeatureTests
         [Theory]
         [InlineData(true)]
         [InlineData(false)]
+        public async Task ReadFormAsync_MultipartWithFileAndQuotedString_ReturnsParsedFormCollection(bool bufferRequest)
```
Tratcher: quoted boundary string

Issue #718, opened 2016-10-04
Add support for facets and structured syntax suffix to MediaType


khellang:
Moved the MediaType parts of issue https://github.com/aspnet/Mvc/issues/5273

It would be nice if `MediaTypeHeaderValue` had support for parsing faceted subtypes and structured syntax suffixes.

See [RFC6838](https://tools.ietf.org/html/rfc6838) for more info.

// @Tratcher

muratg:
Related to #825

muratg:
cc @Eilon we'll be moving some functionality when we do consolidate header parsing logic from MVC (#825)

Eilon:
Please set up a mtg with @Tratcher , @rynowak , @javiercn , @danroth27 , and me.

muratg:
@glennc could you schedule a design meeting?

jkotalik:
@muratg May I go ahead and schedule this?

Tratcher:
MVC wants this, correct?

rynowak:
https://www.youtube.com/watch?v=CeRoi43azds

Eilon:
@rynowak your meme vid has scrombled me.

dougbu:
Put another way&hellip;

Yes, we currently have structured syntax suffix support (see https://tools.ietf.org/html/rfc6839) in our `MediaType` structure.

Eilon:
@jkotalik ỳ̱̻̠̹e̢̼͔s̹̟̜͉̟̱ ̳̟̤͚̹̞͇p͏̠̯͖͖l͇̦e͎a̯̱̬̖s̶̝̗̞͔e̢͈̰͖̣ ̻̳̣͓̘͎͜s̡͇c͔̙͍̪̣h̙̬e͔̞̖̬̺d̦́u̧̩l̡̘͇͙̭̭e̯͍.

jkotalik:
Alright boss.

jkotalik:
@khellang @tuespetre We are currently in the middle of adding this feature to HttpAbstractions. Suffixes will be added in for sure, but we were wondering what are the main use cases for facets that you two are interested in? Currently we are considering providing a property that would return a list of facets. What else, if anything, would you like on top of that?

tuespetre:
[...]

PR #921, opened 2017-08-22
Adds structured syntax suffixes and facets to MediaTypeHeaderValue

jkotalik:
Fix for #718 
This PR does a few things to synchronize MediaType (from MVC) and MediaTypeHeaderValue.
- Adds support for structured syntax suffixes
- Adds support for facets
- Adds Wildcard parameter support (from MediaType).
- Adds doc comments to all public methods and properties.


src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -170,55 +205,141 @@ public IList<NameValueHeaderValue> Parameters
...
+        }
+
+        /// <summary>
+        /// Gets the subtype of the <see cref="MediaTypeHeaderValue"/>, excluding any structured syntax suffix.
+        /// </summary>
+        /// <example>
+        /// For the media type <c>"application/vnd.example+json"</c>, the property gives the value
+        /// <c>"vnd.example"</c>.
+        /// </example>
+        public StringSegment SubTypeWithoutSuffix
```
jkotalik: I know @Tratcher doesn't like the names that MVC used for Suffixes, that is up for discussion.

Tratcher: I can only assume the spec has a better name for this segment.

khellang: I don't really know why you need with and without suffix. Why not just `SubType` and `Suffix`?

jkotalik: We would like to keep the functionality of Subtype the same (currently includes the suffix too). Mvc did the same thing with [MediaType](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Formatters/MediaType.cs#L211-L229)

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -31,18 +40,32 @@ private MediaTypeHeaderValue()
             // Used by the parser to create a new instance of this type.
         }
 
+        /// <summary>
+        /// Initializes a <see cref="MediaTypeHeaderValue"/> instance.
+        /// </summary>
+        /// <param name="mediaType">The <see cref="StringSegment"/> with the media type.
+        /// The media type constructed here must not have any trailing parameters.</param>
```
rynowak: This is a bit obtuse, maybe:
```
A <see cref="StringSegment"/> representation of a media type.  The text provided must be a single media type without parameters.
```

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -170,55 +205,141 @@ public IList<NameValueHeaderValue> Parameters
...
+        /// <example>
+        /// For the media type <c>"application/vnd.example+json"</c>, the property gives the value:
+        /// <c>{"vnd", "example"}</c>
+        /// </example>
+        public IList<StringSegment> Facets
         {
             get
             {
-                return SubType.Equals("*", StringComparison.Ordinal);
+                return SubTypeWithoutSuffix.Split(new char[] { PeriodCharacter } ).ToList();
```
davidfowl: This looks very allocatey.. Cache the array and remove the ToList()

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -31,18 +40,32 @@ private MediaTypeHeaderValue()
...
+        /// The media type constructed here must not have any trailing parameters.</param>
+        /// <param name="quality">The <see cref="double"/> with the quality of the media type.</param>
         public MediaTypeHeaderValue(StringSegment mediaType, double quality)
             : this(mediaType)
         {
             Quality = quality;
         }
 
+        /// <summary>
+        /// Gets the charset parameter of the <see cref="MediaTypeHeaderValue"/> if it has one.
```
rynowak: I realize these are copy-paste but these doc strings are meh. Usually you'd say something like:
```
Gets or sets the value of the charset parameter. Will return <see cref="StringSegment.Empty"> if the charset parameter is not present.
```

I'm not going to comment on all of these, but take care that they say 'gets' or 'gets or sets' as appropriate.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -77,6 +100,9 @@ public StringSegment Charset
             }
         }
 
+        /// <summary>
+        /// Gets the <see cref="System.Text.Encoding"/> of the <see cref="MediaTypeHeaderValue"/> if it has one.
+        /// </summary>
```
rynowak: This is the encoding that *corresponds* to the value of the charset parameter. This should probably point to `Encoding.WebName` to explain that's how they are associated.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
jkotalik: Yeah I was a bit confused by that in MVC. I'll update appropriately.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
jkotalik: The way MTHV is design is to do the parsing on get, rather than on set. I'm assuming you are just referring to caching the new char[] { PeriodCharacter } though.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -170,55 +205,141 @@ public IList<NameValueHeaderValue> Parameters
...
+        /// <example>
+        /// For the media type <c>"application/vnd.example+json"</c>, the property gives the value
+        /// <c>"vnd.example"</c>.
+        /// </example>
+        public StringSegment SubTypeWithoutSuffix
+        {
+            get
+            {
+                var subTypeSuffixLength = _mediaType.LastIndexOf(PlusCharacter);
+                var length = _mediaType.IndexOf(ForwardSlashCharacter) + 1;
```
rynowak: Neither of these are represent the length of anything, they are are the end (exclusive) and start indices of the subtype (respectively).

rynowak: The code is does the right thing, but the name are silly

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -170,55 +205,141 @@ public IList<NameValueHeaderValue> Parameters
...
+        /// <summary>
+        /// Gets whether this <see cref="MediaTypeHeaderValue"/> matches all subtypes, ignoring any structured syntax suffix.
+        /// </summary>
+        /// <example>
+        /// For the media type <c>"application/*+json"</c>, this property is <c>true</c>.
+        /// </example>
+        /// <example>
+        /// For the media type <c>"application/vnd.example+json"</c>, this property is <c>false</c>.
+        /// </example>
+        public bool MatchesAllSubTypesWithoutSuffix =>
```
rynowak: @Tratcher @davidfowl do you object to having this in the HTTP abstractions codebase? I'm not sure any spec says anything about wildcard and 'matching' but it's the common and obvious usage.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -247,56 +368,14 @@ public bool IsSubsetOf(MediaTypeHeaderValue otherMediaType)
...
-            return true;
+            return MatchesType(otherMediaType) &&
+                MatchesSubtype(otherMediaType) &&
+                MatchesParameters(otherMediaType);
         }
 
         /// <summary>
         /// Performs a deep copy of this object and all of it's NameValueHeaderValue sub components,
-        /// while avoiding the cost of revalidating the components.
+        /// while avoiding the cost of re-validating the components.
```
rynowak:
```
Wow
                     such cost

         Many re-validate
```

jkotalik: @davidfowl stood behind me and forced me to change it.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
Tratcher: I don't mind having it in abstractions if it's generally useful, but maybe some of it should move to a ContentNegotiation class to separate the header from the processing.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
jkotalik: Also I am not sure if we can remove the ToList() call.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -31,18 +42,33 @@ private MediaTypeHeaderValue()
...
 
+        /// <summary>
+        /// Initializes a <see cref="MediaTypeHeaderValue"/> instance.
+        /// </summary>
+        /// <param name="mediaType">A <see cref="StringSegment"/> representation of a media type.
+        /// The text provided must be a single media type without parameters. </param>
         public MediaTypeHeaderValue(StringSegment mediaType)
         {
-            CheckMediaTypeFormat(mediaType, "mediaType");
+            CheckMediaTypeFormat(mediaType, MediaTypeString);
```
Tratcher: `nameof(mediaType)`?

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -170,55 +212,148 @@ public IList<NameValueHeaderValue> Parameters
...
+        /// <c>"application/json"</c>.
+        /// </example>
         public StringSegment MediaType
         {
             get { return _mediaType; }
             set
             {
                 HeaderUtilities.ThrowIfReadOnly(IsReadOnly);
-                CheckMediaTypeFormat(value, "value");
+                CheckMediaTypeFormat(value, ValueString);
```
Tratcher: nameof

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -13,8 +13,19 @@ namespace Microsoft.Net.Http.Headers
 {
     public class MediaTypeHeaderValue
```
Tratcher: This needs a comment that links to the spec your following.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -170,55 +212,148 @@ public IList<NameValueHeaderValue> Parameters
...
+        /// For the media type <c>"application/vnd.example+json"</c>, the property gives the value
+        /// <c>"vnd.example"</c>.
+        /// </example>
+        public StringSegment SubTypeWithoutSuffix
         {
             get
             {
-                return MediaType.Equals("*/*", StringComparison.Ordinal);
+                var startOfSuffix = _mediaType.LastIndexOf(PlusCharacter);
+                var startOfSubType = _mediaType.IndexOf(ForwardSlashCharacter) + 1;
```
Tratcher: You can simplify this by calling SubType first.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -170,55 +212,148 @@ public IList<NameValueHeaderValue> Parameters
...
+        /// <example>
+        /// For the media type <c>"application/vnd.example+json"</c>, the property gives the value
+        /// <c>"json"</c>.
+        /// </example>
+        public StringSegment SubTypeSuffix
         {
             get
             {
-                return SubType.Equals("*", StringComparison.Ordinal);
+                var startOfSuffix = _mediaType.LastIndexOf(PlusCharacter);
```
Tratcher: this breaks on `foo+bar/subtype`

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -170,55 +212,148 @@ public IList<NameValueHeaderValue> Parameters
...
+        /// <summary>
+        /// Get a <see cref="IList{T}"/> of facets of the <see cref="MediaTypeHeaderValue"/>. Facets are a
+        /// period separated list of StringSegments in the <see cref="SubTypeWithoutSuffix"/>.
+        /// See <see href="https://tools.ietf.org/html/rfc6838#section-3">The RFC documentation on facets.</see>
+        /// </summary>
+        /// <example>
+        /// For the media type <c>"application/vnd.example+json"</c>, the property gives the value:
+        /// <c>{"vnd", "example"}</c>
+        /// </example>
+        public IList<StringSegment> Facets
```
Tratcher: IEnumerable?

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
javiercn: Is that an scenario? I believe types need to be registered, so anything before the '/' would should be treated as a string, shouldn't it?

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
javiercn: Should we do custom enumerators for this? I'm just bringing it up cuz is one of the things we did for perfz on MVC. @rynowak

Tratcher: StringSegment.Split already does

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
Tratcher: I'm just saying the indexing is wrong. You should be looking for the last index in the SubType.

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -170,55 +212,148 @@ public IList<NameValueHeaderValue> Parameters
...
-        /// SubType = "*"
+        /// Gets the structured syntax suffix of the <see cref="MediaTypeHeaderValue"/> if it has one.
+        /// See <see href="https://tools.ietf.org/html/rfc6838#section-4.8">The RFC documentation on structured syntaxes.</see>
         /// </summary>
-        public bool MatchesAllSubTypes
+        /// <example>
+        /// For the media type <c>"application/vnd.example+json"</c>, the property gives the value
+        /// <c>"json"</c>.
+        /// </example>
+        public StringSegment SubTypeSuffix
```
Tratcher: Can we just call it `Suffix`?

jkotalik: @rynowak @javiercn Complaints?

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
jkotalik: I think it is more useful to have indexing on facets (for example, if we want to check the version on a facet). However, I'll change it for now.

test/Microsoft.Net.Http.Headers.Tests/MediaTypeHeaderValueTest.cs
```diff
@@ -39,6 +40,76 @@ public void Ctor_MediaTypeInvalidFormat_ThrowFormatException()
             AssertFormatException("text/plain;charset=utf-8"); // ctor takes only media-type name, no parameters
         }
 
+        public static IEnumerable<string[]> MediaTypesWithSuffixes
+        {
+            get
```
javiercn: You have the unique opportunity to earn @pranavkm points by doing 

```C#
public static IEnumerable<string[]> MediaTypesWithSuffixes =>
    new List<string []>{ ... }
```

There is also a TheoryData type on Xunit that makes it a bit easier (you save the new[] and it's more strongly typed)

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
```diff
@@ -11,10 +11,22 @@
...
+        private const string CharsetString = "charset";
+        private const string MatchesAllString = "*/*";
+        private const string QualityString = "q";
+        private const string WildcardString = "*";
+
+        private const char ForwardSlashCharacter = '/';
+        private const char PeriodCharacter = '.';
+        private const char PlusCharacter = '+';
+
+        private readonly char[] PeriodCharacterArray = new char[] { PeriodCharacter };
```
JamesNK: Why no static?

davidfowl: Should be static

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
jkotalik: @Tratcher Suffix has been changed, you okay with this name?

Tratcher: I don't have an alternate. Any proposals?

JunTaoLuo: Suffixless? NonSuffixed? Unsuffixed? https://en.wiktionary.org/wiki/suffixless

src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs
jkotalik: @Tratcher Yes or no on separating this into its own class? I'd prefer to just keep these as is, staying in line with MVC. However I can be convinced otherwise.

Tratcher: Is there a line you can draw between "This is the header type and basic information about it" vs "This is content negotiation"? Which members would be pulled out?

jkotalik: I'd say none. I think that these properties provide basic information about the header.

Tratcher: IsSubsetOf gets pretty close, but that's an existing method we can't move.

test/Microsoft.Net.Http.Headers.Tests/MediaTypeHeaderValueTest.cs
```diff
@@ -681,6 +746,74 @@ public void IsSubsetOf_NegativeCases(string mediaType1, string mediaType2)
...
+            Assert.False(result);
+        }
+
+        public static TheoryData<string, List<StringSegment>> MediaTypesWithFacets =>
+                 new TheoryData<string, List<StringSegment>>
+                 {
+                     { "application/vdn.github",
+                         new List<StringSegment>(){ new StringSegment("vdn"), new StringSegment("github") } },
+                     { "application/vdn.github+json",
+                         new List<StringSegment>(){ new StringSegment("vdn"), new StringSegment("github") } },
```
Tratcher: There's an implicit converter from string to StringSegment, you don't need to say `new StringSegment` here do you?

PR #922, opened 2017-08-22
Upgrade to xunit 2.3.0-beta4

natemcmaster:
Upgrade to xunit required some changes due to xunit.analyzers becoming smarter.

FYI - there are some problems in the test code. For now I've suppressed xUnit1026, but this can be code smell, especially when InlineData is used.


test/Microsoft.AspNetCore.Http.Abstractions.Tests/PathStringTests.cs
```diff
@@ -200,7 +198,12 @@ public void StartsWithSegmentsWithRemainder_DoesMatchUsingSpecifiedComparison(st
         [InlineData("encoded", "/http%3a%2f%2f[foo]%3A5000/%", "/http%3a%2f%2f%5Bfoo%5D%3A5000/%25")]
         [InlineData("encoded", "/http%3a%2f%2f[foo]%3A5000/%2", "/http%3a%2f%2f%5Bfoo%5D%3A5000/%252")]
         [InlineData("encoded", "/http%3a%2f%2f[foo]%3A5000/%2F", "/http%3a%2f%2f%5Bfoo%5D%3A5000/%2F")]
-        public void ToUriComponentEscapeCorrectly(string category, string input, string expected)
+        public void ToUriComponentEscapeCorrectly(
+#pragma warning disable xUnit1026 // Theory methods should use all of their parameters
+            string category,
```
natemcmaster: This value is never used in the test method. Any idea what should be done with it? If not, let's remove it.

test/Microsoft.Net.Http.Headers.Tests/CookieHeaderValueTest.cs
```diff
@@ -301,25 +294,31 @@ public void CookieHeaderValue_ParseList_ExcludesInvalidValues(IList<CookieHeader
...
+            var result = CookieHeaderValue.TryParseList(input, out var results);
             Assert.Equal(cookies, results);
             Assert.Equal(cookies?.Count > 0, result);
         }
 
         [Theory]
         [MemberData(nameof(ListWithInvalidCookieHeaderDataSet))]
-        public void CookieHeaderValue_ParseStrictList_ThrowsForAnyInvalidValues(IList<CookieHeaderValue> cookies, string[] input)
+        public void CookieHeaderValue_ParseStrictList_ThrowsForAnyInvalidValues(
+#pragma warning disable xUnit1026 // Theory methods should use all of their parameters
```
jkotalik: @natemcmaster same issue here?

test/Microsoft.AspNetCore.Http.Abstractions.Tests/PathStringTests.cs
Tratcher: Troy added it for no apparent reason. Kill it.
https://github.com/aspnet/HttpAbstractions/commit/150b4708f158c0cdcc2d49f2ec72f3e82c6c64e4#diff-1388310260e4da60f42f87fde7d9f702R198

test/Microsoft.AspNetCore.WebUtilities.Tests/HttpResponseStreamWriterTest.cs
```diff
@@ -521,15 +521,15 @@ protected override void Dispose(bool disposing)
             }
         }
 
-        private static IEnumerable<object[]> HttpResponseStreamWriterData()
+        public static IEnumerable<object[]> HttpResponseStreamWriterData()
```
Tratcher: why are the static public?

test/Microsoft.Net.Http.Headers.Tests/CookieHeaderValueTest.cs
natemcmaster: Yup. Because we re-use `[MemberData(nameof(ListWithInvalidCookieHeaderDataSet))]` in many tests, this is the quickest solution. You have to have a parameter to match each value in the MemberData, but if it's unused in the test method, xunit.analyzers will issue a build warning. 

Going forward it will be better to factor MemberData into separate collections that match their actual usage, if possible.

Tratcher: I assume it's shared MemberData and only some of the tests consume this param?

test/Microsoft.AspNetCore.WebUtilities.Tests/HttpResponseStreamWriterTest.cs
natemcmaster: xunit.analyzers requires it.

natemcmaster: https://github.com/xunit/xunit.analyzers/blob/f1e29fec17f78a535aee5d780104b979abe7e371/src/xunit.analyzers/Descriptors.cs#L94-L97

Issue #924, opened 2017-08-24
FeatureCache is over-checked and over-reset


benaadams:
A simple plantext middleware as below, will check the `Frame` `IFeatureCollection` 7 times and initialize/reset the Context, Resquest and Responses feature caches twice each.
```csharp
public static Task WriteResponse(HttpResponse response)
{
    var payloadLength = _helloWorldPayload.Length;
    response.StatusCode = 200;
    response.ContentType = "text/plain";
    response.ContentLength = payloadLength;
    return response.Body.WriteAsync(_helloWorldPayload, 0, payloadLength);
}
```

![](https://aoa.blob.core.windows.net/aspnet/feature-cache.png)

The extra reset and checks are from initalization; then reinitialization after `RequestServicesFeature` is added to the collection by the host, increasing the `IFeatureCollection`'s version count.

muratg:
@davidfowl thoughts?  (PR is assigned to you too.)

aspnet-hello:
This issue was moved to aspnet/Home#2693

Issue #677, opened 2016-07-22
FormFeature allocates when unused


benaadams:
![form-feature-allocation](https://aoa.blob.core.windows.net/aspnet/form-feature.png)

Resolved by: https://github.com/aspnet/HttpAbstractions/pull/675

muratg:
Investigation

muratg:
Investigation

pakrym:
We don't see difference in performance even when removing FormFeature and CookieFeature allocations completely (https://github.com/aspnet/HttpAbstractions/commit/72d30a5ec0945f60308f235dafa4471069c8d81e).

davidfowl:
You mean throughput. We should be looking at memory as well which this does improve right?

pakrym:
@davidfowl it does, a bit, question is is 88 bytes worth it?

davidfowl:
Every byte counts. If it's easy and non-breaking lets do it. If it complicates things too much then naw.

Tratcher:
So far the solutions have been pretty complicated.

benaadams:
> most real world applications use both cookies and forms.

But not most requests > 50% of webpages ([8M sample](https://css-tricks.com/average-web-page-data-analyzing-8-million-websites/)) have a form and they will be served by a GET request without a form data; as well as all the assets on page (images, fonts, script etc)

A lot of requests and sites are unauthenticated don't have cookies (asset servers) and/or don't care about the cookies (3rd party analytic cookie). There's also vogue not to use cookies but to use [local storage and tokens for auth](https://auth0.com/blog/ten-things-you-should-know-about-tokens-and-cookies/).

So having the features created for every unauthed GET request seems unnecessarily.

> question is is 88 bytes worth it? 

It is when you start densely packing sites on a server with more restricted RAM so the GC becomes more aggressive. You can [create the processes in Jobs](https://blogs.msdn.microsoft.com/maoni/2016/05/23/job-support-in-the-gc/) to simulate this.

Yves57:
Just a question for curiosity: why Factories are not a solution (at least for the less used features)?
I imagine something like `public void Set<TFeature>(Func<TFeature> factory)` in `IFeatureCollection`, that would allow create the feature only the first time it is requested...

pakrym:
@Yves57 that's exactly what I've tried: https://github.com/aspnet/HttpAbstractions/pull/691

benaadams:
[...]

PR #814, opened 2017-04-10
Improve DefaultHttpXxx default paths

benaadams:
* Reduce the number of (non-inlining) calls via interface to features `get_Version`
* Set up `IHttpRequestFeature` and `IHttpResponseFeature` while creating the context

Additionally there are compromises in `DefaultHttpXxx` for pooling; however this can be achieved without these compromises as well as improving the call chains to support devirtualization https://github.com/dotnet/coreclr/issues/9908.

* Implement `PooledHttpContext` though `DefaultHttpContext` rather than via inheritance.
* seal `DefaultHttpXxx` types to support devirtualization of overriden methods
* make `DefaultHttpXxx` private members direct types rather than parent types to support devirtualization
* Remove and don't call virtual methods from `DefaultHttpXxx` .ctors
* Make virtual methods in `DefaultHttpXxx` non-virtual

Non-DefaultHttpXxx can still be archived by inheriting from `HttpXxx` and either implementing via wrapping `DefaultHttpXxx` as per the Pooling example to make use of the feature caches; or by doing something alternative and using a completely different implementation.

PR #925, opened 2017-08-24
Lazy init FeatureReferences

benaadams:
Follow up on https://github.com/aspnet/HttpAbstractions/pull/814 "Improve DefaultHttpXxx default paths"

Fixes https://github.com/aspnet/HttpAbstractions/issues/924 "FeatureCache is over-checked and over-reset"

The Context feature cache is checked and initializes the Response and Request caches at the point one is accessed, so the version check via interface (and sum through collection counts) is shared; and the FeatureReferences structs are only initialized once.

It cuts the version checks from 7 to 5; and the cache initialization to 2x Context, 1x Request, 1x Response, so down 2.

Pre

Post

The early init of the FormFeature triggered the access of the HttpRequest and caused initalization; so I had to move that to be on demand - so it also resolves:

Fixes https://github.com/aspnet/HttpAbstractions/issues/677 "FormFeature allocates when unused"
Fixes https://github.com/aspnet/HttpAbstractions/issues/880 "Don't allocate the FormFeature eagerly per request"

Also

Sealed and made concrete the `DefaultHttpRequest` and `DefaultHttpResponse` class/references for devirtualization and made their initialization/uninitialization simpler.

Someone can always inherit from `HttpContext`/`HttpRequest`/`HttpResponse` if they want a non-Default. 

However it doesn't preclude alternative uses. As an example, reimplemented the `PooledHttpContext` sample by composing with `DefaultHttpContext` rather than via inheritance.

/cc @davidfowl @muratg @vancem @Tratcher @HaoK

### DIFF
--- a/src/Microsoft.AspNetCore.Http/breakingchanges.netcore.json
+++ b/src/Microsoft.AspNetCore.Http/breakingchanges.netcore.json
@@ -8,5 +8,65 @@
       "TypeId": "public class Microsoft.AspNetCore.Http.HttpContextFactory : Microsoft.AspNetCore.Http.IHttpContextFactory",
       "MemberId": "public .ctor(Microsoft.Extensions.ObjectPool.ObjectPoolProvider poolProvider, Microsoft.Extensions.Options.IOptions<Microsoft.AspNetCore.Http.Features.FormOptions> formOptions, Microsoft.AspNetCore.Http.IHttpContextAccessor httpContextAccessor)",
       "Kind": "Removal"
+    },
+    {
+      "TypeId": "public class Microsoft.AspNetCore.Http.DefaultHttpContext : Microsoft.AspNetCore.Http.HttpContext",
+      "MemberId": "protected virtual Microsoft.AspNetCore.Http.Authentication.AuthenticationManager InitializeAuthenticationManager()",
+      "Kind": "Removal"
+    },
+    {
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Http/breakingchanges.netcore.json
```diff
@@ -8,5 +8,65 @@
       "TypeId": "public class Microsoft.AspNetCore.Http.HttpContextFactory : Microsoft.AspNetCore.Http.IHttpContextFactory",
       "MemberId": "public .ctor(Microsoft.Extensions.ObjectPool.ObjectPoolProvider poolProvider, Microsoft.Extensions.Options.IOptions<Microsoft.AspNetCore.Http.Features.FormOptions> formOptions, Microsoft.AspNetCore.Http.IHttpContextAccessor httpContextAccessor)",
       "Kind": "Removal"
+    },
+    {
+      "TypeId": "public class Microsoft.AspNetCore.Http.DefaultHttpContext : Microsoft.AspNetCore.Http.HttpContext",
```
davidfowl: @benaadams We can't make breaking changes in a point release.

davidfowl:
@benaadams This introduces breaking changes. Do you have something that doesn't break APIs?

davidfowl:
@benaadams Closing this out. I think we need to find an approach that doesn't break things. Maybe we write a different HttpContext implementation that does this instead of changing the default.

benaadams:
Non-breaking https://github.com/aspnet/HttpAbstractions/pull/999

PR #928, opened 2017-08-28
Update documentation

smarts:
…to use non-class-specific language since the class-specific language resulted in problems when copying/pasting documentation between classes.


Tratcher:
Good catch. We would prefer to use the (correct) type references in the doc comments though.

jkotalik:
If you could change `<see cref="HeaderDictionary" />` to `<see cref="QueryCollection" />, that would be better.

smarts:
sorry, this slipped off my radar. i'll try to get to it today

muratg:
Thanks @smarts!

smarts:
done @jkotalik  @muratg @Tratcher . not sure what went wrong w/ the build

Tratcher:
The changes look good but it's on the wrong branch, it needs to go to the dev branch. Can you get that fixed? If not then @jkotalik can move it over.

PR #943, opened 2017-09-21
[Prototype] Async RequestOptions

HaoK:
Just playing around with this so far...

Adds an Async configure that is given the HttpContext

            services.ConfigureAsync<TestOptions>((name, o, ctx) => {
                o.Context = ctx; // Compute whatever you want with the HttpContext
                return Task.FromResult(0);
            });

Also short cut to access request options directly off HttpContext:

            var options = await context.GetOptionsAsync<TestOptions>(Options.DefaultName);

This would be purely additive, but give us both async and per request options (with direct access to the context)

Any Thoughts?  @davidfowl @ajcvickers @divega

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Extensions/RequestOptions.cs
+++ b/src/Microsoft.AspNetCore.Http.Extensions/RequestOptions.cs
@@ -1,0 +1,99 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.Http
+{
+    public static class RequestOptionsExtensions {
+        public static IServiceCollection AddRequestOptions(this IServiceCollection services)
+        {
+            services.AddOptions();
+            services.AddScoped(typeof(IRequestOptions<>), typeof(RequestOptions<>));
+            services.AddTransient(typeof(IRequestOptionsFactory<>), typeof(RequestOptionsFactory<>));
+            return services;
+        }
+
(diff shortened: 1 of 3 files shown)


davidfowl:
Why? I'm no type sure I like this first class way of coupling options to the http request. This is about the flighting scenarios right?

src/Microsoft.AspNetCore.Http.Extensions/RequestOptions.cs
```diff
@@ -0,0 +1,100 @@
...
+        public static IServiceCollection AddRequestOptions(this IServiceCollection services)
+        {
+            services.AddOptions();
+            services.AddScoped(typeof(IRequestOptions<>), typeof(RequestOptions<>));
+            services.AddTransient(typeof(IRequestOptionsFactory<>), typeof(RequestOptionsFactory<>));
+            return services;
+        }
+
+        public static IServiceCollection ConfigureAsync<TOptions>(this IServiceCollection services, Func<string, TOptions, HttpContext, Task> configure) where TOptions : class, new()
+            => services.AddSingleton<IConfigureRequestOptions<TOptions>>(new ConfigureRequestOptions<TOptions>(configure));
```
davidfowl: Why isn't this called ConfigureRequestOptins? Why is this even first class?

davidfowl (changes requested):
This looks like a parallel subsystem to options with an additional object (the http request). Why can people Just use the IHttpContextAccessor if they need this

Tratcher:
Who asked for this? What were their scenarios? There has been some discussion around this for Authentication, no?

HaoK:
How to compute options per request is one of the most common questions, but yeah this mostly grew out of the per request authentication PR.  We have also ran into not being able to do async options config in a few places as well, usually related to wanting to configure options by calling into an async method.

So if we had something like this, then we could accomplish switching between default authentication policies per request like this.

```
services.ConfigureRequestOptions<AuthenticationOptions>(options, ctx => options.DefaultPolicy = ComputeDefault(ctx))
```

We'd just need to update the middleware to use request options instead of normal options.  

But yeah this is basically sugar for what would be today: use IOptionsSnapshot + custom IConfigureOptions that takes HttpContextAccessor and uses it to compute options for the request

davidfowl:
Closing this out @HaoK

PR #944, opened 2017-09-29
Add Strict-Transport-Security header name

jkotalik:
To be eventually used by the https and hsts middleware (and other places that need it).


Tratcher:
RFC reference?

Tratcher:
https://tools.ietf.org/html/rfc6797?

Tratcher (approved):

PR #947, opened 2017-10-02
Add helper to register IHttpContextAccessor

### LINKED ISSUE #946: Add a helper to register the IHttpContextAccessor correctly
People mess this up all the time, so lets add a helper that adds the service right the right lifetime.

henkmollema:
Added helper extension method to register `IHttpContextAccessor` with correct lifetime. Added tests as well.

Resolves #946


davidfowl:
Nice @henkmollema !

PR #949, opened 2017-10-03
Add Origin and Referer to GetTypedHeaders

### LINKED ISSUE #948: RequestHeaders does not expose Referer and Origin
Using the `GetTypedHeaders` method.

The `RequestHeaders` class does not expose any property for `Referer` and `Origin` which could be of type `Uri`.

vanillajonathan:
Closes #948


src/Microsoft.AspNetCore.Http.Extensions/HeaderDictionaryTypeExtensions.cs
```diff
@@ -156,6 +156,7 @@ internal static void SetDate(this IHeaderDictionary headers, string name, DateTi
             { typeof(RangeHeaderValue), new Func<string, RangeHeaderValue>(value => { RangeHeaderValue result; return RangeHeaderValue.TryParse(value, out result) ? result : null; }) },
             { typeof(EntityTagHeaderValue), new Func<string, EntityTagHeaderValue>(value => { EntityTagHeaderValue result; return EntityTagHeaderValue.TryParse(value, out result) ? result : null; }) },
             { typeof(DateTimeOffset?), new Func<string, DateTimeOffset?>(value => { DateTimeOffset result; return HeaderUtilities.TryParseDate(value, out result) ? result : (DateTimeOffset?)null; }) },
+            { typeof(Uri), new Func<string, Uri>(value => { Uri result; return Uri.TryCreate(value, UriKind.RelativeOrAbsolute, out result) ? result : null; }) },
```
Tratcher: out var result

src/Microsoft.AspNetCore.Http.Extensions/RequestHeaders.cs
```diff
@@ -249,6 +249,18 @@ public RangeConditionHeaderValue IfRange
...
 
+        public Uri Origin
+        {
+            get
+            {
+                return Headers.Get<Uri>(HeaderNames.Origin);
+            }
+            set
+            {
+                Headers.Set(HeaderNames.Origin, value);
```
Tratcher: We can't use the default Uri.ToString, it is not properly formatted for headers. See how we do it for the Location header.
https://github.com/vanillajonathan/HttpAbstractions/blob/e7814525527f4f6ea7a62beae23f1a9f57b1b038/src/Microsoft.AspNetCore.Http.Extensions/ResponseHeaders.cs#L132-L147

vanillajonathan: Okay, I think I fixed this. I did write the getter and setter in the same way as for the Location header.

src/Microsoft.AspNetCore.Http.Extensions/HeaderDictionaryTypeExtensions.cs
vanillajonathan: I wrote it in the same way as the `DateTimeOffset?` above and the `long?` below.
Do you want me to inline the var 'out' declaration?
I.e. `.TryParseDate(value, out var result)`

Tratcher: yes

src/Microsoft.AspNetCore.Http.Extensions/RequestHeaders.cs
```diff
@@ -249,6 +249,23 @@ public RangeConditionHeaderValue IfRange
             }
         }
 
+        public Uri Origin
```
Tratcher: The spec format is a list. https://tools.ietf.org/html/rfc6454#section-7, and I'm not sure how you'd handle the 'null' case.

src/Microsoft.AspNetCore.Http.Extensions/RequestHeaders.cs
```diff
@@ -260,6 +277,23 @@ public RangeHeaderValue Range
                 Headers.Set(HeaderNames.Range, value);
             }
         }
+        
+        public Uri Referer
```
Tratcher: Referer spec for reference: https://tools.ietf.org/html/rfc7231#section-5.5.2

vanillajonathan: Looks good. Nothing sticking out in the specification.

src/Microsoft.AspNetCore.Http.Extensions/RequestHeaders.cs
vanillajonathan: Oh this is a problem, not only the 'null' case but apparently there can be multiple origins too.

> In some cases, a number of origins contribute to causing the user
   agents to issue an HTTP request.  In those cases, the user agent MAY
   list all the origins in the Origin header field.  For example, if the
   HTTP request was initially issued by one origin but then later
   redirected by another origin, the user agent MAY inform the server
   that two origins were involved in causing the user agent to issue the
   request.

vanillajonathan: Should I remove the addition of the Origin stuff and keep this commit to only have the Referer stuff?

Tratcher: Sure

vanillajonathan: Okay, I removed the Origin part.

src/Microsoft.AspNetCore.Http.Extensions/HeaderDictionaryTypeExtensions.cs
```diff
@@ -156,6 +156,7 @@ internal static void SetDate(this IHeaderDictionary headers, string name, DateTi
             { typeof(RangeHeaderValue), new Func<string, RangeHeaderValue>(value => { RangeHeaderValue result; return RangeHeaderValue.TryParse(value, out result) ? result : null; }) },
             { typeof(EntityTagHeaderValue), new Func<string, EntityTagHeaderValue>(value => { EntityTagHeaderValue result; return EntityTagHeaderValue.TryParse(value, out result) ? result : null; }) },
             { typeof(DateTimeOffset?), new Func<string, DateTimeOffset?>(value => { DateTimeOffset result; return HeaderUtilities.TryParseDate(value, out result) ? result : (DateTimeOffset?)null; }) },
+            { typeof(Uri), new Func<string, Uri>(value => { return Uri.TryCreate(value, UriKind.RelativeOrAbsolute, out var result) ? result : null; }) },
```
Tratcher: This is not longer used, you can revert this file

src/Microsoft.AspNetCore.Http.Extensions/RequestHeaders.cs
```diff
@@ -248,7 +248,7 @@ public RangeConditionHeaderValue IfRange
                 Headers.SetDate(HeaderNames.LastModified, value);
             }
         }
-
+
```
Tratcher: cleanup

src/Microsoft.AspNetCore.Http.Extensions/RequestHeaders.cs
```diff
@@ -260,6 +260,23 @@ public RangeHeaderValue Range
...
+                Uri uri;
+                if (Uri.TryCreate(Headers[HeaderNames.Referer], UriKind.RelativeOrAbsolute, out uri))
+                {
+                    return uri;
+                }
+                return null;
+            }
+            set
+            {
+                Headers.Set(HeaderNames.Referer, value == null ? null : UriHelper.Encode(value));
```
Tratcher: Hmm, the Travis builder says this line doesn't compile. Are you missing a using?

vanillajonathan: Oops. Yeah, it was a missing `using` statement for the `UriHelper.Encode` extension method.

PR #950, opened 2017-10-03
Add Content-Security-Policy

vanillajonathan:


Tratcher:
Spec reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP

Where do you intend to use this?

vanillajonathan:
Convenient to use together with `HttpContext.Response` so as to refer to a constant instead of a hard-coded string. Can be used from inside a controller action or a action filter attribute.

Controller
```cs
public IActionResult Index()
{
    context.Response.Headers.Add(HeaderNames.ContentSecurityPolicy, "default-src 'self'");
    return View();
}
```

Action filter
```cs
public class ContentSecurityPolicyAttribute : IActionFilter
{
    public void OnActionExecuting(ActionExecutingContext context)
    {
        context.Response.Headers.Add(HeaderNames.ContentSecurityPolicy, "default-src 'self'");
    }

    public void OnActionExecuted(ActionExecutedContext context) { }
}
```

Tratcher:
It's needs based. I'm not clear if this one is common enough. @blowdart may be more familiar with it.

'All' is an open ended discussion. This one comes from a spec we wouldn't normally consider.

blowdart:
If it's just header names I have no problem with giving them a "strong type". Implementing it however is a much more interesting discussion :)

vanillajonathan:
Since the class is declared as public, perhaps it ought to contain a more complete collection of headers.
Though perhaps only headers that are REC (Recommendations) not WD (Working Draft).

PR #951, opened 2017-10-03
Add header names for CORS

vanillajonathan:


Tratcher:
Is the objective to replace these constants from CORS? https://github.com/aspnet/CORS/blob/df571655acafb39703935a71a977feec023977d0/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsConstants.cs
Or is there some other place you planned to use these?

Spec reference:
https://www.w3.org/TR/cors/

vanillajonathan:
@Tratcher I was not aware of the existence of the `CorsConstants` class.

I just looked at the `HeaderNames` and noticed that were no headers for CORS. Not sure if they should be in same class or different classes.

Also not sure why one class use `public const string` and the other use `public static readonly string`.

Tratcher:
Filed https://github.com/aspnet/CORS/issues/130 to remove the duplication.

PR #958, opened 2017-11-04
Implement read-only HeaderDictionary

Tratcher:
This is part of https://github.com/aspnet/Hosting/issues/1253.
Note this does not change the contract of IHeaderDictionary, only the concrete HeaderDictionary.
This will enable TestServer to mark its response headers as readonly when the response starts. It will also allow HeaderDictionary to replace HttpSys and IIS's HeaderCollection type.


src/Microsoft.AspNetCore.Http/HeaderDictionary.cs
```diff
@@ -125,25 +130,13 @@ public HeaderDictionary(int capacity)
...
         /// </summary>
         /// <returns>true if the <see cref="HeaderDictionary" /> is in read-only mode; otherwise, false.</returns>
-        public bool IsReadOnly
-        {
-            get
-            {
-                return false;
-            }
-        }
+        public bool IsReadOnly { get; set; }
```
pranavkm: Should this throw if it's been set to ReadOnly? It'd be weird if people reset this bit to work around the exception.

Tratcher: I'm not as concerned about that, they'd have to downcast to get to it in the first place and at that point it becomes pretty obvious they're doing something wrong.

test/Microsoft.AspNetCore.Http.Tests/HeaderDictionaryTests.cs
```diff
@@ -68,5 +68,40 @@ public void EmtpyQuotedHeaderSegmentsAreIgnored()
...
+        [Fact]
+        public void ReadActionsWorkWhenReadOnly()
+        {
+            var headers = new HeaderDictionary(
+                new Dictionary<string, StringValues>(StringComparer.OrdinalIgnoreCase)
+                {
+                    { "Header1", "Value1" }
+                });
+
+            headers.IsReadOnly = true;
```
JunTaoLuo: How are these tests passing when you set IsReadOnly to true?

JunTaoLuo: Nevermind I misread the test name and thought these were write operations.

Issue #955, opened 2017-10-21
Print all registered scheme names in exception message for invalid scheme


davidfowl:
To help with issues like this https://github.com/aspnet/Security/issues/1219#issuecomment-337314228.

/cc @HaoK @Tratcher

Tratcher:
Here: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs

PR #961, opened 2017-11-09
Print available scheme names in exception message for invalid scheme

Tratcher:
#955 A lot of users mess up their auth scheme registrations. This adds more information to the error messages to help them figure out what's wrong.


src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
```diff
@@ -62,7 +63,9 @@ public AuthenticationService(IAuthenticationSchemeProvider schemes, IAuthenticat
             var handler = await Handlers.GetHandlerAsync(context, scheme);
             if (handler == null)
             {
-                throw new InvalidOperationException($"No authentication handler is configured to authenticate for the scheme: {scheme}");
+                var schemes = await GetAllSchemeNames();
```
HaoK: Helper like `throw await CreateMissingHandlerError(scheme)`?

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
```diff
@@ -183,13 +201,61 @@ public AuthenticationService(IAuthenticationSchemeProvider schemes, IAuthenticat
...
+            await signOutHandler.SignOutAsync(properties);
+        }
+
+        private async Task<string> GetAllSchemeNames()
+        {
+            var schemes = string.Join(", ", (await Schemes.GetAllSchemesAsync()).Select(sch => sch.Name));
+
+            if (string.IsNullOrEmpty(schemes))
+            {
+                return "[none]";
```
davidfowl: Naw, I think the message should be different instead of saying none.

Tratcher: "There are no registered authentication schemes."?
Then what?

HaoK: "You need to add your authentication scheme to the IAuthenticationBuilder returned by AddAuthentication()"

HaoK: Or "Did you forget to call AddAuthentication().AddScheme()?"

Tratcher: There's only so much we can cram in here. How about a fwlink to the auth docs instead?

davidfowl: The EF team is notorious for good error message like this. It's *super* valuable.

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
```diff
@@ -183,13 +190,114 @@ public AuthenticationService(IAuthenticationSchemeProvider schemes, IAuthenticat
...
+        {
+            var schemes = await GetAllSignInSchemeNames();
+
+            var mismatchError = $"The authentication handler registered for scheme '{scheme}' is '{handler.GetType().Name}' which cannot be used for SignInAsync. ";
+
+            if (string.IsNullOrEmpty(schemes))
+            {
+                // CookieAuth is the only implementation of sign-in.
+                return new InvalidOperationException(mismatchError
+                    + $"Did you intended to call AddAuthentication().AddCookies(\"Cookies\") and SignInAsync(\"Cookies\",...)?");
```
HaoK: Typo: intended => intend

HaoK: Or should just be consistent and use forget?

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs
```diff
@@ -183,13 +190,114 @@ public AuthenticationService(IAuthenticationSchemeProvider schemes, IAuthenticat
...
+        {
+            var schemes = await GetAllSignOutSchemeNames();
+
+            var mismatchError = $"The authentication handler registered for scheme '{scheme}' is '{handler.GetType().Name}' which cannot be used for {nameof(SignOutAsync)}. ";
+
+            if (string.IsNullOrEmpty(schemes))
+            {
+                // CookieAuth is the most common implementation of sign-out, but OpenIdConnect and WsFederation also support it.
+                return new InvalidOperationException(mismatchError
+                    + $"Did you intended to call AddAuthentication().AddCookies(\"Cookies\") and {nameof(SignOutAsync)}(\"Cookies\",...)?");
```
HaoK: Typo here too

PR #977, opened 2017-12-08
Copy ServerAddress from Kestrel to HttpAbstractions

jkotalik:
BasicMiddleware needs to parse server address, however we don't want to have BasicMiddleware depend on Kestrel. We also can't delete the ServerAddress from Kestrel because it is public https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Kestrel.Core/ServerAddress.cs. We will first create this new class in HttpAbstractions, have Kestrel wrap ServerAddressTemplate with ServerAddress, and then modify HttpSysServer to use this type.


src/Microsoft.AspNetCore.Http/Internal/ServerAddressTemplate.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        public override string ToString()
+        {
+            if (IsUnixPipe)
+            {
+                return Scheme.ToLowerInvariant() + "://" + Host.ToLowerInvariant();
+            }
+            else
+            {
+                return Scheme.ToLowerInvariant() + "://" + Host.ToLowerInvariant() + ":" + Port.ToString(CultureInfo.InvariantCulture);
```
Tratcher: PathBase.

src/Microsoft.AspNetCore.Http/Internal/ServerAddressTemplate.cs
```diff
@@ -0,0 +1,151 @@
...
+            if (other == null)
+            {
+                return false;
+            }
+            return string.Equals(Scheme, other.Scheme, StringComparison.OrdinalIgnoreCase)
+                && string.Equals(Host, other.Host, StringComparison.OrdinalIgnoreCase)
+                && Port == other.Port;
+        }
+
+        public static ServerAddressTemplate FromUrl(string url)
```
Tratcher: Parse

src/Microsoft.AspNetCore.Http/Internal/ServerAddressTemplate.cs
jkotalik: @halter73 was there any reason why the PathBase was omitted from the ToString()?

src/Microsoft.AspNetCore.Http/Internal/ServerAddressTemplate.cs
```diff
@@ -0,0 +1,151 @@
...
+            }
+
+            if (string.IsNullOrEmpty(serverAddress.Host))
+            {
+                throw new FormatException($"Invalid url: '{url}'");
+            }
+
+            if (url[url.Length - 1] == '/')
+            {
+                serverAddress.PathBase = url.Substring(pathDelimiterEnd, url.Length - pathDelimiterEnd - 1);
```
halter73: PathBase only exists for legacy reasons. I think it's better to just throw from this method if a path is present.

The only place PathBase is used in Kestrel is [here](https://github.com/aspnet/KestrelHttpServer/blob/668f8e3b4b4f6e74e7a36433439847575889058e/src/Kestrel.Core/Internal/AddressBinder.cs#L129-L132) just to verify it's null or empty.

halter73:
Why do we want ServerAddress here instead of a shared source package in the Common repo?

src/Microsoft.AspNetCore.Http/Internal/ServerAddressTemplate.cs
Tratcher: HttpSys is going to start using this too, and it requires PathBase.

Tratcher:
An API for this is better than shared source here because it applies generally to anyone consuming the IServerAddressesFeature.

jkotalik:
Also ServerAddress is a public type. We can't change the namespace on it (moving it to a shared package) without breaking people.

jkotalik:
@Tratcher @halter73 and I discussed this and he originally wanted ServerAddress to be an internal type rather than public to begin with. I'm am okay with either option, depending on how reusable we want to make this.

Tratcher:
The travis failures look like issues that need to be addressed.

src/Microsoft.AspNetCore.Http/Internal/ServerAddressTemplate.cs
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        public override bool Equals(object obj)
+        {
+            var other = obj as ServerAddressTemplate;
+            if (other == null)
+            {
+                return false;
+            }
+            return string.Equals(Scheme, other.Scheme, StringComparison.OrdinalIgnoreCase)
```
Tratcher: This also needs PathBase

src/Microsoft.AspNetCore.Http/Internal/ServerAddressTemplate.cs
```diff
@@ -0,0 +1,151 @@
...
+                && string.Equals(Host, other.Host, StringComparison.OrdinalIgnoreCase)
+                && Port == other.Port;
+        }
+
+        public static ServerAddressTemplate ParseFromUrl(string url)
```
danroth27: To make it clearer that these URLs have a template syntax, how about renaming `ParseFromUrl` to just `Parse` and rename the `url` parameter to `template`?

src/Microsoft.AspNetCore.Http/Internal/ServerAddressTemplate.cs
```diff
@@ -0,0 +1,151 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Globalization;
+
+namespace Microsoft.AspNetCore.Http.Internal
```
danroth27: This seems like a generally useful type. Can we move it out of the `Internal` namespace?

Tratcher:
Offline agreement is to call it BindingAddress and leave it pubternal.

jkotalik:
🆙 📅

src/Microsoft.AspNetCore.Http/Internal/RoutingAddress.cs
```diff
@@ -0,0 +1,151 @@
...
+            if (other == null)
+            {
+                return false;
+            }
+            return string.Equals(Scheme, other.Scheme, StringComparison.OrdinalIgnoreCase)
+                && string.Equals(Host, other.Host, StringComparison.OrdinalIgnoreCase)
+                && Port == other.Port;
+        }
+
+        public static RoutingAddress ParseFromUrl(string url)
```
jkotalik: @danroth27 I kept this as ParseFromUrl as we are staying away from the name template. That okay?

Tratcher: Parse(string address)

jkotalik:
@halter73 if we change kestrel to consume this type, it will have to throw if the PathBase is set. I'll make that change once this is in.

src/Microsoft.AspNetCore.Http/Internal/RoutingAddress.cs
```diff
@@ -0,0 +1,151 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Globalization;
+
+namespace Microsoft.AspNetCore.Http.Internal
+{
+    public class RoutingAddress
```
rynowak: It would be dope to avoid the name routing if possible - or make this internal internal . This isn't the routing package and isn't integrated with routing.

Tratcher: I thought it was supposed to be BindingAddress?

jkotalik: Yeah oops.

src/Microsoft.AspNetCore.Http/Internal/BindingAddress.cs
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        public override bool Equals(object obj)
+        {
+            var other = obj as BindingAddress;
+            if (other == null)
+            {
+                return false;
+            }
+            return string.Equals(Scheme, other.Scheme, StringComparison.OrdinalIgnoreCase)
```
Tratcher: PathBase

src/Microsoft.AspNetCore.Http/Internal/BindingAddress.cs
```diff
@@ -0,0 +1,151 @@
...
+            {
+                return Host.StartsWith(Constants.UnixPipeHostPrefix, StringComparison.Ordinal);
+            }
+        }
+
+        public string UnixPipePath
+        {
+            get
+            {
+                Debug.Assert(IsUnixPipe);
```
halter73: Now that this is in another assembly it might be better to just change this to an if/throw.

PR #979, opened 2017-12-18
Consistently initialize HeaderDictionary case insensitive

### LINKED ISSUE #978: HeaderDictionary.Add(string key, StringValues value) initializes without StringComparer.OrdinalIgnoreCase
In the other three places where Store is initialized, it is case-insensitive:

`new Dictionary<string, StringValues>(1, StringComparer.OrdinalIgnoreCase)`

but in the Add(string key, StringValues value) method Store appears to be initialized as case-sensitive:

`new Dictionary<string, StringValues>(1)`

Shouldn't this match the other three places where Store is initialized? Could this lead to unexpected cases where the HeaderDictionary will be case-sensitive?

user1336:
Fixes #978

### DIFF
--- a/src/Microsoft.AspNetCore.Http/HeaderDictionary.cs
+++ b/src/Microsoft.AspNetCore.Http/HeaderDictionary.cs
@@ -32,10 +32,18 @@ namespace Microsoft.AspNetCore.Http
 
         public HeaderDictionary(int capacity)
         {
-            Store = new Dictionary<string, StringValues>(capacity, StringComparer.OrdinalIgnoreCase);
+            EnsureStore(capacity);
         }
 
         private Dictionary<string, StringValues> Store { get; set; }
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Http/HeaderDictionary.cs
```diff
@@ -195,7 +195,7 @@ public void Add(string key, StringValues value)
 
             if (Store == null)
             {
-                Store = new Dictionary<string, StringValues>(1);
+                Store = new Dictionary<string, StringValues>(1, StringComparer.OrdinalIgnoreCase);
```
Tratcher: Nice. Can you take it one step further by adding an EnsureStore() method with this `if` in it? Then all of the creation code paths can call the same method.

Tratcher:
👏

Issue #980, opened 2017-12-21
RequestCookieCollection inconsistent case sensitivity


lundog:
When RequestCookieCollection is instantiated with a capacity it initializes its inner Store dictionary with StringComparer.OrdinalIgnoreCase:
https://github.com/aspnet/HttpAbstractions/blob/2ce2d8b6c5c4d3111f5bbfe642c478f229bd955f/src/Microsoft.AspNetCore.Http/Internal/RequestCookieCollection.cs#L33

However, when instantiated via the static Parse method it initializes its inner Store dictionary without the OrdinalIgnoreCase option:
https://github.com/aspnet/HttpAbstractions/blob/2ce2d8b6c5c4d3111f5bbfe642c478f229bd955f/src/Microsoft.AspNetCore.Http/Internal/RequestCookieCollection.cs#L74

Similar to #978 
It seems as if the Parse method should initialize Store with OrdinalIgnoreCase to be consistent.

Tratcher:
Agreed.

PR #981, opened 2017-12-22
Initialize RequestCookieCollection case insensitive

lundog:
Addresses #980


src/Microsoft.AspNetCore.Http/Internal/RequestCookieCollection.cs
```diff
@@ -71,7 +71,7 @@ public static RequestCookieCollection Parse(IList<string> values)
                     return Empty;
                 }
 
-                var store = new Dictionary<string, string>(cookies.Count);
+                var store = new Dictionary<string, string>(cookies.Count, StringComparer.OrdinalIgnoreCase);
```
Tratcher: This can call the RequestCookieCollection(int capacity) constructor directly.

lundog:
How's that?

PR #984, opened 2018-01-02
AuthenticationProperties code cleanup

Tratcher:
Adding these as part of https://github.com/aspnet/Security/pull/1582/files/9f7f8a857b258593bddf9067d740bb56d02bb90a#diff-9c70897a7c08d33ccecc1707cd3d7cfd

Also in-lining some variables.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
@@ -46,24 +46,8 @@ namespace Microsoft.AspNetCore.Authentication
         /// </summary>
         public bool IsPersistent
         {
-            get { return Items.ContainsKey(IsPersistentKey); }
-            set
-            {
-                if (Items.ContainsKey(IsPersistentKey))
-                {
-                    if (!value)
-                    {
-                        Items.Remove(IsPersistentKey);
-                    }
-                }
-                else
-                {
-                    if (value)
-                    {
-                        Items.Add(IsPersistentKey, string.Empty);
-                    }
-                }
-            }
+            get => GetString(IsPersistentKey) != null;
+            set => SetString(IsPersistentKey, value ? string.Empty : null);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
```diff
@@ -193,5 +188,55 @@ public string RedirectUri
...
+        /// </summary>
+        public string Error
+        {
+            get
+            {
+                return Items.TryGetValue(ErrorKey, out string value) ? value : null;
+            }
+            set
+            {
+                if (value != null)
```
HaoK: add a shared Get/SetHelper(key, value) to consolidate all of these?

HaoK: Maybe its worth adding a handful of tests for AuthenticationProperties since it looks like we don't have any today as well to verify the consolidated behavior

muratg: Who will be setting these values?

Tratcher: Me in the linked PR.

HaoK: Its effectively public API, we set them in auth, but all of the core IAuthenticationHandler apis also accept arbitrary AuthenticationProperties, its basically a settings bag that is passed along with authentication requests/results

HaoK: Actually this brings up a good point, this error logic is only needed for HandleRequestResult/RemoteHandlers to flow error info back right?  These properties should probably go into HandleRequestResult instead of AuthenticationProperties, since it looks like its only the RemoteHandler that would ever use this, and you can just store the error/description from HandleRemoteAuthenticateAsync here https://github.com/aspnet/Security/blob/9f7f8a857b258593bddf9067d740bb56d02bb90a/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs#L56

Tratcher:
Ok, the Security PR no longer needs this PR so I've reduced it to some of the code cleanup suggestions.

PR #986, opened 2018-01-21
Add `HttpRequestRewindUtility`

dougbu:
- aspnet/Home#2684
- makes the `BufferingHelper` methods used in MVC and WebHooks `public`
- no callers of these methods are using them as extension methods

### DIFF
--- a/src/Microsoft.AspNetCore.Http/Extensions/HttpRequestRewindUtility.cs
+++ b/src/Microsoft.AspNetCore.Http/Extensions/HttpRequestRewindUtility.cs
@@ -8,17 +8,25 @@ namespace Microsoft.AspNetCore.Http.Extensions
     public static class HttpRequestRewindUtility
     {
         /// <summary>
-        /// Ensure the <paramref name="request"/> <see cref="HttpRequest.Body"/> can be read multiple times.
+        /// Ensure the <paramref name="request"/> <see cref="HttpRequest.Body"/> can be read multiple times. Normally
+        /// stores request bodies in memory; writes requests larger than 30K bytes to disk.
         /// </summary>
         /// <param name="request">The <see cref="HttpRequest"/> to prepare.</param>
         /// <returns>The updated <paramref name="request"/>.</returns>
+        /// <remarks>
+        /// Temporary files for larger requests are written to the location named in the <c>ASPNETCORE_TEMP</c>
+        /// environment variable, if any. If that environment variable is not defined, these files are written to the
+        /// current user's temporary folder. Files are automatically deleted at the end of the associated requests.
+        /// </remarks>
         public static HttpRequest EnableRewind(HttpRequest request)
         {
             return BufferingHelper.EnableRewind(request);
         }
 
         /// <summary>
-        /// Ensure the <paramref name="request"/> <see cref="HttpRequest.Body"/> can be read multiple times.
+        /// Ensure the <paramref name="request"/> <see cref="HttpRequest.Body"/> can be read multiple times. Normally
+        /// stores request bodies in memory; writes requests larger than <paramref name="bufferThreshold"/> bytes to
+        /// disk.
         /// </summary>
         /// <param name="request">The <see cref="HttpRequest"/> to prepare.</param>
         /// <param name="bufferThreshold">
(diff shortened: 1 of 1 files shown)


dougbu:
Note this PR's proposal is additive. Don't want to break MVC users who upgrade HttpAbstractions.

src/Microsoft.AspNetCore.Http/Extensions/HttpRequestRewindUtility.cs
```diff
@@ -0,0 +1,66 @@
...
+            return BufferingHelper.EnableRewind(request);
+        }
+
+        /// <summary>
+        /// Ensure the <paramref name="request"/> <see cref="HttpRequest.Body"/> can be read multiple times.
+        /// </summary>
+        /// <param name="request">The <see cref="HttpRequest"/> to prepare.</param>
+        /// <param name="bufferThreshold">
+        /// The maximum size in bytes of the in-memory <see cref="System.Buffers.ArrayPool{Byte}"/> used to buffer the
+        /// stream. Larger request bodies are written to disk.
```
kichalla: We write to disk in all the cases of these methods. We should have a similar remark but mentioning the default threshold (and its value) in the method on line no. 15

src/Microsoft.AspNetCore.Http/Extensions/HttpRequestRewindUtility.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <summary>
+        /// Ensure the <paramref name="request"/> <see cref="HttpRequest.Body"/> can be read multiple times.
+        /// </summary>
+        /// <param name="request">The <see cref="HttpRequest"/> to prepare.</param>
+        /// <param name="bufferLimit">
+        /// The maximum size in bytes of the request body. If non-<see langword="null"/>, an attempt to read beyond
+        /// this limit will cause an <see cref="System.IO.IOException"/>.
+        /// </param>
+        /// <returns>The updated <paramref name="request"/>.</returns>
+        public static HttpRequest EnableRewind(HttpRequest request, long? bufferLimit)
```
kichalla: why make it nullable? A user could use the first method in this class for it if they do not want to provide a value for this.

src/Microsoft.AspNetCore.Http/Extensions/HttpRequestRewindUtility.cs
```diff
@@ -0,0 +1,66 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http.Internal;
+
+namespace Microsoft.AspNetCore.Http.Extensions
+{
+    public static class HttpRequestRewindUtility
+    {
+        /// <summary>
+        /// Ensure the <paramref name="request"/> <see cref="HttpRequest.Body"/> can be read multiple times.
```
kichalla: Should there be a comment on these methods or the class about the location where large request bodies would get written?

src/Microsoft.AspNetCore.Http/Extensions/HttpRequestRewindUtility.cs
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http.Internal;
+
+namespace Microsoft.AspNetCore.Http.Extensions
+{
+    public static class HttpRequestRewindUtility
```
kichalla: Should we have a DI service instead? Since it does write to the disk sometimes, it could be surprising for a test. A test should be able to mock this. 
Thinking more this is probably not a problem as during test a request body is usually a seekable stream, in which case we don't write to disk anyways.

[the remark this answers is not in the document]
dougbu: It is an extension method today but nothing uses it that way. That is, both Mvc And WebHooks call it as `BufferingHelper.EnableRewind(...)`. Also our [engineering guidelines](https://github.com/aspnet/Home/wiki/Engineering-guidelines#extension-method-patterns) recommend *not* doing them.

Of course, suggestions for better names much appreciated.

dougbu:
:up::date:

davidfowl:
I wonder if we should just expose something more native on the HttpRequest. System.Web did something a bit messy, it had 2 modes:

- Bufferless (our default)
- Buffered (what we're adding)

These are represented by `HttpRequest.GetBufferlessInputStream()` and `HttpRequest.GetBufferedInputStream()`.Not saying we should copy this but I think it's worth thinking about since we're duplicating these pubternal extension methods and trying to expose something more first class.

dougbu:
> trying to expose something more first class

Hmm, I was working just to expose this feature and avoid MVC / WebHooks relying on a pseudo-internal class. I toyed with `HttpRequest` extension methods in the Microsoft.AspNetCore.Http namespace but didn't think the scenarios warranted that treatment. If you want first class, I can go back to that approach.

The "wonder" in your voice implies you're on the fence. What will help nail this down?

dougbu:
In a quick offline conversation, a few of us decided the following:
- do this work (i.e. fix aspnet/Home#2684) in preview2; not preview1
- expose the methods as `HttpRequest` extension methods named `EnableBuffering(...)`
- consume the new location service aspnet/Hosting#1291 provides using `request.HttpContext.RequestServices`
  - this implies we can lock on approximately the current API
  - also implies we don't have to get aspnet/Hosting#1291 in before this PR; can consume new service in a later change

src/Microsoft.AspNetCore.Http/Extensions/HttpRequestRewindUtility.cs
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http.Internal;
+
+namespace Microsoft.AspNetCore.Http.Extensions
```
Eilon: This is a very odd namespace. It's even more odd that there's 1 type in it. And even if we add a 2nd type, it's still quite odd. Thoughts?

src/Microsoft.AspNetCore.Http/Extensions/HttpRequestRewindUtility.cs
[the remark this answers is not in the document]
Eilon: If it's not going to be an extension method, why return `HttpRequest`? That's presumably for chaining, and that's exactly a case where extension methods are a good fit, no?

Not sure what the naming concerns are here though.

src/Microsoft.AspNetCore.Http/Extensions/HttpRequestRewindUtility.cs
```diff
@@ -0,0 +1,92 @@
...
+{
```
Eilon: Add type-level doc comments too.

src/Microsoft.AspNetCore.Http/Extensions/HttpRequestRewindUtility.cs
dougbu: Won't be in this namespace because I'll be changing to extension methods in the next iteration. Think that addresses all but your class-level doc comments concern.

Hmm, do we document `[EditorBrowsable(EditorBrowsableState.Never)]` classes?

Eilon: Why would it be browsable never? We shouldn't need to apply that to anything...

Eilon: But yes that otherwise sounds good thanks!

davidfowl:
Is this PR updated?

dougbu:
> Is this PR updated?

Not yet. Still have Preview1 things on my plate.

davidfowl:
@dougbu closing this out. When you make the new changes, open a new PR

PR #990, opened 2018-01-29
Added protected ctor to AuthenticationSchemeProvider

khellang:
Closes https://github.com/aspnet/Home/issues/2827

This will let me derive from `AuthenticationSchemeProvider`, pass in `StringComparer.OrdinalIgnoreCase` instead of the default `StringComparer.Ordinal`, and swap the implementation in the container.

It does not affect registration in any way ([`SchemeMap` in `AuthenticationOptions` is still case-sensitive](https://github.com/aspnet/HttpAbstractions/blob/9f64c69a2a1005aed3b044209bbb228210d9a510/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs#L22)), but it will let me do scheme lookups without having to worry about the correct casing. I.e. `github` and `GitHub` will both return the correct scheme.

Personally, I'd prefer to take it one step further by adding a `SchemeNameComparer` option to `AuthenticationOptions` ([as shown in this branch](https://github.com/khellang/HttpAbstractions/compare/scheme-provider-ctor...khellang:scheme-comparer-option)). That would allow me to skip the custom class altogether and just set the property on the options object.

What do you think? Too crazy? 😄 

// @HaoK @davidfowl @Eilon @Tratcher


src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -19,9 +19,17 @@ public class AuthenticationSchemeProvider : IAuthenticationSchemeProvider
         /// </summary>
         /// <param name="options">The <see cref="AuthenticationOptions"/> options.</param>
         public AuthenticationSchemeProvider(IOptions<AuthenticationOptions> options)
+            : this(options, StringComparer.Ordinal)
```
HaoK: Instead of passing the comparer, can we just have the ctor take the map instance instead?

i.e. `protected AuthenticationSchemeProvider(IDictionary<string, AuthenticationScheme> map)`  this seems generally more flexible than a specific comparer behavior.

HaoK:
Too crazy, we don't want to be responsible for making that work everywhere if someone turned this new switch on...

khellang:
> Too crazy, we don't want to be responsible for making that work everywhere if someone turned this new switch on...

I'm assuming that was in response to my option proposal? I guess the ctor is a nice compromise 😄

HaoK:
Yes, the idea is I'm fine making our implementations more extensible (for things like case-insensitivity), I'm not fine with trying to make case-insensitivity first class

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
khellang: I made the change, but I'm not so sure about it... It feels kinda weird passing in a dictionary, just for the purpose of changing the `StringComparer`. I.e. am I supposed to add something to it? Am I *allowed* to add something to it? What happens if I do? What happens if I keep a reference to it and mutate it?

HaoK: Yep its pretty powerful, in this case, you hold the source of truth, so if you mutate that reference, you are directly accessing the underlying dictionary backing the provider.  Seems fine to give developers this power though since its a protected ctor, this is something they are responsible for using 'wisely' :)

Maybe instead of calling it map, name the parameter something like schemeCache to make it clear that this is what will be used to serve the data

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -19,9 +19,17 @@ public class AuthenticationSchemeProvider : IAuthenticationSchemeProvider
         /// </summary>
         /// <param name="options">The <see cref="AuthenticationOptions"/> options.</param>
         public AuthenticationSchemeProvider(IOptions<AuthenticationOptions> options)
+            : this(options, new Dictionary<string, AuthenticationScheme>(StringComparer.Ordinal))
+        {
+        }
+
+        protected AuthenticationSchemeProvider(IOptions<AuthenticationOptions> options, IDictionary<string, AuthenticationScheme> map)
```
HaoK: Add doc comments, which would be a good place to add remarks saying the map is used as is to serve scheme data.

khellang: 👍

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
khellang: Another thing is that the dictionary is accessed within locks inside the `AuthenticationSchemeProvider` (at least in some places). Is that something we should care about?

HaoK: Its probably worth mentioning in the doc comments.  Basically the scheme provider locks to prevent concurrency issues when modifying the list of schemes, so there's definitely issues if someone is back door modifying/accessing the underlying dictionary directly.

HaoK: But basically its the `IAuthenticationSchemeProviders` job to manage the schemes, the dictionary passed in is store backing the provider, so you can probably just say its best to always go through the interface instead of modifying the dictionary.

test/Microsoft.AspNetCore.Authentication.Core.Test/AuthenticationSchemeProviderTests.cs
```diff
@@ -117,6 +119,33 @@ public class AuthenticationSchemeProviderTests
...
+        public async Task LookupUsesProvidedStringComparer()
+        {
+            var services = new ServiceCollection().AddOptions()
+                .AddSingleton<IAuthenticationSchemeProvider, IgnoreCaseSchemeProvider>()
+                .AddAuthenticationCore(o => o.AddScheme<Handler>("signin", "whatever"))
+                .BuildServiceProvider();
+
+            var provider = services.GetRequiredService<IAuthenticationSchemeProvider>();
+
+            Assert.NotNull(await provider.GetSchemeAsync("signin"));
```
Eilon: Would it make sense to be even more explicit and check that each of the returned instances is identical to the "whatever" handler?

test/Microsoft.AspNetCore.Authentication.Core.Test/AuthenticationSchemeProviderTests.cs
```diff
@@ -117,6 +119,33 @@ public class AuthenticationSchemeProviderTests
...
+        [Fact]
+        public void SchemeRegistrationIsCaseSensitive()
+        {
+            var services = new ServiceCollection().AddOptions().AddAuthenticationCore(o =>
+            {
+                o.AddScheme<Handler>("signin", "whatever");
+                o.AddScheme<Handler>("signin", "whatever");
+            }).BuildServiceProvider();
+
+            Assert.Throws<InvalidOperationException>(() => services.GetRequiredService<IAuthenticationSchemeProvider>());
```
HaoK: Is this assertion really testing that scheme registration is case sensitive?

khellang: Yes. It throws an `InvalidOperationException` with the text `"Scheme already exists: signin"` [here](https://github.com/aspnet/HttpAbstractions/blob/9f64c69a2a1005aed3b044209bbb228210d9a510/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs#L127).

HaoK: Ok for tests like this its usually better to do something like this so its clear just from reading the code

```
var error = Assert.Throws<InvalidOperationException>(() => services.GetRequiredService<IAuthenticationSchemeProvider>());
Assert.Contains("Scheme already exists: signin", error.Message);
```

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -15,13 +15,28 @@ namespace Microsoft.AspNetCore.Authentication
...
+        /// Creates an instance of <see cref="AuthenticationSchemeProvider"/>
+        /// using the specified <paramref name="options"/> and <paramref name="schemes"/>.
+        /// </summary>
+        /// <param name="options">The <see cref="AuthenticationOptions"/> options.</param>
+        /// <param name="schemes">The dictionary used to store authentication schemes.</param>
+        protected AuthenticationSchemeProvider(IOptions<AuthenticationOptions> options, IDictionary<string, AuthenticationScheme> schemes)
         {
             _options = options.Value;
 
+            _schemes = schemes;
```
Tratcher: null check

HaoK: Are we allowed to use default parameters?

Since this seems like the cleanest way if so...
```C#
ctor(options, schemes = null) {
  _schemes = schemes ?? new Dictionary()
}
```

HaoK: We actually don't need to have the constructor be protected, since its not like new'ing up one of these does anything without putting it into DI at which point its not like making it protected does anything if they can do that...

khellang: > Are we allowed to use default parameters?

Default parameters in public methods [is a bad idea](https://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/)...

khellang: > We actually don't need to have the constructor be protected

We don't *need to*, but it's much nicer to have just a single public ctor. Besides, why expose more than necessary? Making the ctor public still doesn't help anyone?

Tratcher: ? This is input from the child class, not DI. It should be checked before using it.

test/Microsoft.AspNetCore.Authentication.Core.Test/AuthenticationSchemeProviderTests.cs
khellang: Sure, if asserting on exception messages is your thing.

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
khellang: I added the check, but didn't make the argument optional or make the ctor public. Should I do that as well?

HaoK: @khellang my point is that no-one is newing these up.  The only reason you would ever call the ctor is if you were extending and calling from base...  I actually am basically agree with your point, a single ctor makes this clearer.  But if we can't use defaults, we have to do an overload, and might as well keep that as protected.  Its only worth having the public/protected debate if we can use default values

HaoK: @Tratcher I don't think its unreasonable for passing in null to mean use the default behavior, its an optional argument after all.

HaoK: But yes it needs to be fixed automatically or throw when its null.

Tratcher: Then you would call the other constructor. Don't assign meaning to null.

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -15,13 +15,28 @@ namespace Microsoft.AspNetCore.Authentication
...
+        /// Creates an instance of <see cref="AuthenticationSchemeProvider"/>
+        /// using the specified <paramref name="options"/> and <paramref name="schemes"/>.
+        /// </summary>
+        /// <param name="options">The <see cref="AuthenticationOptions"/> options.</param>
+        /// <param name="schemes">The dictionary used to store authentication schemes.</param>
+        protected AuthenticationSchemeProvider(IOptions<AuthenticationOptions> options, IDictionary<string, AuthenticationScheme> schemes)
         {
             _options = options.Value;
 
+            _schemes = schemes ?? throw new ArgumentNullException(nameof(schemes));
```
HaoK: Personally I like this syntax, but some (including @Eilon I think) are zealously against using this syntax sugar... 

But maybe things have changed, I'd love to use this elsewhere :)

Tratcher: @davidfowl has grudgingly accepted this.

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
HaoK: I'm only arguing this if we can get a single public ctor with a default value, otherwise I agree with you guys :)

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
HaoK: yay, means I can finally make the VS squiggles go away :)

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
Tratcher: Isn't this type DI activated? Not sure how you'd DI activate the misc dictionary.

HaoK: `services.AddSingleton<IAuthenticationSchemeProvider, YourSchemeProvider>()`

HaoK: The derived ctor news up a dictionary and calls base...

HaoK: Or you insert an instance, we don't care is the point...

HaoK: This actually convinces me more that ctor should be public :)  if its good enough for derived classes, it should be good enough for anyone who wants to new one of these up and stick it in DI as an instance.

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
Eilon: Naah I'm cool with it.

HaoK:
Thanks @khellang

khellang:
Thank you! ❤

PR #992, opened 2018-01-31
Adding Http2 headers to HeaderNames class

mikaelm12:
First part of issue: https://github.com/aspnet/Home/issues/2812


Tratcher:
For use when implementing the protocol here: https://github.com/aspnet/KestrelHttpServer/blob/526dfdb332b30fd6042d8ef8d1a2ba99cc654010/src/Kestrel.Core/Internal/Http2/Http2Stream.cs#L55-L66

PR #994, opened 2018-02-04
Allow null value in query string KVP per RFC 3986

robertmiles3:
Fixes https://github.com/aspnet/Home/issues/2679.

This not only fixes the issue specified, but also allows a null KVP value in other methods per [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.4) referenced by @jkotalik.

I added a test as well as enhanced existing tests to validate the null option.


src/Microsoft.AspNetCore.Http.Abstractions/QueryString.cs
```diff
@@ -159,15 +163,21 @@ public static QueryString Create(IEnumerable<KeyValuePair<string, StringValues>>
         {
             var builder = new StringBuilder();
             bool first = true;
+
             foreach (var pair in parameters)
             {
+                // If nothing in this pair.Values, append null value and continue
+                if (pair.Value.Count == 0)
```
davidfowl: StringValues.IsNullOrEmpty

src/Microsoft.AspNetCore.Http.Abstractions/QueryString.cs
```diff
@@ -209,7 +215,10 @@ public QueryString Add(string name, string value)
             builder.Append("&");
             builder.Append(UrlEncoder.Default.Encode(name));
             builder.Append("=");
-            builder.Append(UrlEncoder.Default.Encode(value));
```
campersau: You can also use `AppendKeyValuePair` with `first = false` here.

src/Microsoft.AspNetCore.Http.Abstractions/QueryString.cs
robertmiles3: Good call. Committed.

src/Microsoft.AspNetCore.Http.Abstractions/QueryString.cs
robertmiles3: Agreed. Committed.

src/Microsoft.AspNetCore.Http.Abstractions/QueryString.cs
```diff
@@ -121,12 +121,19 @@ public static QueryString Create(string name, string value)
             {
                 throw new ArgumentNullException(nameof(name));
             }
-            if (value == null)
+
+            return new QueryString($"?{UrlEncoder.Default.Encode(name)}={(string.IsNullOrEmpty(value) ? null : UrlEncoder.Default.Encode(value))}");
```
jkotalik: nit: Add some indentation here, bit hard to read.

src/Microsoft.AspNetCore.Http.Abstractions/QueryString.cs
```diff
@@ -121,12 +121,19 @@ public static QueryString Create(string name, string value)
...
+        }
+
+        private static void AppendKeyValuePair(StringBuilder builder, string key, string value, bool first)
```
jkotalik: Move the private method to the end of the file.

src/Microsoft.AspNetCore.Http.Abstractions/QueryString.cs
```diff
@@ -195,21 +205,14 @@ public QueryString Add(string name, string value)
...
             {
                 return Create(name, value);
             }
 
             var builder = new StringBuilder(Value);
-            builder.Append("&");
-            builder.Append(UrlEncoder.Default.Encode(name));
-            builder.Append("=");
-            builder.Append(UrlEncoder.Default.Encode(value));
+            AppendKeyValuePair(builder, name, value, false);
```
khellang: nit: `first: false`?

robertmiles3: Agreed. Committed, thanks.

src/Microsoft.AspNetCore.Http.Abstractions/QueryString.cs
robertmiles3: I added a few spaces for you. I'm not sure exactly what you had in mind, so if you want further readability tweaks just let me know.

src/Microsoft.AspNetCore.Http.Abstractions/QueryString.cs
```diff
@@ -121,12 +121,8 @@ public static QueryString Create(string name, string value)
             {
                 throw new ArgumentNullException(nameof(name));
             }
-            if (value == null)
-            {
-                throw new ArgumentNullException(nameof(value));
-            }
 
-            return new QueryString($"?{UrlEncoder.Default.Encode(name)}={UrlEncoder.Default.Encode(value)}");
+            return new QueryString($"?{UrlEncoder.Default.Encode(name)}={( string.IsNullOrEmpty(value) ? null : UrlEncoder.Default.Encode(value) )}");
```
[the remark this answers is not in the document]
robertmiles3: Committed. How's that?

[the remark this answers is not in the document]
Tratcher: Better, thanks.

davidfowl:
@robertmiles3 Thanks for the PR, we should really add some benchmarkdotnet microbenchmarks for this code.

robertmiles3:
My first "real" PR in the books for dotnet; that's key. Thanks everyone for the feedback.

PR #995, opened 2018-02-06
Reduce the amount of argument checking by flattening the call graph.

davidfowl:
- Reduce the amount of duplicate argument checks by flattening the call graph a bit


Tratcher:
What are you saving? The null checks on response and file, and maybe one duplicate call to CheckRange? I'm not sure why you bothered.

davidfowl:
@Tratcher I'm planning to use these methods from StaticFiles. I'm just cleaning up the current codez a bit before I make more changes.

PR #996, opened 2018-02-07
Add `HttpRequestRewindExtensions`

dougbu:
- aspnet/Home#2684
- makes the `BufferingHelper` methods used in MVC and WebHooks `public`


dougbu:
This PR replaces #986.

I addressed all comments in that PR and side conversations except those about aspnet/Hosting#1291. Reaction to that PR should include updates to `BufferingHelper` implementation and `HttpRequestRewindExtensions` doc comments. Do we need an issue to track that reaction?

src/Microsoft.AspNetCore.Http/Extensions/HttpRequestRewindExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http.Internal;
+
+namespace Microsoft.AspNetCore.Http
+{
+    /// <summary>
+    /// Extension methods for enabling buffering in an <see cref="HttpRequest"/>.
+    /// </summary>
+    public static class HttpRequestRewindExtensions
```
davidfowl: Why do these methods return the HttpRequest?

dougbu: Works for me either way. Make the call.

I followed the chaining pattern to match `IApplicationBuilder`, `IServiceCollection`, and similar setup extensions. `BufferingHelper` methods also support method chaining.

But, MVC and WebHooks don't require this pattern. Other `HttpRequest` methods and its few extensions aren't following it either.

dougbu: Discussed offline. Will go with `void`.

dougbu:
:up::date:

dougbu:
7b9da556fb

PR #998, opened 2018-02-09
Add HostString.MatchesAny

Tratcher:
https://github.com/aspnet/Home/issues/2863 Add an API useful for matching host names against wildcard patterns. It's static so you can choose not to do HostString's punycode->unicode conversion before matching.

Also reduced a few string allocations by using StringSegments.

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
@@ -2,9 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using Microsoft.AspNetCore.Http.Abstractions;
 using Microsoft.AspNetCore.Http.Internal;
+using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Http
 {
@@ -73,11 +75,9 @@ namespace Microsoft.AspNetCore.Http
         {
             get
             {
-                string host, port;
+                GetParts(_value, out var host, out var port);
 
-                GetParts(out host, out port);
-
-                return host;
+                return host.ToString();
             }
         }
 
@@ -89,17 +89,15 @@ namespace Microsoft.AspNetCore.Http
         {
             get
             {
-                string host, port;
-                int p;
+                GetParts(_value, out var host, out var port);
 
-                GetParts(out host, out port);
-
-                if (string.IsNullOrEmpty(port) || !int.TryParse(port, out p))
+                if (!StringSegment.IsNullOrEmpty(port)
+                    && int.TryParse(port.ToString(), NumberStyles.None, CultureInfo.InvariantCulture, out var p))
                 {
-                    return null;
+                    return p;
                 }
 
-                return p;
+                return null;
             }
         }
 
@@ -135,15 +133,14 @@ namespace Microsoft.AspNetCore.Http
 
             if (i != _value.Length)
             {
-                string host, port;
-                GetParts(out host, out port);
+                GetParts(_value, out var host, out var port);
 
                 var mapping = new IdnMapping();
-                host = mapping.GetAscii(host);
+                var encoded = mapping.GetAscii(host.Buffer, host.Offset, host.Length);
 
-                return string.IsNullOrEmpty(port)
-                    ? host
-                    : string.Concat(host, ":", port);
+                return StringSegment.IsNullOrEmpty(port)
+                    ? encoded
+                    : string.Concat(encoded, ":", port.ToString());
             }
 
             return _value;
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Http.Abstractions.Tests/HostStringTest.cs
```diff
@@ -123,5 +124,43 @@ public void NotEquals_EmptyHostStringAndNonNullHostString()
...
             Assert.NotEqual(hostString, new HostString(string.Empty));
         }
+
+        [Theory]
+        [InlineData("localHost", "localhost")]
+        [InlineData("localHost", "*")] // Any - Used by HttpSys
+        [InlineData("localhost:9090", "localHost")]
+        [InlineData("example.com:443", "example.com")]
+        [InlineData("foo.eXample.com:443", "*.exampLe.com")]
+        [InlineData("f.eXample.com:443", "*.exampLe.com")]
```
halter73: Any test case to verify that "a.b.c.example.com" matches the "*.example.com" filter?

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -208,6 +205,66 @@ public static HostString FromUriComponent(Uri uri)
...
 
+        /// <summary>
+        /// Matches the host portion of a host header value against a list of patterns.
+        /// The host may be the encoded punycode or decoded unicode form so long as the pattern
+        /// uses the same format.
+        /// </summary>
+        /// <param name="value">Host header value with or without a port.</param>
+        /// <param name="patterns">A set of pattern to match, without ports.</param>
+        /// <remarks>
+        /// The port on the given value is ignored. The patterns should not have ports.
```
halter73: Would it be better to reject patterns with ports to make this more clear? A downside is that whatever is translates binding config into to patterns would have to strip the ports.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -270,43 +327,43 @@ public override int GetHashCode()
...
             }
-            else if ((index = _value.IndexOf(']')) >= 0)
+            else if ((index = value.IndexOf(']')) >= 0)
             {
                 // IPv6 in brackets [::1], maybe with port
-                host = _value.Substring(0, index + 1);
+                host = value.Subsegment(0, index + 1);
 
-                if ((index = _value.IndexOf(':', index + 1)) >= 0)
+                if ((index = value.IndexOf(':', index + 1)) >= 0)
```
halter73: I know this logic didn't change, but why call IndexOf here? Shouldn't the ':' immediately follow the ']' if it's present at all.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -270,43 +327,43 @@ public override int GetHashCode()
...
+                host = $"[{value}]";
                 port = null;
             }
             else if (index >= 0)
             {
                 // Has a port
-                host = _value.Substring(0, index);
-                port = _value.Substring(index + 1);
+                host = value.Subsegment(0, index);
+                port = value.Subsegment(index + 1);
```
halter73: Do you think we should verify the port is actually a number? Having a public MatchesAny API that strips ports that might not actually be ports seems fragile.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
halter73: I misunderstood. Failing to ever match is OK.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
```diff
@@ -208,6 +205,66 @@ public static HostString FromUriComponent(Uri uri)
...
+        /// <param name="patterns">A set of pattern to match, without ports.</param>
+        /// <remarks>
+        /// The port on the given value is ignored. The patterns should not have ports.
+        /// The patterns may be exact matches like "example.com", a top level wildcard "*"
+        /// that matches all hosts, or a subdomain wildcard like "*.example.com" that matches
+        /// "abc.example.com:443" but not "example.com:443".
+        /// Matching is case insensitive.
+        /// </remarks>
+        /// <returns></returns>
+        public static bool MatchesAny(string value, IList<string> patterns)
```
davidfowl: Any reason these aren't StringSegments?

test/Microsoft.AspNetCore.Http.Abstractions.Tests/HostStringTest.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Threading.Tasks;
```
davidfowl: Why did you have to add this?

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
Tratcher: Neither are expected to be pre processed, only raw input.

davidfowl: This should be a StringSegment. This is a generic helper API that should be able to work on slices of strings. 

Ideally it would be `ReadOnlySpan<char>` but we haven't taken on that dependency widely as yet.

src/Microsoft.AspNetCore.Http.Abstractions/HostString.cs
Tratcher: The format is validated elsewhere.

halter73: That would be a more compelling argument if this wasn't called as part of a public API that takes arbitrary strings and doesn't validate the port is a number. Unless you're saying MatchesAny does this somewhere I'm not seeing.

Tratcher: And if it's not numeric, then what? Include it in the host?

halter73: Either that or throw a FormatException.

PR #999, opened 2018-02-10
Reduce FeatureReference checking, lazy init FormFeature

### LINKED ISSUE #677: FormFeature allocates when unused

### LINKED ISSUE #880: Don't allocate the FormFeature eagerly per request
Today, in order to pass the form options to the `FormFeature`, we eagerly allocate it here https://github.com/aspnet/HttpAbstractions/blob/163836fe1f31dd1d2b288659c2dcaf1d827da3c4/src/Microsoft.AspNetCore.Http/HttpContextFactory.cs#L49. We should find a creative way to avoid doing this so eagerly in the common cases. Here are some options:

- Invent a new way to pass things to features lazily (hard, we tried this and ended up with crazy things)
- Only new up the `FormOptions` if they were overridden. We'd need a way to detect if the `FormOptions` were actually set. This is doable but could get ugly as there's no way to detect if options were configured. /cc @HaoK 
- Only set the `FormFeature` if the request is a POST and `HasFormContentType`. Gross, but would be more pay for play than what we have today.

/cc @vancem @Tratcher

### LINKED ISSUE #924: FeatureCache is over-checked and over-reset

benaadams:
Update to https://github.com/aspnet/HttpAbstractions/pull/925

Resolves #924 "FeatureCache is over-checked and over-reset"
Resolves #677 "FormFeature allocates when unused"
Resolves #880 "Don't allocate the FormFeature eagerly per request"

/cc @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Http/DefaultHttpContext.cs
+++ b/src/Microsoft.AspNetCore.Http/DefaultHttpContext.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 using System.Security.Claims;
 using System.Threading;
 using Microsoft.AspNetCore.Http.Authentication;
@@ -24,6 +25,7 @@ namespace Microsoft.AspNetCore.Http
         private readonly static Func<IFeatureCollection, ISessionFeature> _nullSessionFeature = f => null;
         private readonly static Func<IFeatureCollection, IHttpRequestIdentifierFeature> _newHttpRequestIdentifierFeature = f => new HttpRequestIdentifierFeature();
 
+        private FormOptions _formOptions;
         private FeatureReferences<FeatureInterfaces> _features;
 
         private HttpRequest _request;
@@ -44,15 +46,26 @@ namespace Microsoft.AspNetCore.Http
         }
 
         public DefaultHttpContext(IFeatureCollection features)
+            : this(features, FormOptions.Default)
         {
-            Initialize(features);
         }
 
-        public virtual void Initialize(IFeatureCollection features)
+        public DefaultHttpContext(IFeatureCollection features, FormOptions formOptions)
+        {
+            Initialize(features, formOptions);
+        }
+
+        public virtual void Initialize(IFeatureCollection features, FormOptions formOptions)
         {
             _features = new FeatureReferences<FeatureInterfaces>(features);
             _request = InitializeHttpRequest();
             _response = InitializeHttpResponse();
+            _formOptions = formOptions;
+        }
+
+        public virtual void Initialize(IFeatureCollection features)
+        {
+            Initialize(features, FormOptions.Default);
         }
 
         public virtual void Uninitialize()
@@ -85,6 +98,7 @@ namespace Microsoft.AspNetCore.Http
                 UninitializeWebSocketManager(_websockets);
                 _websockets = null;
             }
+            _formOptions = null;
         }
 
         private IItemsFeature ItemsFeature =>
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Http/DefaultHttpContext.cs
```diff
@@ -44,15 +46,26 @@ public DefaultHttpContext()
         }
 
         public DefaultHttpContext(IFeatureCollection features)
+            : this(new FeatureCollection(), FormOptions.Default)
```
davidfowl: Shouldn’t we be passing the feature collection and not a new one?

test/Microsoft.AspNetCore.Http.Tests/DefaultHttpContextTests.cs
```diff
@@ -158,7 +158,9 @@ public void UpdateFeatures_ClearsCachedFeatures()
             features.Set<IHttpWebSocketFeature>(new TestHttpWebSocketFeature());
 
             // featurecollection is set. all cached interfaces are null.
-            var context = new DefaultHttpContext(features);
+            var context = new DefaultHttpContext(features, FormOptions.Default);
```
davidfowl: Make sure we validate the default ctor as well

davidfowl (changes requested):

benaadams:
Need to unpick this a bit more; lot of calls going on 

(prior)
image

test/Microsoft.AspNetCore.Http.Tests/DefaultHttpContextTests.cs
benaadams: Done

src/Microsoft.AspNetCore.Http/DefaultHttpContext.cs
benaadams: Done

benaadams:
Before
image

After
image

Still a bucket load of calls; but less

src/Microsoft.AspNetCore.Http/Features/FormOptions.cs
```diff
@@ -8,6 +8,8 @@ namespace Microsoft.AspNetCore.Http.Features
 {
     public class FormOptions
     {
+        public static readonly FormOptions Default = new FormOptions();
```
Tratcher: We can't make a mutable instance a public static

benaadams: Good point

src/Microsoft.AspNetCore.Http/Internal/DefaultHttpRequest.cs
```diff
@@ -15,26 +15,39 @@ public class DefaultHttpRequest : HttpRequest
         // Lambdas hoisted to static readonly fields to improve inlining https://github.com/dotnet/roslyn/issues/13624
         private readonly static Func<IFeatureCollection, IHttpRequestFeature> _nullRequestFeature = f => null;
         private readonly static Func<IFeatureCollection, IQueryFeature> _newQueryFeature = f => new QueryFeature(f);
-        private readonly static Func<HttpRequest, IFormFeature> _newFormFeature = r => new FormFeature(r);
+        private readonly static Func<(HttpRequest request, FormOptions options), IFormFeature> _newFormFeature = (r) => new FormFeature(r.request, r.options ?? FormOptions.Default);
```
davidfowl: Tuples? 😄

benaadams: Custom struct or another Fetch overload?

muratg:
@davidfowl @Tratcher is this still relevant/useful in your opinion? If so, let's move it forward. If not, let's get it closed.

Ben, thanks for the PR!

davidfowl:
@benaadams I'm going to move this to 3.0

davidfowl (approved):

natemcmaster:
@davidfowl I'm about to merge this repo into aspnet/AspNetCore. If you'd like this PR merged, can you bring it in today? If not, I'm going to close this and you'll have to re-open the PR.

davidfowl:
@natemcmaster close it and I'll make @benaadams reopen it 😄

PR #1001, opened 2018-02-22
[PoC] HttpResponseStreamWriter on Stream on Pipe

benaadams:
No additional buffers or `ArrayPool` buffers when Stream implements `IDuplexPipe`

Also current `HttpResponseStreamWriter`  never ever called `Flush`; even if the user called it or when disposed O_o

Added

/cc @davidfowl @rynowak

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -19,6 +19,9 @@
     <MicrosoftNETTestSdkPackageVersion>15.3.0</MicrosoftNETTestSdkPackageVersion>
     <MoqPackageVersion>4.7.49</MoqPackageVersion>
     <SystemBuffersPackageVersion>4.5.0-preview2-26130-01</SystemBuffersPackageVersion>
+    <SystemMemoryPackageVersion>4.5.0-preview2-26130-01</SystemMemoryPackageVersion>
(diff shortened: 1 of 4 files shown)


benaadams:
Needs the `WriteAsync` in pipelines that takes a cancellation token.

This means *if* Kestrel implemented `IDuplexPipe` or better `IPipeWriter` on its `ResponseStream` the intermediate arrays could go away and encoding could be written directly to the `Pipe`; and MVC could take advantage of this with no changes.

davidfowl:
This is optimized for synchronous writing of utf8 text to an underlying `IBufferWriter<byte>`, https://github.com/aspnet/SignalR/blob/f7fc2647de0f91fbe6322aec95b334ad1437f427/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/Utf8BufferTextWriter.cs. We could try using it here as an alternative to what MVC uses (or try gutting this impl).

muratg:
@davidfowl @Tratcher is this still relevant/useful in your opinion? If so, let's move it forward. If not, let's get it closed.

Ben, thanks for the PR!

muratg:
Triage decision: We’re closing this PR because we don’t feel that this change is a good fit for the product at this time. 

PR #1002, opened 2018-02-22
Commas in doc comments

JunTaoLuo:
Noticed this while using the APIs. We separate values with `,` not with unconsciousness.


BrennanConroy:
:shipit:

Tratcher:
unconsciousness would be more effective.

PR #1003, opened 2018-02-27
Adds some security headers

joseftw:
Adds Referrer-Policy, X-Content-Type-Options, X-Frame-Options and X-XSS-Protection headers

### DIFF
--- a/src/Microsoft.Net.Http.Headers/HeaderNames.cs
+++ b/src/Microsoft.Net.Http.Headers/HeaderNames.cs
@@ -57,6 +57,7 @@ namespace Microsoft.Net.Http.Headers
         public const string ProxyAuthorization = "Proxy-Authorization";
         public const string Range = "Range";
         public const string Referer = "Referer";
+        public const string ReferrerPolicy = "Referrer-Policy";
         public const string RetryAfter = "Retry-After";
         public const string Scheme = ":scheme";
         public const string Server = "Server";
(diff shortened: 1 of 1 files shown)


Tratcher:
Spec references?

joseftw:
Referrer-Policy: https://www.w3.org/TR/referrer-policy/
X-Content-Type-Options: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
X-Frame-Options: https://www.w3.org/TR/UISecurity/
X-XSS-Protection: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

Sorry for some mozilla links but could not find any w3 links.

Tratcher:
@blowdart

Eilon:
@Tratcher / @blowdart - any further action planned on this PR?

blowdart:
It could be useful, I have no strong feelings.

Tratcher:
Yeah, it's not clear that adding only the header names is much value. If there were also APIs associated with these it would make more sense.

E.g. Referrer-Policy is not much use without it's list of possible values: "no-referrer" / "no-referrer-when-downgrade" / "strict-origin" / "strict-origin-when-cross-origin" / "same-origin" / "origin" / "origin-when-cross-origin" / "unsafe-url"
X-Content-Type-Options has only one defined value "nosniff".
X-Frame-Options and X-XSS-Protection both have some structure to them and you'd want some formatters.

muratg:

PR #1006, opened 2018-03-13
Fix trimming line if CR not directly before LF

richardhopton:
If the line contains a CR in any location than directly before the LF it would detect a valid line and then trim 2 characters from the end of the line, losing a "real" character.

e.g.
`This line\r would be treated as valid\n`
would be accepted and the resultant line output would be
`This line\r would be treated as vali`


Tratcher:
(Tests re-started)

PR #1008, opened 2018-03-20
Add AuthenticationProperties.Parameters property bag

poke:
As discussed over in aspnet/Security#1691, this is adding a `Parameters` property bag to the `AuthenticationProperties` object which can be used to pass values to the authentication handler.

I’ve also made the helper methods for the `Items` collection protected to allow them to be reused in subclasses.

I also considered adding tests for the helpers since there were none; just say if you’d like that.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
@@ -20,26 +20,41 @@ namespace Microsoft.AspNetCore.Authentication
         internal const string UtcDateTimeFormat = "r";
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="AuthenticationProperties"/> class
+        /// Initializes a new instance of the <see cref="AuthenticationProperties"/> class.
         /// </summary>
         public AuthenticationProperties()
-            : this(items: null)
-        {
-        }
+            : this(items: null, parameters: null)
+        { }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="AuthenticationProperties"/> class
+        /// Initializes a new instance of the <see cref="AuthenticationProperties"/> class.
         /// </summary>
-        /// <param name="items"></param>
+        /// <param name="items">State values dictionary to use.</param>
         public AuthenticationProperties(IDictionary<string, string> items)
+            : this(items, parameters: null)
+        { }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AuthenticationProperties"/> class.
+        /// </summary>
+        /// <param name="items">State values dictionary to use.</param>
+        /// <param name="parameters">Parameters dictionary to use.</param>
+        public AuthenticationProperties(IDictionary<string, string> items, IDictionary<string, object> parameters)
         {
             Items = items ?? new Dictionary<string, string>(StringComparer.Ordinal);
+            Parameters = parameters ?? new Dictionary<string, object>(StringComparer.Ordinal);
         }
 
         /// <summary>
         /// State values about the authentication session.
         /// </summary>
         public IDictionary<string, string> Items { get; }
+
+        /// <summary>
+        /// Collection of parameters that are passed to the authentication handler. These are not intended for
+        /// serialization or persistence, only for flowing data between call sites.
+        /// </summary>
+        public IDictionary<string, object> Parameters { get; }
 
         /// <summary>
         /// Gets or sets whether the authentication session is persisted across multiple requests.
@@ -86,12 +101,22 @@ namespace Microsoft.AspNetCore.Authentication
             set => SetBool(RefreshKey, value);
         }
 
-        private string GetString(string key)
+        /// <summary>
+        /// Get a string value from the <see cref="Items"/> collection.
+        /// </summary>
+        /// <param name="key">Property key.</param>
+        /// <returns>Retrieved value or <c>null</c> if the property is not set.</returns>
+        public string GetString(string key)
         {
             return Items.TryGetValue(key, out string value) ? value : null;
         }
 
-        private void SetString(string key, string value)
+        /// <summary>
+        /// Set a string value in the <see cref="Items"/> collection.
+        /// </summary>
+        /// <param name="key">Property key.</param>
+        /// <param name="value">Value to set or <c>null</c> to remove the property.</param>
+        public void SetString(string key, string value)
         {
             if (value != null)
             {
@@ -103,16 +128,44 @@ namespace Microsoft.AspNetCore.Authentication
             }
         }
 
-        private bool? GetBool(string key)
+        /// <summary>
+        /// Get a parameter from the <see cref="Parameters"/> collection.
+        /// </summary>
+        /// <typeparam name="T">Parameter type.</typeparam>
+        /// <param name="key">Parameter key.</param>
+        /// <returns>Retrieved value or the default value if the property is not set.</returns>
+        public T GetParameter<T>(string key)
+            => Parameters.TryGetValue(key, out var obj) && obj is T value ? value : default;
+
+        /// <summary>
+        /// Set a parameter value in the <see cref="Parameters"/> collection.
+        /// </summary>
+        /// <typeparam name="T">Parameter type.</typeparam>
+        /// <param name="key">Parameter key.</param>
+        /// <param name="value">Value to set.</param>
+        public void SetParameter<T>(string key, T value)
+            => Parameters[key] = value;
+
+        /// <summary>
+        /// Get a bool value from the <see cref="Items"/> collection.
+        /// </summary>
+        /// <param name="key">Property key.</param>
+        /// <returns>Retrieved value or <c>null</c> if the property is not set.</returns>
+        protected bool? GetBool(string key)
         {
-            if (Items.TryGetValue(key, out string value) && bool.TryParse(value, out bool refresh))
+            if (Items.TryGetValue(key, out string value) && bool.TryParse(value, out bool boolValue))
             {
-                return refresh;
+                return boolValue;
             }
             return null;
         }
 
-        private void SetBool(string key, bool? value)
+        /// <summary>
+        /// Set a bool value in the <see cref="Items"/> collection.
+        /// </summary>
+        /// <param name="key">Property key.</param>
+        /// <param name="value">Value to set or <c>null</c> to remove the property.</param>
+        protected void SetBool(string key, bool? value)
         {
             if (value.HasValue)
             {
@@ -124,7 +177,12 @@ namespace Microsoft.AspNetCore.Authentication
             }
         }
 
-        private DateTimeOffset? GetDateTimeOffset(string key)
+        /// <summary>
+        /// Get a DateTimeOffset value from the <see cref="Items"/> collection.
+        /// </summary>
+        /// <param name="key">Property key.</param>
+        /// <returns>Retrieved value or <c>null</c> if the property is not set.</returns>
+        protected DateTimeOffset? GetDateTimeOffset(string key)
         {
             if (Items.TryGetValue(key, out string value)
                 && DateTimeOffset.TryParseExact(value, UtcDateTimeFormat, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind, out DateTimeOffset dateTimeOffset))
@@ -134,7 +192,12 @@ namespace Microsoft.AspNetCore.Authentication
             return null;
         }
 
-        private void SetDateTimeOffset(string key, DateTimeOffset? value)
+        /// <summary>
+        /// Set a DateTimeOffset value in the <see cref="Items"/> collection.
+        /// </summary>
+        /// <param name="key">Property key.</param>
+        /// <param name="value">Value to set or <c>null</c> to remove the property.</param>
+        protected void SetDateTimeOffset(string key, DateTimeOffset? value)
         {
             if (value.HasValue)
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
```diff
@@ -20,27 +20,33 @@ public class AuthenticationProperties
...
             : this(items: null)
         {
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="AuthenticationProperties"/> class
+        /// Initializes a new instance of the <see cref="AuthenticationProperties"/> class.
         /// </summary>
         /// <param name="items"></param>
         public AuthenticationProperties(IDictionary<string, string> items)
```
Tratcher: We probably need an additional constructor that takes both.

HaoK:
yes please to tests :)

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
HaoK: Do we want a copy constructor/overload that lets you easily copy both Items and Parameters into a new AuthenticationProperties?

poke: Hmm, I didn’t want to add one for the `Properties` dictionary as that would likely be confusing but a copy constructor could be useful.

HaoK: Yeah either a copy constructor that takes AuthenticationProperties or one that takes items + properties, should have at least one of those options.

Tratcher: Make it one that takes both collections, not a copy constructor. We may need this if we have to wrap Identitie's result.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
```diff
@@ -86,12 +92,12 @@ public string RedirectUri
             set => SetBool(RefreshKey, value);
         }
 
-        private string GetString(string key)
```
Tratcher: Make them public? They're operating on public state.

poke: All of them? Don’t want it to be too noisy and distract from the properties.

Tratcher: Hmm, ok, leave it protected, but it still needs a basic doc comment.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
```diff
@@ -86,12 +92,12 @@ public string RedirectUri
...
+        protected string GetString(string key)
```
Tratcher: Two more methods to add: `public/protected object GetParameter(string name)` that does the same TryGetValue to avoid dictionary throwing. And add SetParemeter for symmetry.

poke: Generic overload that attempts a type cast yes/no?

HaoK: I think we should see what the actual derived ChallengeProperties look like before adding the sugar/helpers that we need for them...  But I don't disagree likely we'll need more helpers for the parameters here...

I just think we should start with none, and add them once we need them

HaoK: Basically I don't think we'll be able to properly review/design the base class helpers without seeing what the ChallengeProperties actually look like for real

Tratcher: We don't want to do too many trips back and forth, x-repo development is even more of a pain for external users. A single GetParameter method is better than nothing. Don't worry about the generic.

HaoK: I'm not sure its any less painful for us :)  But I'd actually recommend doing both PRs in parallel instead of back and forth per say...

Start with the initial one to get the type to show up in the official bits (if needed), and then have a parallel one with additional helpers added as you go that gets checked in at the same time as security.

HaoK: But given how this is purely additive, he can probably just create both PRs at the same time and we can merge them together at once they are fully complete

poke: We have at least an `ICollection<string>` for the scopes, a `TimeSpan?` for the OIDC’s `MaxAge`, and a `bool` for Google’s `IncludeGrantedScopesParameterKey`.

Having a generic `T GetParameter<T>(string key)` and a `void SetParameter<T>(string key, T value)` would easily allow us to cover all of this, so we wouldn’t need more helpers than those for the properties.

HaoK: Yeah that's how I'd have done it, sounds good to me, do what feels best/cleanest :)

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
poke: The current constructor will not copy the passed `Items` though but will use the passed collection as is. Would you want the same behavior for the `Parameters` dictionary then? This would be kind of the opposite of a copy constructor then.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
Tratcher: The generic only hides a cast. It's not giving you serialization like the other helpers here.

`GetParameter<ICollection<string>>` vs `(ICollection<string>)GetParameter` is pretty much the same.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
Tratcher: Yes, take the existing collection. It's a very shallow copy 😁

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
HaoK: Going back to my point...just implement the actual properties classes...before trying to design the helpers :)

poke: I was actually doing the following in my implementation:

```
Parameters.TryGetValue(key, out var obj) && obj is T value ? value : default
```

I would personally rather ignore parameters that have the wrong type than have them raise an exception for them (or to add error handling into the handlers for everything).

HaoK: My personal guess is that we would have a common set of helpers, that takes the Dictionary as an argument.

`GetString(Items | Parameters, key),`

Tratcher: The dictionaries are different types...

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
```diff
@@ -109,6 +127,40 @@ protected void SetString(string key, string value)
...
+        /// <typeparam name="T">Parameter type,</typeparam>
+        /// <param name="key">Parameter key.</param>
+        /// <param name="value">Value to set or <c>null</c> to remove the property.</param>
+        public void SetParameter<T>(string key, T value)
+        {
+            if (value != null)
+            {
+                Parameters[key] = value;
+            }
+            else if (Parameters.ContainsKey(key))
```
HaoK: Do we really need this null clearing logic?  Can't they just do Parameters.Remove(key) directly instead of set with null?

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
```diff
@@ -109,6 +127,40 @@ protected void SetString(string key, string value)
...
 
+        /// <summary>
+        /// Get a parameter from the <see cref="Parameters"/> collection.
+        /// </summary>
+        /// <typeparam name="T">Parameter type.</typeparam>
+        /// <param name="key">Parameter key.</param>
+        /// <returns>Retrieved value or the default value if the property is not set.</returns>
+        public T GetParameter<T>(string key)
+        {
+            return Parameters.TryGetValue(key, out var obj) && obj is T value ? value : default;
```
HaoK: nit if you are going to right a fancy complex one liner, you might as well use => and drop the return and get rid of the extra { }'s :)

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
poke: That was mostly to get a symmetric behavior to the Set~ methods on the `Items` dictionary.

HaoK: I don't think that's an important concern as the Items are used for serialization where this behavior is much more important

poke: That’s a good point! I’ll remove that part then.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
```diff
@@ -20,27 +20,41 @@ public class AuthenticationProperties
...
+            Parameters = parameters ?? new Dictionary<string, object>(StringComparer.Ordinal);
         }
 
         /// <summary>
         /// State values about the authentication session.
         /// </summary>
         public IDictionary<string, string> Items { get; }
 
+        /// <summary>
+        /// Collection of parameters that are passed to the authentication handler.
```
Tratcher: "These are not intended for serialization or persistence, only for flowing data between call sites."

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
davidfowl: Why are we exposing these things?

poke:
Alright, squashed and cleaned up! I guess this is ready if @HaoK has no objections.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
poke: (GitHub’s UI sometimes makes it really easy to miss comments…)

@davidfowl I made the string-based helpers public so that they can serve as a more friendly way to interact with the `Items` dictionary. And especially so that handlers have a simple API to retrieve values (for when they actually want to persist data inside `Items`).

HaoK: I thought we were going to keep these as protected?

HaoK: We don't actually need Get/SetParameter public either right?  We are going to have properties that expose the parameters right?

Tratcher:
I tried moving your PR to the release/2.1 branch but it needs a clean rebase. We'll be working in release/2.1 for the remainder of preview2.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
davidfowl: We don't do this anywhere else in the stack.

Tratcher: We do, but they're usually extension methods.

davidfowl: Where?

poke: I don’t mind either way but as soon as a handler needs a value from the items collection, they are effectively going to reimplement the thing again. And since the `GetParameter` method is public, I don’t think a set of public methods for the `Items` collection hurts.

poke:
Rebased onto release/2.1.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
Tratcher: https://github.com/aspnet/Caching/blob/dev/src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs

Tratcher:
@poke can you send me your e-mail address to the address in my profile? I have something to send you.

poke:
@Tratcher Btw. did you receive my email? I never got anything back (I’m actually losing trust in gmail at the moment >_<)

Tratcher:
Ack, it's taking longer to put together than I expected.

poke:
Alright, just wanted to make sure that it wasn’t my email not working properly :)

PR #1009, opened 2018-03-21
Do not rely on the implicit StringValues to array converter.

Tratcher:
Fixes http://aspnetci/viewLog.html?buildId=425449&tab=buildResultsDiv&buildTypeId=Lite_UniverseTest

https://github.com/aspnet/Common/pull/323 made a subtle change that causes `string[] myArray = new StringValues(new string[0])` to implicitly become null. This broke many components up stack that were expecting GetCommaSeparatedValues to return a non-null value even for an empty or missing header resulting in and caused 76 tests failures.

This should fix everything except the MVC test ValueProviderResultTest.Construct_With_None.

@benaadams @ryanbrandenburg

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -4,22 +4,22 @@
   </PropertyGroup>
   <PropertyGroup Label="Package Versions">
     <InternalAspNetCoreSdkPackageVersion>2.1.0-preview2-15728</InternalAspNetCoreSdkPackageVersion>
-    <MicrosoftAspNetCoreTestingPackageVersion>2.1.0-preview2-30272</MicrosoftAspNetCoreTestingPackageVersion>
-    <MicrosoftExtensionsActivatorUtilitiesSourcesPackageVersion>2.1.0-preview2-30272</MicrosoftExtensionsActivatorUtilitiesSourcesPackageVersion>
-    <MicrosoftExtensionsCopyOnWriteDictionarySourcesPackageVersion>2.1.0-preview2-30272</MicrosoftExtensionsCopyOnWriteDictionarySourcesPackageVersion>
-    <MicrosoftExtensionsDependencyInjectionPackageVersion>2.1.0-preview2-30272</MicrosoftExtensionsDependencyInjectionPackageVersion>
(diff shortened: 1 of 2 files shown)


benaadams:
Should `StringValues` change
```csharp
public static implicit operator string[] (StringValues value)
{
    return value.GetArrayValue();
}
```
to
```csharp
public static implicit operator string[] (StringValues value)
{
    return value.ToArray();
}
```

`string[] myArray = new StringValues()` or `string[] myArray = default(StringValues);` would have same issue?

(Though is fixed by not relying on it being non-null)

Tratcher:
The implicit converters should round trip, calling ToArray just inverts the problem.

halter73:
@Tratcher Is it critical for converters to round trip if they stabilize to a conical form?

I'm thinking that @benaadams might be right. Converting empty StringValues to an empty array instead of a null array seems safer. If we did this, would this get Universe passing?

benaadams:
Prior to the change `StringValues` was ambiguous and had multiple ways to represent the low counts:

Count == 0
* `null`
* `string[0]`

Count == 1
* `string`
* `string[1]`

Count == 2
* `string[2]`

This means the check `.Count == 0` becomes
```csharp
if ((_value != null ? 1 : (_values?.Length ?? 0)) == 0)
```
So the cast of a `.Count == 0` to an `string[]` could return become both `string[0]` and `null` anyway.

After https://github.com/aspnet/Common/pull/323 there is a single representation for each count

Count == 0
* `null`

Count == 1
* `string`

Count == 2
* `string[2]`

There is a fast path for `.Count == 0` of `.IsNull` becoming
```csharp
if (_value == null && _values == null)
```
Obviously the inverse for `.Count > 0` being `!value.IsNull`

And the cast of a `.Count == 0` to an `string[]` can only return `null`

Though, its easy enough to have it return `Array.Empty<string>()` instead

benaadams:
I think there is a different change needed...

benaadams:
Is this the change that's needed to fix Universe? https://github.com/aspnet/HttpAbstractions/pull/1010

benaadams:
With https://github.com/aspnet/Common/pull/330

PR #1010, opened 2018-03-21
[NoMerge] React to StringValues

benaadams:
Isn't it only `HeaderDictionary` that needs the change to not add `null`s to the dictionary in the `IDictionary` path as it does in the `IHeaderDictionary` path?

Also needs https://github.com/aspnet/Common/pull/330

/cc @Tratcher  @ryanbrandenburg @dougbu @rynowak @halter73


benaadams:
e.g. currently its saying there is a header; then returning `null`; whereas it should say there isn't a header

ryanbrandenburg:
Looks like the tests fail with this change.

benaadams:
> Looks like the tests fail with this change.

Also needs aspnet/Common#330

benaadams:
I can put a fix in (e.g. put `.ToArray();` back), then revert after?

benaadams:
Added fix back

Tratcher:
Heads up, the original change was reverted. We'll reintroduce it more cautiously by doing full universe builds before merging.

muratg:
We’re closing this PR because we don’t feel that this change is a good fit for the product at this time. 

PR #1011, opened 2018-03-26
Make stream public and add doc

Tornhoof:
Changes:
- Made the stream public
- Added comments to all public members, the interesting read line comments were taken and modified from https://github.com/dotnet/coreclr/blob/b5f4d2df2e087401f2c3aab2c37021e326707915/src/mscorlib/shared/System/IO/StreamReader.cs#L795-L800. Most other comments were either <inheritdoc>


src/Microsoft.AspNetCore.WebUtilities/BufferedReadStream.cs
```diff
@@ -10,7 +10,11 @@
 
 namespace Microsoft.AspNetCore.WebUtilities
 {
-    internal class BufferedReadStream : Stream
+    /// <summary>
+    /// A Stream that wraps another stream and allows reading lines
```
Eilon: End the sentence with a period. Doc comments are kind of like HTML, where newlines are just whitespace. For that reason, every sentence needs to be a complete sentence because it will all show up as one paragraph in the final docs.

src/Microsoft.AspNetCore.WebUtilities/BufferedReadStream.cs
```diff
@@ -22,11 +26,22 @@ internal class BufferedReadStream : Stream
         private int _bufferCount = 0;
         private bool _disposed;
 
+        /// <summary>
+        /// Creates a new stream
+        /// </summary>
+        /// <param name="inner">Wrapped stream</param>
```
Eilon: Would this be `The stream to wrap.`?

src/Microsoft.AspNetCore.WebUtilities/BufferedReadStream.cs
```diff
@@ -39,36 +54,45 @@ public BufferedReadStream(Stream inner, int bufferSize, ArrayPool<byte> bytePool
             _buffer = bytePool.Rent(bufferSize);
         }
 
+        /// <summary>
+        /// The currently buffered data
```
Eilon: See earlier comment about ending sentences with periods. Please apply to all new doc comments.

src/Microsoft.AspNetCore.WebUtilities/BufferedReadStream.cs
```diff
@@ -202,6 +235,10 @@ public override int Read(byte[] buffer, int offset, int count)
             return await _inner.ReadAsync(buffer, offset, count, cancellationToken);
         }
 
+        /// <summary>
+        /// Ensures that the buffer is not empty
+        /// </summary>
+        /// <returns>buffer has data</returns>
```
Eilon: For these types of docs, you can do something similar to this: https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/src/Microsoft.AspNetCore.Mvc.Abstractions/Formatters/IOutputFormatter.cs#L18

```
        /// <returns>Returns <c>true</c> if the formatter can write the response; <c>false</c> otherwise.</returns>
```

src/Microsoft.AspNetCore.WebUtilities/BufferedReadStream.cs
```diff
@@ -226,6 +268,11 @@ public bool EnsureBuffered()
             return _bufferCount > 0;
         }
 
+        /// <summary>
+        /// Ensures that a minimum amount of buffered data is available
+        /// </summary>
+        /// <param name="minCount">Minimum amount of buffered data</param>
```
Eilon: `The minimum...`

src/Microsoft.AspNetCore.WebUtilities/BufferedReadStream.cs
```diff
@@ -226,6 +268,11 @@ public bool EnsureBuffered()
...
+        /// <returns>Minimum amount of buffered data available</returns>
```
Eilon: `The minimum...`

PR #1019, opened 2018-06-13
Add ITlsHandshakeFeature

Tratcher:
Part of https://github.com/aspnet/KestrelHttpServer/issues/2661


davidfowl:
I think this should be a connection feature instead of an http feature

Tratcher:
Where do those live? Kestrel.Core/Features?  How does HttpSys implement it?

Tratcher:
Ah, Connections.Abstractions/Features?

PR #1030, opened 2018-08-09
Copy core endpoint routing types to HttpAbstractions

JamesNK:
Note that this is in master for 3.0
Will type forward from routing abstractions.

https://github.com/aspnet/Routing/issues/607

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Microsoft.AspNetCore.Http.Abstractions.csproj
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Microsoft.AspNetCore.Http.Abstractions.csproj
@@ -20,6 +20,7 @@ Microsoft.AspNetCore.Http.HttpResponse</Description>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.ActivatorUtilities.Sources" PrivateAssets="All" Version="$(MicrosoftExtensionsActivatorUtilitiesSourcesPackageVersion)" />
+    <PackageReference Include="Microsoft.Extensions.PropertyHelper.Sources" Version="$(MicrosoftExtensionsPropertyHelperSourcesPackageVersion)" PrivateAssets="All" />
     <PackageReference Include="System.Text.Encodings.Web" Version="$(SystemTextEncodingsWebPackageVersion)" />
   </ItemGroup>
 
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Routing/Endpoint.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Routing/Endpoint.cs
@@ -1,0 +1,49 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Http
+{
+    /// <summary>
+    /// Respresents a logical endpoint in an application.
+    /// </summary>
+    public class Endpoint
+    {
+        /// <summary>
+        /// Creates a new instance of <see cref="Endpoint"/>.
+        /// </summary>
+        /// <param name="requestDelegate">The delegate used to process requests for the endpoint.</param>
+        /// <param name="metadata">
+        /// The endpoint <see cref="EndpointMetadataCollection"/>. May be null.
+        /// </param>
+        /// <param name="displayName">
+        /// The informational display name of the endpoint. May be null.
+        /// </param>
+        public Endpoint(
+            RequestDelegate requestDelegate,
+            EndpointMetadataCollection metadata,
+            string displayName)
+        {
+            // All are allowed to be null
+            RequestDelegate = requestDelegate;
+            Metadata = metadata ?? EndpointMetadataCollection.Empty;
+            DisplayName = displayName;
+        }
+
+        /// <summary>
+        /// Gets the informational display name of this endpoint.
+        /// </summary>
+        public string DisplayName { get; }
+
+        /// <summary>
+        /// Gets the collection of metadata associated with this endpoint.
+        /// </summary>
+        public EndpointMetadataCollection Metadata { get; }
+
+        /// <summary>
+        /// Gets the delegate used to process requests for the endpoint.
+        /// </summary>
+        public RequestDelegate RequestDelegate { get; }
+
+        public override string ToString() => DisplayName ?? base.ToString();
+    }
+}
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Routing/EndpointMetadataCollection.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Routing/EndpointMetadataCollection.cs
@@ -1,0 +1,201 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.AspNetCore.Http
+{
+    /// <summary>
+    /// A collection of arbitrary metadata associated with an endpoint.
+    /// </summary>
+    /// <remarks>
+    /// <see cref="EndpointMetadataCollection"/> instances contain a list of metadata items
+    /// of arbitrary types. The metadata items are stored as an ordered collection with
+    /// items arranged in ascending order of precedence.
+    /// </remarks>
+    public sealed class EndpointMetadataCollection : IReadOnlyList<object>
+    {
+        /// <summary>
+        /// An empty <see cref="EndpointMetadataCollection"/>.
+        /// </summary>
+        public static readonly EndpointMetadataCollection Empty = new EndpointMetadataCollection(Array.Empty<object>());
+
+        private readonly object[] _items;
+        private readonly ConcurrentDictionary<Type, object[]> _cache;
+
+        /// <summary>
+        /// Creates a new <see cref="EndpointMetadataCollection"/>.
+        /// </summary>
+        /// <param name="items">The metadata items.</param>
+        public EndpointMetadataCollection(IEnumerable<object> items)
+        {
+            if (items == null)
+            {
+                throw new ArgumentNullException(nameof(items));
+            }
+
+            _items = items.ToArray();
+            _cache = new ConcurrentDictionary<Type, object[]>();
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="EndpointMetadataCollection"/>.
+        /// </summary>
+        /// <param name="items">The metadata items.</param>
+        public EndpointMetadataCollection(params object[] items)
+            : this((IEnumerable<object>)items)
+        {
+        }
+
+        /// <summary>
+        /// Gets the item at <paramref name="index"/>.
+        /// </summary>
+        /// <param name="index">The index of the item to retrieve.</param>
+        /// <returns>The item at <paramref name="index"/>.</returns>
+        public object this[int index] => _items[index];
+
+        /// <summary>
+        /// Gets the count of metadata items.
+        /// </summary>
+        public int Count => _items.Length;
+
+        /// <summary>
+        /// Gets the most significant metadata item of type <typeparamref name="T"/>.
+        /// </summary>
+        /// <typeparam name="T">The type of metadata to retrieve.</typeparam>
+        /// <returns>
+        /// The most significant metadata of type <typeparamref name="T"/> or <c>null</c>.
+        /// </returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public T GetMetadata<T>() where T : class
+        {
+            if (_cache.TryGetValue(typeof(T), out var result))
+            {
+                var length = result.Length;
+                return length > 0 ? (T)result[length - 1] : default;
+            }
+
+            return GetMetadataSlow<T>();
+        }
+
+        private T GetMetadataSlow<T>() where T : class
+        {
+            var array = GetOrderedMetadataSlow<T>();
+            var length = array.Length;
+            return length > 0 ? array[length - 1] : default;
+        }
+
+        /// <summary>
+        /// Gets the metadata items of type <typeparamref name="T"/> in ascending
+        /// order of precedence.
+        /// </summary>
+        /// <typeparam name="T">The type of metadata.</typeparam>
+        /// <returns>A sequence of metadata items of <typeparamref name="T"/>.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public IEnumerable<T> GetOrderedMetadata<T>() where T : class
+        {
+            if (_cache.TryGetValue(typeof(T), out var result))
+            {
+                return (T[])result;
+            }
+
+            return GetOrderedMetadataSlow<T>();
+        }
+
+        private T[] GetOrderedMetadataSlow<T>() where T : class
+        {
+            var items = new List<T>();
+            for (var i = 0; i < _items.Length; i++)
+            {
+                if (_items[i] is T item)
+                {
+                    items.Add(item);
+                }
+            }
+
+            var array = items.ToArray();
+            _cache.TryAdd(typeof(T), array);
+            return array;
+        }
+
+        /// <summary>
+        /// Gets an <see cref="IEnumerator"/> of all metadata items.
+        /// </summary>
+        /// <returns>An <see cref="IEnumerator"/> of all metadata items.</returns>
+        public Enumerator GetEnumerator() => new Enumerator(this);
+
+        /// <summary>
+        /// Gets an <see cref="IEnumerator{Object}"/> of all metadata items.
+        /// </summary>
+        /// <returns>An <see cref="IEnumerator{Object}"/> of all metadata items.</returns>
+        IEnumerator<object> IEnumerable<object>.GetEnumerator() => GetEnumerator();
+
+        /// <summary>
+        /// Gets an <see cref="IEnumerator"/> of all metadata items.
+        /// </summary>
+        /// <returns>An <see cref="IEnumerator"/> of all metadata items.</returns>
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        /// <summary>
+        /// Enumerates the elements of an <see cref="EndpointMetadataCollection"/>.
+        /// </summary>
+        public struct Enumerator : IEnumerator<object>
+        {
+            // Intentionally not readonly to prevent defensive struct copies
+            private object[] _items;
+            private int _index;
+
+            internal Enumerator(EndpointMetadataCollection collection)
+            {
+                _items = collection._items;
+                _index = 0;
+                Current = null;
+            }
+
+            /// <summary>
+            /// Gets the element at the current position of the enumerator
+            /// </summary>
+            public object Current { get; private set; }
+
+            /// <summary>
+            /// Releases all resources used by the <see cref="Enumerator"/>.
+            /// </summary>
+            public void Dispose()
+            {
+            }
+
+            /// <summary>
+            /// Advances the enumerator to the next element of the <see cref="Enumerator"/>.
+            /// </summary>
+            /// <returns>
+            /// <c>true</c> if the enumerator was successfully advanced to the next element;
+            /// <c>false</c> if the enumerator has passed the end of the collection.
+            /// </returns>
+            public bool MoveNext()
+            {
+                if (_index < _items.Length)
+                {
+                    Current = _items[_index++];
+                    return true;
+                }
+
+                Current = null;
+                return false;
+            }
+
+            /// <summary>
+            /// Sets the enumerator to its initial position, which is before the first element in the collection.
+            /// </summary>
+            public void Reset()
+            {
+                _index = 0;
+                Current = null;
+            }
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Routing/IEndpointFeature.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Routing/IEndpointFeature.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Http.Features
+{
+    /// <summary>
+    /// A feature interface for endpoint routing. Use <see cref="HttpContext.Features"/>
+    /// to access an instance associated with the current request.
+    /// </summary>
+    public interface IEndpointFeature
+    {
+        /// <summary>
+        /// Gets or sets the selected <see cref="Http.Endpoint"/> for the current
+        /// request.
+        /// </summary>
+        Endpoint Endpoint { get; set; }
+    }
+}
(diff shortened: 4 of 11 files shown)


src/Microsoft.AspNetCore.Http.Abstractions/Routing/IEndpointFeature.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Features
+{
+    /// <summary>
+    /// A feature interface for endpoint routing. Use <see cref="HttpContext.Features"/>
+    /// to access an instance associated with the current request.
+    /// </summary>
+    public interface IEndpointFeature
```
davidfowl: This should be in the http.features assembly

src/Microsoft.AspNetCore.Http.Abstractions/Microsoft.AspNetCore.Http.Abstractions.csproj
```diff
@@ -20,6 +20,7 @@ Microsoft.AspNetCore.Http.HttpResponse</Description>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.ActivatorUtilities.Sources" PrivateAssets="All" Version="$(MicrosoftExtensionsActivatorUtilitiesSourcesPackageVersion)" />
+    <PackageReference Include="Microsoft.Extensions.PropertyHelper.Sources" Version="$(MicrosoftExtensionsPropertyHelperSourcesPackageVersion)" PrivateAssets="All" />
```
davidfowl: Why?

davidfowl: Ah found it.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/IEndpointFeature.cs
JamesNK: It depends on all the other types moved in this PR. If IEndpointFeature is in http.features then everything else is as well.

Yay/nay?

src/Microsoft.AspNetCore.Http.Abstractions/Microsoft.AspNetCore.Http.Abstractions.csproj
JamesNK: RouteValueDictionary does some reflection when taking an `object`.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/Endpoint.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing
```
davidfowl: We should reconsider this namespace.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/EndpointMetadataCollection.cs
```diff
@@ -0,0 +1,178 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Routing
```
davidfowl: We should reconsider this namespace.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/IEndpointFeature.cs
davidfowl: Maybe, this is in 3.0 so I'm ok doing that and iterating. Session is similar.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/Endpoint.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    /// <summary>
+    /// Respresents a logical endpoint in an application.
+    /// </summary>
+    public abstract class Endpoint
```
davidfowl: What about:

```
public interface IEndpoint 
{ 
    EndpointMetadataCollection Metadata { get; } 
}
```

cc @rynowak

rynowak: This spawned an IM discussion. I've tried a few different approaches and the way things are is the best compromise I've found.


Goals:

I don't want Endpoint/IEndpoint to become something that application/framework authors subclass. I want it to be **data** not a service. The world breaks when these becomes services. I also don't want it to be polymorphic because we are creating a builder API to create these. Builder APIs get wierd when when the thing you're buildering is polymorphic. (I can give examples if necessary).

I don't want to leak all of https://github.com/aspnet/Routing/tree/release/2.2/src/Microsoft.AspNetCore.Routing/Patterns into this repo. I think it's important that the primary type (`MatcherEndpoint`) exposes details like the route pattern as a first-class property, accepts it in the constructor, etc.

The separation largely enforces the layering we already know we want. If you want to see and understand what routing did in a middleware, you just need `Endpoint`. If you want to register things in routing, you need to reference routing to use `MatcherEndpoint`. It doesn't make much sense to move `MatcherEndpoint` down to this level, because you wouldn't be able to use them for anything without routing.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/IEndpointFeature.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// Gets or sets the selected <see cref="Routing.Endpoint"/> for the current
+        /// request.
+        /// </summary>
+        Endpoint Endpoint { get; set; }
+
+        /// <summary>
+        /// Gets or sets a delegate that can be used to invoke the current
+        /// <see cref="Routing.Endpoint"/>.
+        /// </summary>
+        Func<RequestDelegate, RequestDelegate> Invoker { get; set; }
```
rynowak: I think we should try to decide on this before merging to the best of our knowledge. I'm not sure that we understand any reason why this takes 'next' and we're paying a perf and complexity cost for it. IIRC we did what it be do largely as a hedge.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/RouteValueDictionary.cs
```diff
@@ -0,0 +1,614 @@
...
+                }
+                else
+                {
+                    // Both null, we need to hold on 'start' since we
+                    // still need to fill it with something.
+                    end--;
+                }
+            }
+
+            return new RouteValueDictionary()
```
rynowak: #PeakRyan

src/Microsoft.AspNetCore.Http.Abstractions/Routing/IEndpointFeature.cs
JamesNK: If Endpoint is in http.abstractions and not http.features then every future feature that uses it will need to be http.abstractions or higher (unless we type forward it again)

src/Microsoft.AspNetCore.Http.Abstractions/Routing/Endpoint.cs
JamesNK: +1 that it is data and not a behavior to bolt onto other classes.

JamesNK:
Why is there no package for the feature branch? https://dotnet.myget.org/feed/aspnetcore-dev/package/nuget/Microsoft.AspNetCore.Http.Abstractions

src/Microsoft.AspNetCore.Http.Abstractions/Routing/IEndpointFeature.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Features
```
rynowak: Do we need this new namespace? This is the only thing in it.

Tratcher:
So why is all of this going into the Http.Abstractions package? I only see one type reference (RequestDelegate) that even relates to Http. I Also see a lot more implementation than I'd expect in the abstractions package. This seems like it would become its own thing like Microsoft.AspNetCore.Routing.Something.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/EndpointMetadataCollection.cs
```diff
@@ -0,0 +1,178 @@
...
+        /// order of precedence.
+        /// </summary>
+        /// <typeparam name="T">The type of metadata.</typeparam>
+        /// <returns>A sequence of metadata items of <typeparamref name="T"/>.</returns>
+        public IEnumerable<T> GetOrderedMetadata<T>() where T : class
+        {
+            for (var i = 0; i < _items.Length; i++)
+            {
+                var item = _items[i] as T;
+                if (item != null)
```
pranavkm: `if (_items[i] is T item)`

src/Microsoft.AspNetCore.Http.Abstractions/Routing/EndpointMetadataCollection.cs
```diff
@@ -0,0 +1,178 @@
...
+
+        /// <summary>
+        /// Enumerates the elements of an <see cref="EndpointMetadataCollection"/>.
+        /// </summary>
+        public struct Enumerator : IEnumerator<object>
+        {
+            // Intentionally not readonly to prevent defensive struct copies
+            private object[] _items;
+            private int _index;
+            private object _current;
```
pranavkm: You could get away with not having this field (just index in to the array as part of `Current`)

src/Microsoft.AspNetCore.Http.Abstractions/Routing/RouteValueDictionary.cs
```diff
@@ -0,0 +1,614 @@
...
+                            nameof(RouteValueDictionary));
+                        throw new InvalidOperationException(message);
+                    }
+
+                    names.Add(property.Name, property);
+                }
+            }
+        }
+
+        private class PropertyCache : ConcurrentDictionary<Type, PropertyHelper[]>
```
pranavkm: This is weird

src/Microsoft.AspNetCore.Http.Abstractions/Routing/RouteValueDictionary.cs
```diff
@@ -0,0 +1,614 @@
...
+                for (var i = 0; i < keys.Length; i++)
+                {
+                    keys[i] = array[i].Key;
+                }
+
+                return keys;
+            }
+        }
+
+        IEnumerable<string> IReadOnlyDictionary<string, object>.Keys
```
pranavkm: `IEnumerable<string> IReadOnlyDictionary<string, object>.Keys => Keys` for 💯 points

kichalla:
> Why is there no package for the feature branch? https://dotnet.myget.org/feed/aspnetcore-dev/package/nuget/Microsoft.AspNetCore.Http.Abstractions

@JamesNK We would ultimately want to move to VSTS and it currently does not have support for feature branches, so @natemcmaster enabled feature branch support only for Routing and MVC last week on TeamCity as we were making cross-repo changes.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/IEndpointFeature.cs
JamesNK: Other features are in a `.Features` namespace. On the other hand the current `IRoutingFeature` in `Microsoft.AspNetCore.Routing` isn't.

I don't have a strong opinion either way.

JamesNK:
@natemcmaster Is it difficult to add feature branch builds to HttpAbstractions?

natemcmaster:
Done. http://aspnetci/viewType.html?buildTypeId=Lite_HttpAbstractions&branch_Lite=endpoint-routing-types&tab=buildTypeStatusDiv

src/Microsoft.AspNetCore.Http.Abstractions/Routing/IEndpointFeature.cs
JamesNK: @davidfowl Do you have a opinion here?

src/Microsoft.AspNetCore.Http.Abstractions/Routing/IEndpointFeature.cs
JamesNK: https://github.com/aspnet/Routing/pull/707

src/Microsoft.AspNetCore.Http.Abstractions/Routing/IEndpointFeature.cs
davidfowl: I think it should go into the top level features namespace

src/Microsoft.AspNetCore.Http.Abstractions/Routing/Endpoint.cs
davidfowl: I had a long discussion with @rynowak and I understand how we landed here but there are things that give me pause. This doesn't feel like an Endpoint, it just feels like something that gives you metadata. The Endpoint is really the `MatcherEndpoint`, that's the thing everyone will create. This type is going to be used by middleware and will hang off the IEndpointFeature. 

It feels more like a container for metadata than anything else.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/EndpointMetadataCollection.cs
rynowak: Yeah, please don't change this, it was done this way on purpose

src/Microsoft.AspNetCore.Http.Abstractions/Routing/Endpoint.cs
rynowak: > It feels more like a container for metadata than anything else.

If we moved the request delegate down from `MatcherEndpoint` to here, would that address your concerns?

src/Microsoft.AspNetCore.Http.Abstractions/Routing/EndpointMetadataCollection.cs
JamesNK: https://github.com/aspnet/Routing/pull/708

src/Microsoft.AspNetCore.Http.Abstractions/Routing/RouteValueDictionary.cs
JamesNK: https://github.com/aspnet/Routing/pull/708

src/Microsoft.AspNetCore.Http.Abstractions/Routing/RouteValueDictionary.cs
JamesNK: https://github.com/aspnet/Routing/pull/708

JamesNK:
I have moved Invoker from `IEndpointFeature` to `Endpoint` from @davidfowl's offline feedback.

Outstanding decisions:
1. Namespaces
    * Should `Endpoint`+`EndpointMetadataCollection` live in `Microsoft.AspNetCore.Routing`?
    * Should `IEndpointFeature` live in `Microsoft.AspNetCore.Routing.Features`?
2. Package
    * Should these types (+`RouteValueDictionary`) be in Http.Abstractions or Http.Features?

These types flow through a lot of repositories and changing them later will be a pain so lets attempt to make the best decision now. If we don't lock these down online in the PR in the next day I'll organize a quick meeting.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/Endpoint.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <summary>
+        /// Gets the collection of metadata associated with this endpoint.
+        /// </summary>
+        public EndpointMetadataCollection Metadata { get; }
+
+        /// <summary>
+        /// Gets or sets a delegate that can be used to invoke the current
+        /// <see cref="Routing.Endpoint"/>.
+        /// </summary>
+        Func<RequestDelegate, RequestDelegate> Invoker { get; set; }
```
davidfowl: `RequestDelegate`

davidfowl:
Consumption:

- Endpoints are first class (metadata + RequestDelegate) and are not tied to routing
- Accessible via the `IEndpointFeature`
   - Middleware can get the selected `Endpoint` and make decisions based on metadata (or whatever else)
    - We'll invoke the RequestDelegate as the last implicit thing in the middleware chain. Today we set a 404 by default (https://github.com/aspnet/HttpAbstractions/blob/bfa183747f6fb528087554c3d6ec58ef05f1c10a/src/Microsoft.AspNetCore.Http/Internal/ApplicationBuilder.cs#L84). We'll need to figure out where this code goes (is it a startup filter, is it part of the ApplicationBuilder, is it in hosting as an IStartupFilter)
- Exposing the matched routes on `HttpRequest` via a new `IRouteValuesFeature`

```C#
public interface IEndpointFeature
{
    Endpoint Endpoint { get; set; }
}
```

```C#
public interface IRouteValuesFeature
{
    RouteValueDictionary RouteValues { get; set; }
}
```

```C#
public class HttpRequest
{
    RouteValueDictionary RouteValues { get; set; }
}
```

Tratcher:
Put it in AppBuilder and fall back to 404 if it doesn't match.

Hmm, if you do that then Map can end up executing the same endpoint regardless of the route taken. Not sure how expected that would be.

src/Microsoft.AspNetCore.Http/Internal/DefaultHttpRequest.cs
```diff
@@ -151,12 +156,19 @@ public override Task<IFormCollection> ReadFormAsync(CancellationToken cancellati
             return FormFeature.ReadFormAsync(cancellationToken);
         }
 
+        public override RouteValueDictionary RouteValues
+        {
+            get => RouteValuesFeature.Values;
+            set => RouteValuesFeature.Values = value;
```
JamesNK: Is this the right way to add a feature with `HttpRequest`?

src/Microsoft.AspNetCore.Http.Abstractions/IEndpointFeature.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Http.Features
+{
+    /// <summary>
+    /// A feature interface for endpoint routing. Use <see cref="HttpContext.Features"/>
+    /// to access an instance associated with the current request.
```
davidfowl: Remove routing from the description

JamesNK:
* Namespaces updated to `Microsoft.AspNetCore.Http` (except RVD which is being type forwarded to)
* `IRouteValuesFeature` added
* `RouteValues` added to `HttpRequest`

Everything is still in Http.Abstractions and not Http.Endpoints. FYI RequestDelegate lives in Http.Abstractions so if Endpoint was moved to Http.Endpoints then RequestDelegate would need to be forwarded.

src/Microsoft.AspNetCore.Http/Features/RouteValuesFeature.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Routing;
+
+namespace Microsoft.AspNetCore.Http.Features
+{
+    public class RouteValuesFeature : IRouteValuesFeature
+    {
+        private readonly HttpRequest _request;
```
davidfowl: This isn’t correct. Take a look at how the other features are implemented on HttpContext and friends. The HttpContext is a wrapper around features not vice versa

src/Microsoft.AspNetCore.Http/Internal/DefaultHttpRequest.cs
davidfowl: Yep. Passing in the HttpRequest isn’t right though.

src/Microsoft.AspNetCore.Http/Internal/ApplicationBuilder.cs
```diff
@@ -81,6 +81,14 @@ public RequestDelegate Build()
         {
             RequestDelegate app = context =>
             {
+                // If an endpoint has been set then implicitly call its request delegate
+                var requestDelegate = ServerFeatures?.Get<IEndpointFeature>()?.Endpoint?.RequestDelegate;
```
davidfowl: `context.Features.Get<IEndpointFeature>`

test/Microsoft.AspNetCore.Http.Tests/Internal/ApplicationBuilderTests.cs
```diff
@@ -31,5 +33,46 @@ public void PropertiesDictionaryIsDistinctAfterNew()
...
+            var app = builder.Build();
+
+            var httpContext = new DefaultHttpContext();
+
+            app.Invoke(httpContext);
+
+            Assert.True(endpointInvoked);
+        }
+
+        private class TestEndpoint : Endpoint
```
davidfowl: This makes me wonder why we don't just make Endpoint non abstract. It'll remove the need for this.

JamesNK: Now that Endpoint has RequestDelegate on it it does feel a lot less abstract. What do you think @rynowak ?

rynowak: I think it would be fine now that the request delegate has moved down.

test/Microsoft.AspNetCore.Http.Tests/Internal/DefaultHttpRequestTests.cs
```diff
@@ -194,6 +195,17 @@ public void Cookies_GetAndSet()
             Assert.Equal(new[] { "name2=value2" }, cookieHeaders);
         }
 
+        [Fact]
+        public void RouteValues_GetAndSet()
```
davidfowl: Add a test that verifies the feature is read from the feature collection if implemented.

```C#
var features = new FeatureCollection();
features.Set<IRouteValuesFeature>(new RouteValuesFeature() { 
   Values = new RouteValueDictionary( new { A = 1 })
 });
var context = new DefaultHttpContext(features);
Assert.Equal(1, context.Request.RouteValues["A"]);
```

test/Microsoft.AspNetCore.Http.Tests/Internal/ApplicationBuilderTests.cs
```diff
@@ -31,5 +33,37 @@ public void PropertiesDictionaryIsDistinctAfterNew()
...
+            var builder = new ApplicationBuilder(null);
+            var app = builder.Build();
+
+            var httpContext = new DefaultHttpContext();
+            httpContext.Features.Set<IEndpointFeature>(new TestEndpointFeature
+            {
+                Endpoint = endpoint
+            });
+
+            app.Invoke(httpContext);
```
davidfowl: I know this is sync but make this an async test.

JamesNK:
Should `IRouteValuesFeature` be in 2.2? We're going to need to remove route values from `IEndpointFeature` in 2.2 so I'm guessing yes unless we want to keep using `IRoutingFeature` in 2.2.

rynowak:
I could be convinced either way. We don't **need** it in 2.2...... but the other details that we are locking in 2.2 will require the design to be the same. So I think it has the same effect with the only delta being things like naming.

JamesNK:
As do I. Maybe there was some reason it needed to be left out of this PR.

I'll add it back.

davidfowl:
How did it disappear?

PR #1032, opened 2018-08-26
Fix FileBufferingReadStream not working with StreamContent

cnblogs-dudu:
Address dotnet/corefx#31866

Before FileBufferingReadStream is first read, it's length is not properly set. Before StreamContent is read, it's inner stream length is used to compute ContentLength. When they get together in HttpClient, an http request with misleading `Content-Length: 0` is produced.

This PR introduces `_innerLength` to return the length of inner stream before FileBufferingReadStream is first read.

### DIFF
--- a/src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
+++ b/src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
@@ -19,6 +19,7 @@ namespace Microsoft.AspNetCore.WebUtilities
     {
         private const int _maxRentedBufferSize = 1024 * 1024; // 1MB
         private readonly Stream _inner;
+        private readonly long? _innerLength;
         private readonly ArrayPool<byte> _bytePool;
         private readonly int _memoryThreshold;
         private readonly long? _bufferLimit;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
```diff
@@ -84,7 +96,7 @@ public class FileBufferingReadStream : Stream
             string tempFileDirectory)
             : this(inner, memoryThreshold, bufferLimit, tempFileDirectory, ArrayPool<byte>.Shared)
         {
-        }
+        }
```
Tratcher: Revert

src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
```diff
@@ -148,7 +160,11 @@ public override bool CanWrite
 
         public override long Length
         {
-            get { return _buffer.Length; }
+            get
+            {
+                var useInnerLength = _innerLength.HasValue && _innerLength > 0 && !_completelyBuffered && _buffer.Position == 0;
```
Tratcher: If _innerLength.HasValue, why not always use it?

Tratcher:
This is at best a partial fix, it still wouldn't work for chunked requests. The end to end still needs more thought.

src/Microsoft.AspNetCore.WebUtilities/FileBufferingReadStream.cs
cnblogs-dudu: Doing so will break existing tests.

Tratcher: Which? Maybe they should be updated instead, this is a behaviour change.

cnblogs-dudu: See https://travis-ci.org/aspnet/HttpAbstractions/jobs/420667950

Tratcher:
I don't think this is the right place to fix your problem. It only addresses the Content-Length scenario and it makes Content-Length and Chunked behave differently for everybody else. I expect you'd be better of creating a custom HttpContent object that managed this for you.

PR #1033, opened 2018-09-01
Fast-path async in HttpResponseStreamWriter

benaadams:
The ratio of non-async paths in `WriteAsync` of `HttpResponseStreamWriter` to `async` paths (including `FlushAsync`) for the RazorRendering benchmarkapp https://github.com/aspnet/Mvc/pull/8366

For 1 request in the currently its 8774 non-async to 12 async.

image

This adds a fast-path to skip the async statemahchine creation for when no `FlushAsync` takes place.

It also means it can be detected upstream in MVC and allow more fast-paths to flow

/cc @davidfowl @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.WebUtilities/HttpResponseStreamWriter.cs
+++ b/src/Microsoft.AspNetCore.WebUtilities/HttpResponseStreamWriter.cs
@@ -231,12 +231,7 @@ namespace Microsoft.AspNetCore.WebUtilities
                 return GetObjectDisposedTask();
             }
 
-            if (value == null)
-            {
-                return Task.CompletedTask;
-            }
-
-            var count = value.Length;
+            var count = value?.Length ?? 0;
             if (count == 0)
             {
                 return Task.CompletedTask;
--- a/src/Microsoft.AspNetCore.WebUtilities/HttpResponseStreamWriter.cs
+++ b/src/Microsoft.AspNetCore.WebUtilities/HttpResponseStreamWriter.cs
@@ -255,7 +255,7 @@ namespace Microsoft.AspNetCore.WebUtilities
             var count = value.Length;
 
             Debug.Assert(count > 0);
-            Debug.Assert(_charBufferSize - _charBufferCount > count);
+            Debug.Assert(_charBufferSize - _charBufferCount < count);
 
             var index = 0;
             while (count > 0)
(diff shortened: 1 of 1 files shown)


JamesNK:
Looks good to me. How much time is spent in that method for a request? 8000k is a fair number of calls but does the state machine overhead add up to much?

benaadams:
8000k statemachines with all their try/catchery

image

benaadams:
>  8000k is a fair number of calls but does the state machine overhead add up to much?

Also this is for one request; so that's 8bn for 1M requests (though they are larger responses)

src/Microsoft.AspNetCore.WebUtilities/HttpResponseStreamWriter.cs
```diff
@@ -186,22 +220,43 @@ public override async Task WriteAsync(char[] values, int index, int count)
...
+        public override Task WriteAsync(string value)
         {
             if (_disposed)
             {
-                throw new ObjectDisposedException(nameof(HttpResponseStreamWriter));
+                return GetObjectDisposedTask();
             }
 
-            if (value == null)
+            var count = value.Length;
```
campersau: This would now throw if `value` is `null`.

benaadams: fixed

benaadams:
> How much time is spent in that method for a request?

As `CopyToCharBuffer` is the workhorse part of the function; it does looks significant in terms of where the rest of the time is spent in this class:

image

benaadams:
Any takers for a review/approve? /cc @rynowak

rynowak:
OK it.

src/Microsoft.AspNetCore.WebUtilities/HttpResponseStreamWriter.cs
```diff
@@ -186,22 +220,48 @@ public override async Task WriteAsync(char[] values, int index, int count)
...
             {
-                return;
+                return Task.CompletedTask;
+            }
+
+            var count = value.Length;
+            if (count == 0)
+            {
+                return Task.CompletedTask;
+            }
```
rynowak: Is it worth folding this together with L234?

benaadams: Done

JunTaoLuo:
Merged in https://github.com/aspnet/HttpAbstractions/commit/800c79c0e4d3ee7423e9a255ad398e429659912f. Thanks @benaadams !

PR #1037, opened 2018-09-12
Minor perf updates for RVD

rynowak:
Porting changes from perf work in
https://github.com/aspnet/Routing/pull/788

Includes porting/adding the RVD benchmarks, as well as a new TryAdd
method.

See perf results in the linked PR:
Before: https://github.com/aspnet/Routing/pull/788#issue-214056584
After: https://github.com/aspnet/Routing/pull/788#issuecomment-419607738


src/Microsoft.AspNetCore.Http.Abstractions/Routing/RouteValueDictionary.cs
```diff
@@ -167,17 +167,17 @@ public RouteValueDictionary(object values)
 
             set
             {
-                if (string.IsNullOrEmpty(key))
+                if (key == null)
```
rynowak: This was a bug. We have no reason to disallow the empty string as a key, and this is the only place we enforced this.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/RouteValueDictionary.cs
```diff
@@ -456,7 +488,10 @@ private void EnsureCapacitySlow(int capacity)
             if (_propertyStorage != null)
             {
                 var storage = _propertyStorage;
-                capacity = Math.Max(storage.Properties.Length, capacity);
+                
+                // If we're converting from properties, it's likely due to an 'add' to make sure we have at least
+                // the default amount of space.
+                capacity = Math.Max(DefaultCapacity, Math.Max(storage.Properties.Length, capacity));
```
rynowak: This was a bug I discovered while testing TryAdd. If we 'upgrade' from property storage to array we used to allocate the minimum number of slots. Since upgrade is usually paired with adding an item, the common case is that you would allocate and then immediately resize.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/RouteValueDictionary.cs
JamesNK: Test for this (I didn't see a test in the PR)

rynowak: We don't usually test null checks 🤷‍♂️

JamesNK: I mean test that an empty string as the key works to avoid a future regression.

rynowak: added

PR #1040, opened 2018-09-26
Add XML documentation for IFormFileCollection

Marusyk:
Issue https://github.com/aspnet/Home/issues/3528

Please review


Tratcher:
Thanks for the update.

src/Microsoft.AspNetCore.Http.Features/IFormFileCollection.cs
```diff
@@ -10,10 +10,33 @@ namespace Microsoft.AspNetCore.Http
...
+        /// </summary>
+        /// <param name="name">The name of the file to get.</param>
+        /// <returns>
+        ///	The requested file, or null if it is not present.
+        /// </returns>
         IFormFile GetFile(string name);
 
+        /// <summary>
+        ///     Gets an <see cref="IReadOnlyList{T}" /> containing the files of the
+        ///     <see cref="IFormFileCollection" /> with the specified name.
```
JamesNK: If you can have multiple files with the same name, what does GetFile and the indexer do when there are multiple? Return the first one/last one/throw an error?

Tratcher: Returns the first.
https://github.com/aspnet/HttpAbstractions/blob/cbed7393a21483cf25cda77dea8b7cda2b51751f/src/Microsoft.AspNetCore.Http/Internal/FormFileCollection.cs#L20-L26

JamesNK: Ok! Could be worth adding to the XML docs, but that's probably what I'd expect to happen in the rare situation it comes up.

Tratcher: up-for-grabs 😁

Marusyk: Should I prepare a new PR for this?

Eilon: @Marusyk that would be great if you're willing to!

Marusyk:
Related #1041

src/Microsoft.AspNetCore.Http.Features/IFormFileCollection.cs
Marusyk: https://github.com/aspnet/HttpAbstractions/pull/1041

PR #1041, opened 2018-09-27
Fix XML documentation for IFormFileCollection

Marusyk:
Issue https://github.com/aspnet/Home/issues/3528

Please review

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Features/IFormFileCollection.cs
+++ b/src/Microsoft.AspNetCore.Http.Features/IFormFileCollection.cs
@@ -11,7 +11,7 @@ namespace Microsoft.AspNetCore.Http
     public interface IFormFileCollection : IReadOnlyList<IFormFile>
     {
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Http.Features/IFormFileCollection.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.AspNetCore.Http
     public interface IFormFileCollection : IReadOnlyList<IFormFile>
     {
         /// <summary>
-        /// Gets the file with the specified name.
+        /// Gets the first file in a collection that satisfies a specified name.
```
JamesNK: What do you think of:
> Gets the file with the specified name. If there are multiple files with this name then the first match is returned.

Marusyk: Yes, thanks. That would be better. I'll change

Eilon:
@Marusyk thank you again!

PR #1042, opened 2018-10-01
Add RoutesValues to HttpRequest

JamesNK:
RouteValues disappeared at some point from https://github.com/aspnet/HttpAbstractions/pull/1030

This is additive so should be simple to merge.

### DIFF
--- a/src/Microsoft.AspNetCore.Http/Internal/DefaultHttpRequest.cs
+++ b/src/Microsoft.AspNetCore.Http/Internal/DefaultHttpRequest.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Routing;
 using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.Http.Internal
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Http/Internal/DefaultHttpRequest.cs
```diff
@@ -151,12 +156,19 @@ public override Task<IFormCollection> ReadFormAsync(CancellationToken cancellati
             return FormFeature.ReadFormAsync(cancellationToken);
         }
 
+        public override RouteValueDictionary RouteValues
```
Tratcher: We should avoid returning a null value here. If the feature doesn't exist, or the value is null then an empty collection should be created, stored, and returned. We do something similar for cookies, query, etc..

test/Microsoft.AspNetCore.Http.Tests/Internal/DefaultHttpRequestTests.cs
```diff
@@ -194,6 +195,54 @@ public void Cookies_GetAndSet()
...
+        {
+            var context = new DefaultHttpContext();
+            var routeValuesFeature = context.Features.Get<IRouteValuesFeature>();
+            // No feature set for initial DefaultHttpRequest
+            Assert.Null(routeValuesFeature);
+
+            var request = context.Request;
+            var routeValues = request.RouteValues;
+            // Route values defaults to null
+            Assert.Null(routeValues);
```
Tratcher: Yeah, this shouldn't be null. The caller shouldn't have to check for null before looking up a value in the collection.

JamesNK:
🆙 📅

// @Tratcher

PR #1043, opened 2018-10-03
Add IHttpResponseTrailersFeature and extensions

Tratcher:
Part of https://github.com/aspnet/KestrelHttpServer/issues/622

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/ResponseTrailerExtensions.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/ResponseTrailerExtensions.cs
@@ -1,0 +1,53 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.Http
+{
+    public static class ResponseTrailerExtensions
+    {
+        private const string Trailer = "Trailer";
+
+        /// <summary>
+        /// Adds the given trailer name to the 'Trailer' response header. This must happen before the response headers are sent.
+        /// </summary>
+        /// <param name="response"></param>
+        /// <param name="trailerName"></param>
+        public static void DeclareTrailer(this HttpResponse response, string trailerName)
+        {
+            response.Headers.AppendCommaSeparatedValues(Trailer, trailerName);
+        }
+
+        /// <summary>
+        /// Indicates if the server supports sending trailer headers for this response.
+        /// </summary>
+        /// <param name="response"></param>
+        /// <returns></returns>
+        public static bool SupportsTrailers(this HttpResponse response)
+        {
+            var feature = response.HttpContext.Features.Get<IHttpResponseTrailersFeature>();
+            return feature?.Trailers != null && !feature.Trailers.IsReadOnly;
+        }
+
+        /// <summary>
+        /// Adds the given trailer header to the trailers collection to be sent at the end of the response body.
+        /// Check <see cref="SupportsTrailers" /> or an InvalidOperationException may be thrown.
+        /// </summary>
+        /// <param name="response"></param>
+        /// <param name="trailerName"></param>
+        /// <param name="trailerValues"></param>
+        public static void AppendTrailer(this HttpResponse response, string trailerName, StringValues trailerValues)
+        {
+            var feature = response.HttpContext.Features.Get<IHttpResponseTrailersFeature>();
+            if (feature?.Trailers == null || feature.Trailers.IsReadOnly)
+            {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Http.Abstractions/Extensions/ResponseTrailerExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public static void DeclareTrailer(this HttpResponse response, string trailerName)
+        {
+            response.Headers.AppendCommaSeparatedValues(Trailer, trailerName);
+        }
+
+        public static void AppendTrailer(this HttpResponse response, string trailerName, StringValues trailerValues)
+        {
+            var feature = response.HttpContext.Features.Get<IHttpResponseTrailersFeature>();
+            feature?.Trailers?.Append(trailerName, trailerValues);
```
halter73: Checking for null when you're already checking for the feature existence seems like it might be a little too defensive. Technically response.Headers could be set to null too, but there's no check in DeclareTrailer.

Is the idea that the server might set the feature but leave the IHeaderDictionary null? Or are you just defending against some middleware setting Trailers to null?

Tratcher: Just paranoia. We make a lot of assumptions about the request and response features because nothing would work without them, but for these very optional features I try not to assume anything.

davidfowl: Is it weird that this can noop? Should it throw if trailers aren't supported?

Tratcher: If it threw you'd need some other way to check for support before calling it. Checking for the presence of the feature is the simplest, but negates using an extension like this. I could add another extension here SupportsTrailers to check for the feature.

We also discussed some Http/1.1 scenarios where the feature may be present but the headers may be discarded with a log message (e.g. content-length)

src/Microsoft.AspNetCore.Http.Features/IHttpResponseTrailersFeature.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Http.Features
+{
+    public interface IHttpResponseTrailersFeature
+    {
+        IHeaderDictionary Trailers { get; set; }
```
JunTaoLuo: Is there a reason why there's a setter here?

JunTaoLuo: Ah nevermind I see the comment on the Kestrel PR.

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/ResponseTrailerExtensions.cs
davidfowl: Err hmm, I'm not sure if I like that, it's very similar to adding headers after the response body has started. Today we throw. I like that model.

jtattermusch:
CC @jtattermusch

Tratcher:
Updated to throw if Trailers can't be added.

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/ResponseTrailerExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// Adds the given trailer header to the trailers collection to be sent at the end of the response body.
+        /// Check <see cref="SupportsTrailers" /> or an InvalidOperationException may be thrown.
+        /// </summary>
+        /// <param name="response"></param>
+        /// <param name="trailerName"></param>
+        /// <param name="trailerValues"></param>
+        public static void AppendTrailer(this HttpResponse response, string trailerName, StringValues trailerValues)
+        {
+            var feature = response.HttpContext.Features.Get<IHttpResponseTrailersFeature>();
+            if (feature?.Trailers == null || feature.Trailers.IsReadOnly)
```
JunTaoLuo: You could extract this into a private method that takes the feature and returns a bool and use it in both SupportsTrailers and AppendTrailer calls.

Tratcher: *shrug* It's only one if.

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/ResponseTrailerExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                throw new InvalidOperationException("Trailers are not supported for this response.");
```
JunTaoLuo: Nit: Could potentially consider adding a test for this.

Issue #1049, opened 2018-10-13
GetSetCookieLength method is not parsing correctly when cookie has invalid keyvalue pair


dhmuralikrishna:
Lets say set-cookie value is 
cookiename=cookievalue; Version=1; Domain=.contoso.com; Path=/; Secure; HttpOnly";

Then it's not parsing to SetcookieHeaderValue in  SetcookieHeaderValue.TryParse is returning null.

Version =1 might not be valid as per the spec. But other non-windows servers are sending the cookie like that. 

Expected Ignore Version =1 and fill remaining values in SetCookieHeaderValue.

Tratcher:
Where are you getting cookies like this from?

dhmuralikrishna:
From one of the dependent systems.

dhmuralikrishna:
I think its java app don't know much details. we are consuming https endpoint int he response set-cookie it has version=1

PR #1050, opened 2018-10-13
Fixed Issues :1049.

dhmuralikrishna:
https://github.com/aspnet/HttpAbstractions/issues/1049


src/Microsoft.Net.Http.Headers/SetCookieHeaderValue.cs
```diff
@@ -452,9 +452,16 @@ private static int GetSetCookieLength(StringSegment input, int startIndex, out S
...
-                    // TODO: skip it? Store it in a list?
+                else 
+                {   
+                    // TODO: skiping it for now to avoid parsing failure? Store it in a list?
+                    // = (no spaces)
+                    if (!ReadEqualsSign(input, ref offset))
+                    {
+                        return 0;
+                    }
+                    // We don't do any detailed validation on the domain.
```
Tratcher: Can you clean up these comments? They look like they were copied from elsewhere.

PR #1051, opened 2018-10-17
Port aspnet/Routing/pull/858

rynowak:


src/Microsoft.AspNetCore.Http.Abstractions/Routing/RouteValueDictionary.cs
```diff
@@ -388,19 +388,62 @@ public bool Remove(string key)
                 return false;
             }
 
-            EnsureCapacity(Count);
+            // Ensure property storage is converted to array storage as we'll be
+            // applying the lookup and removal on the array
+            EnsureCapacity(_count);
```
rynowak: Diff is wierd. The change to this method is to use the field directly and add the comment.

PR #1052, opened 2018-10-22
Add safe copy for enumeration

HaoK:
Possible fix for https://github.com/aspnet/AspNetCore/issues/3619

@Eilon @Tratcher @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Options;
@@ -49,6 +51,9 @@ namespace Microsoft.AspNetCore.Authentication
 
         private readonly IDictionary<string, AuthenticationScheme> _schemes;
         private readonly List<AuthenticationScheme> _requestHandlers;
+        // Used as a safe return value for enumeration apis
+        private IEnumerable<AuthenticationScheme> _schemesCopy = new AuthenticationScheme[0];
+        private IEnumerable<AuthenticationScheme> _requestHandlersCopy = new AuthenticationScheme[0];
 
         private Task<AuthenticationScheme> GetDefaultSchemeAsync()
             => _options.DefaultScheme != null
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -49,6 +51,9 @@ protected AuthenticationSchemeProvider(IOptions<AuthenticationOptions> options,
 
         private readonly IDictionary<string, AuthenticationScheme> _schemes;
         private readonly List<AuthenticationScheme> _requestHandlers;
+        // Used as a safe return value for enumeration apis
+        private IEnumerable<AuthenticationScheme> _schemesCopy = new AuthenticationScheme[0];
+        private IEnumerable<AuthenticationScheme> _requestHandlersCopy = new AuthenticationScheme[0];
```
poke: Why no love for `Array.Empty<AuthenticationScheme>()`? 😄

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -144,8 +149,10 @@ public virtual void AddScheme(AuthenticationScheme scheme)
                 if (typeof(IAuthenticationRequestHandler).IsAssignableFrom(scheme.HandlerType))
                 {
                     _requestHandlers.Add(scheme);
+                    _requestHandlersCopy = _requestHandlers.ToList();
```
poke: Wouldn’t `.ToArray()` be a bit lighter here since we only use the result as an enumerable source? Or am I micro-optimizing too much?

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -164,13 +171,19 @@ public virtual void RemoveScheme(string name)
                 if (_schemes.ContainsKey(name))
                 {
                     var scheme = _schemes[name];
-                    _requestHandlers.Remove(scheme);
-                    _schemes.Remove(name);
+                    if (_requestHandlers.Remove(scheme))
+                    {
+                        _requestHandlersCopy = _requestHandlers.ToList();
+                    }
+                    if (_schemes.Remove(name))
```
poke: This check is redundant because of the `_schemes.ContainsKey(name)` check inside the lock above. We already know that the scheme will be removed.

src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs
```diff
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
```
poke: nit: This using shouldn’t be needed for this change.

Tratcher (approved):
+ what poke said.

HaoK:
Updated

poke (approved):

poke (approved):

PR #1053, opened 2018-10-23
Add support for HttpCookie expires date format

MerickOWA:
The format generated by System.Web.HttpCookie when formated is non-standard, a mix between RFC1123 and RFC850 (At least as of .NET Framework 4.6.1)

The exact format string used is `"ddd, dd-MMM-yyyy HH':'mm':'ss 'GMT'"` as seen [here](http://referencesource.microsoft.com/#System.Web/httpserverutility.cs,1526)

This doesn't quite match RFC1123 which is `"ddd, d MMM yyyy H:m:s 'GMT'"` as it uses dashes between the day, month, and year instead of spaces, but doesn't quite match RFC850 which is `"dddd, d'-'MMM'-'yy H:m:s 'GMT'"` as it uses a short day of the week instead of a longer form.

This change allows the SetCookieHeaderValue to properly parse a response from the an API using the older System.Web.HttpCookie class to generate a HTTP response.

### DIFF
--- a/src/Microsoft.Net.Http.Headers/HttpRuleParser.cs
+++ b/src/Microsoft.Net.Http.Headers/HttpRuleParser.cs
@@ -33,6 +33,8 @@ namespace Microsoft.Net.Http.Headers
             "ddd, d MMM yyyy H:m:s", // RFC 5322 no zone
             "d MMM yyyy H:m:s zzz", // RFC 5322 no day-of-week
             "d MMM yyyy H:m:s", // RFC 5322 no day-of-week, no zone
+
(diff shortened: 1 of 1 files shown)


Tratcher:
Interesting find 
1. Can you add an entry in the tests as well?
1. Please file a bug for tracking purposes.
1. Can you rebase on the release/2.2 branch? That's the next branch we're shipping. Otherwise this won't release until 3.0.

MerickOWA:
Yep, I'll close this one and reopen once I have the tests written and bug filed.

PR #1057, opened 2018-10-30
Do not throw a null reference from request.GetDisplayUrl()

filipw:
This is based on a customer bug. It is possible for a middleware and other participants in the request pipeline to modify things like properties of the request (i.e. `PathBase`).
Then, when `request.GetDisplayUrl()` is called on the same request, there are no null checks on the request properties needed to compute the URL, which in extreme cases causes it to throw an unhandled exception.

One example of running into this, is a following middleware.

```
            app.Use((context, next) =>
            {
                context.Request.PathBase = new PathString(context.Request.Headers["x-forwarded-prefix"]);
                return next();
            });
```

When issuing a request without `x-forwarded-prefix` that then calls into `request.GetDisplayUrl()`, we get a null ref.

While in this case it can be argued that the middleware should validate what it is actually doing, in general I don't think a framework method like `request.GetDisplayName()` should ever throw a null reference.

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
+++ b/src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
@@ -195,16 +195,16 @@ namespace Microsoft.AspNetCore.Http.Extensions
         /// <returns></returns>
         public static string GetDisplayUrl(this HttpRequest request)
         {
-            var host = request.Host.Value;
-            var pathBase = request.PathBase.Value;
-            var path = request.Path.Value;
-            var queryString = request.QueryString.Value;
+            var host = request.Host.Value ?? "";
+            var pathBase = request.PathBase.Value ?? "";
+            var path = request.Path.Value ?? "";
+            var queryString = request.QueryString.Value ?? "";
 
             // PERF: Calculate string length to allocate correct buffer size for StringBuilder.
-            var length = request.Scheme?.Length ?? 0 + SchemeDelimiter.Length + host?.Length ?? 0
-                + pathBase?.Length ?? 0 + path?.Length ?? 0 + queryString?.Length ?? 0;
+            var length = request.Scheme.Length + SchemeDelimiter.Length + host.Length
+                + pathBase.Length + path.Length + queryString.Length;
 
-            return new StringBuilder(length)
+            return new StringBuilder()
                 .Append(request.Scheme)
                 .Append(SchemeDelimiter)
                 .Append(host)
--- a/src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
+++ b/src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
@@ -195,17 +195,18 @@ namespace Microsoft.AspNetCore.Http.Extensions
         /// <returns></returns>
         public static string GetDisplayUrl(this HttpRequest request)
         {
+            var scheme = request.Scheme ?? "";            
             var host = request.Host.Value ?? "";
             var pathBase = request.PathBase.Value ?? "";
             var path = request.Path.Value ?? "";
             var queryString = request.QueryString.Value ?? "";
 
             // PERF: Calculate string length to allocate correct buffer size for StringBuilder.
-            var length = request.Scheme.Length + SchemeDelimiter.Length + host.Length
+            var length = scheme.Length + SchemeDelimiter.Length + host.Length
                 + pathBase.Length + path.Length + queryString.Length;
 
             return new StringBuilder(length)
-                .Append(request.Scheme)
+                .Append(scheme)
                 .Append(SchemeDelimiter)
                 .Append(host)
                 .Append(pathBase)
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
```diff
@@ -201,8 +201,8 @@ public static string GetDisplayUrl(this HttpRequest request)
             var queryString = request.QueryString.Value;
 
             // PERF: Calculate string length to allocate correct buffer size for StringBuilder.
-            var length = request.Scheme.Length + SchemeDelimiter.Length + host.Length
-                + pathBase.Length + path.Length + queryString.Length;
+            var length = request.Scheme?.Length ?? 0 + SchemeDelimiter.Length + host?.Length ?? 0
+                + pathBase?.Length ?? 0 + path?.Length ?? 0 + queryString?.Length ?? 0;
```
poke: That’s a lot of null checks in a calculation that is done for “PERF” reasons :confused:

filipw: you are right I improved the code and it should be better now, there appears to be no impact at all on perf

```
BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17134.345 (1803/April2018Update/Redstone4)
Intel Core i7-8850H CPU 2.60GHz (Max: 2.59GHz) (Coffee Lake), 1 CPU, 6 logical and 6 physical cores
.NET Core SDK=2.2.100-preview2-009404
  [Host]     : .NET Core 2.1.5 (CoreCLR 4.6.26919.02, CoreFX 4.6.26919.02), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.5 (CoreCLR 4.6.26919.02, CoreFX 4.6.26919.02), 64bit RyuJIT


                Method |     Mean |    Error |    StdDev |   Median |
---------------------- |---------:|---------:|----------:|---------:|
      GetDisplayUrlOld | 895.2 ns | 39.06 ns | 112.68 ns | 832.6 ns |
      GetDisplayUrlNew | 854.9 ns | 20.18 ns |  52.08 ns | 840.2 ns |
```

src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
```diff
@@ -195,16 +195,16 @@ public static string GetEncodedPathAndQuery(this HttpRequest request)
...
+            var pathBase = request.PathBase.Value ?? "";
+            var path = request.Path.Value ?? "";
+            var queryString = request.QueryString.Value ?? "";
 
             // PERF: Calculate string length to allocate correct buffer size for StringBuilder.
             var length = request.Scheme.Length + SchemeDelimiter.Length + host.Length
                 + pathBase.Length + path.Length + queryString.Length;
 
-            return new StringBuilder(length)
+            return new StringBuilder()
```
Tratcher: Revert, that's the whole point in pre-calculating the length.

filipw: oops you are right, I edited on GH directly and copy pasted the wrong thing 🤠

src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
```diff
@@ -195,10 +195,10 @@ public static string GetEncodedPathAndQuery(this HttpRequest request)
...
-            var pathBase = request.PathBase.Value;
-            var path = request.Path.Value;
-            var queryString = request.QueryString.Value;
+            var host = request.Host.Value ?? "";
+            var pathBase = request.PathBase.Value ?? "";
+            var path = request.Path.Value ?? "";
+            var queryString = request.QueryString.Value ?? "";
 
             // PERF: Calculate string length to allocate correct buffer size for StringBuilder.
             var length = request.Scheme.Length + SchemeDelimiter.Length + host.Length
```
Tratcher: Might as well make the same change for scheme, it's the only one not following the pattern.

src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs
```diff
@@ -195,17 +195,18 @@ public static string GetEncodedPathAndQuery(this HttpRequest request)
         /// <returns></returns>
         public static string GetDisplayUrl(this HttpRequest request)
         {
-            var host = request.Host.Value;
-            var pathBase = request.PathBase.Value;
-            var path = request.Path.Value;
-            var queryString = request.QueryString.Value;
+            var scheme = request.Scheme.Value ?? "";
```
JamesNK: nit: `string.Empty` vs `""`

I prefer string.Empty but I'm not sure what the standard is for this repo.

filipw: makes sense, they will both be interned but indeed, it looks like `string.Empty` is used everywhere here

Tratcher:
Thanks

PR #1058, opened 2018-11-03
Add EndpointModel and IEndpointConventionBuilder

JamesNK:
Fixes https://github.com/aspnet/Routing/issues/899

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/EndpointModel.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/EndpointModel.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    /// <summary>
+    /// A base class for modeling <see cref="Endpoint"/>.
+    /// </summary>
+    public abstract class EndpointModel
+    {
+        /// <summary>
+        /// Gets or sets the delegate used to process requests for the endpoint.
+        /// </summary>
+        public RequestDelegate RequestDelegate { get; set; }
+
+        /// <summary>
+        /// Gets or sets the informational display name of this endpoint.
+        /// </summary>
+        public string DisplayName { get; set; }
+
+        /// <summary>
+        /// Gets the collection of metadata associated with this endpoint.
+        /// </summary>
+        public IList<object> Metadata { get; } = new List<object>();
+
+        /// <summary>
+        /// Creates an instance of <see cref="Endpoint"/> from the <see cref="EndpointModel"/>.
+        /// </summary>
+        /// <returns>The created <see cref="Endpoint"/>.</returns>
+        public abstract Endpoint Build();
+    }
+}
--- a/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/IEndpointModelConvention.cs
+++ b/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/IEndpointModelConvention.cs
@@ -1,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    /// <summary>
+    /// Allows customization of the <see cref="EndpointModel"/>.
+    /// </summary>
+    /// <remarks>
+    /// This interface is used at application startup to customize endpoints for the application.
+    /// </remarks>
+    public interface IEndpointConventionBuilder
+    {
+        /// <summary>
+        /// Called to apply the convention to the <see cref="EndpointModel"/>.
+        /// </summary>
+        /// <param name="convention">The convention to apply to the <see cref="EndpointModel"/>.</param>
+        void Apply(Action<EndpointModel> convention);
+    }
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNetCore.Http.Abstractions/Extensions/EndpointModel.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    /// <summary>
+    /// A base class for modeling <see cref="Endpoint"/>.
+    /// </summary>
+    public abstract class EndpointModel
```
davidfowl: Why doesn’t the builder have the Build method? Why are these different types?

davidfowl (changes requested):

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/EndpointModel.cs
JamesNK: EndpointModel is used to create one endpoint, e.g. an action on a controller.

IEndpointConventionBuilder is used to register conventions that are applied to multiple endpoints as they are being created. The models are passed to `Action<EndpointModel>` that applies the convention changes, e.g. all endpoints created by MVC with a conventional route pattern should have IAuthorizeData metadata.

JamesNK: Thought about how to explain this more:
* EndpointModel: mutable version of Endpoint that are created when a datasource initializes. Needed because we want to apply changes to endpoints as they are built with a convention
* IEndpointConventionBuilder: start up time way of registering conventions. These conventions are then run when a datasource initializes. Data source creates models, conventions run on them, then models are turned into immutable endpoints

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/IEndpointModelConvention.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.AspNetCore.Builder
+{
+    /// <summary>
+    /// Allows customization of the <see cref="EndpointModel"/>.
+    /// </summary>
+    /// <remarks>
+    /// This interface is used at application startup to customize endpoints for the application.
+    /// </remarks>
+    public interface IEndpointConventionBuilder
```
Tratcher: IEndpointConvention~~Builder~~?

JamesNK: Yes the action is the convention. Example of it being used:

```cs
public static class EndpointConventionBuilderExtensions
{
    public static IEndpointConventionBuilder RequireAuthorization(this IEndpointConventionBuilder builder, params string[] roles)
    {
        builder.Apply(endpoint => endpoint.Metadata.Add(new AuthorizationMetadata(roles)));
        return builder;
    }
}
```

Sure you could replace the action with an interface. I don't see any benefit from that though, and creating a new type and having it keep state has a lot of overhead compared to a lambda + closure. Right now it is very easy to create an extension method (or use Apply directly when building endpoints) to customize endpoints.

JamesNK:
I've responded to all questions. I would like to get this PR moving.

rynowak:
After some personal reflection, I'm not 100% sure this is baked. We designed these APIs based on the 2.2 version of MVC's action discovery.

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/EndpointModel.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Gets or sets the informational display name of this endpoint.
+        /// </summary>
+        public string DisplayName { get; set; }
+
+        /// <summary>
+        /// Gets the collection of metadata associated with this endpoint.
+        /// </summary>
+        public IList<object> Metadata { get; } = new List<object>();
```
drieseng: Is there are specific reason to only expose the interface (except for making an abstraction of the underlying implementation)? This isn't a standalone case, I know.

Now that devirtualization is implemented in the JIT this may no longer be a problem though.

JamesNK: It's standard to expose an interface publicly in case the internal implementation changes.

For cases like this it doesn't make a huge amount of sense, but this also isn't a type that will be used when perf is important 🤷‍♂️

src/Microsoft.AspNetCore.Http.Abstractions/Extensions/EndpointModel.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Builder
```
rynowak: Do we know the reason why these need to be in the HttpAbstractions? I think it's for the following reasons:
1. So that middleware can use the builder to become routerware
2. So that cross-cutting concerns like auth don't have to depend on routing

Is this valuable? I'm not sure that it matters what repo we define this stuff in.

rynowak (approved):

natemcmaster:
I'm about to merge this repo to aspnet/AspNetCore. Should this PR be merged today? If not, the PR will need to be re-opened against the new repo.

JamesNK:
Ill rebase it post merge

natemcmaster:
Ok, thanks.

PR #1059, opened 2018-11-05
Implicitly execute matched endpoint at the end of middleware pipeline

JamesNK:
Removes the need for `app.UseEndpoint()`. If an endpoint is matched in UseEndpointRouting then reaching the end of the pipeline will automatically execute it.

```cs
public void Configure(IApplicationBuilder app)
{
    app.UseEndpointRouting();

    app.UseStaticFiles();

    // Imagine some more stuff here...

    app.UseEndpoint();
}
```


src/Microsoft.AspNetCore.Http/Internal/ApplicationBuilder.cs
```diff
@@ -81,6 +81,13 @@ public RequestDelegate Build()
         {
             RequestDelegate app = context =>
             {
+                // Implicitly execute matched endpoint at the end of the pipeline instead of returning 404
+                var endpointRequestDelegate = context.Features.Get<IEndpointFeature>()?.Endpoint?.RequestDelegate;
```
Tratcher: add a context.GetEndpoint() extension. See https://github.com/aspnet/AuthSamples/pull/65#discussion_r231239371

PR #1060, opened 2018-11-06
Add Endpoint extension methods to HttpContext

JamesNK:
re: https://github.com/aspnet/AuthSamples/pull/65#discussion_r231239371

Are the extension methods in the right package/namespace?


test/Microsoft.AspNetCore.Http.Abstractions.Tests/EndpointHttpContextExtensionsTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            context.SetEndpoint(endpoint);
+
+            // Assert
+            var feature = context.Features.Get<IEndpointFeature>();
+            Assert.Equal(initialFeature, feature);
+            Assert.Equal(endpoint, feature.Endpoint);
+        }
+
+        [Fact]
+        public void SetEndpoint_NullOnContextWithFeature_NullSetOnExistingFeature()
```
Tratcher: You never actually call SetEndpoint and GetEndpoint in the same test to ensure it round trips.

src/Microsoft.AspNetCore.Http.Abstractions/Routing/EndpointHttpContextExtensions.cs
```diff
@@ -0,0 +1,70 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Http.Features;
+
+namespace Microsoft.AspNetCore.Http
```
davidfowl: Why do we need this to show up in all applications?

JamesNK: What do you want instead?

JamesNK: Getting the endpoint is relatively easy. Setting requires creating a class that implements `IEndpointFeature`. Making people implement an interface to set an endpoint feels like a pit of failure.

Both of these helper methods makes it feel much easier:

`context.GetEndpoint()` and `context.SetEndpoint(endpoint)`

vs

 `context.Features.Get<IEndpointFeature>()?.Endpoint` and `context.Features.Set<IEndpointFeature>(new EndpointFeature { Endpoint = endpoint })`

davidfowl: Can we change the namespace then?

Tratcher: Quite a few middleware will be using this soon. We have sometimes move extensions out of the expected namespace as a means of de-emphasizing them. I don't see any harm in keeping this one, nor have any suggestions for an alternate namespace.

davidfowl: I disagree. 4 middleware will be using it.  A brand new one that we wrote (Authorization), static files (maybe), CORS, Response caching. Putting extension methods on HttpContext. 

We should not have this by default on HttpContext in every ASP.NET Core application. It just doesn't need to be that first class.

JamesNK: There is a Microsoft.AspNetCore.Http.Extensions project it could move to. The extension methods in it still live in the `Microsoft.AspNetCore.Http` namespace.

Tratcher: It's the namespace that's the problem, not the package.

Actually I expect this to get used in most of the middleware if only to access the metadata. I know Authentication needs it, and likely response compression, buffering, etc..

davidfowl: Change the namespace and I'll be happy.

JamesNK: I can't think of another namespace. All the endpoint types live in Microsoft.AspNetCore.Http.

davidfowl: Those extensions are worthy of being that top level, this one isn’t. Add something else to the namespace like Endpoints

JamesNK: It would be alone in Microsoft.AspNetCore.Http.Endpoints but it makes the most amount of sense.

Tratcher: Sure, go for it.

PR #1063, opened 2018-11-10
Use Nullable<T>.GetValueOrDefault() instead of Nullable<T>.Value

drieseng:
Use `Nullable<T>.GetValueOrDefault()` instead of `Nullable<T>.Value` when it is known to have a value.
This bring us a minor performance increase, and reduces code size.

When - for example - applied to the `RangeItemHeaderValue(long? from, long? to)` ctor, this gets us the following result:

**Before:**

|    Method |     Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|---------- |---------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
| FromAndTo | 5.989 ns | 0.0848 ns | 0.0708 ns |      0.0114 |           - |           - |                48 B |

**After:**

|    Method |     Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|---------- |---------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
| FromAndTo | 5.605 ns | 0.0872 ns | 0.0773 ns |      0.0114 |           - |           - |                48 B |

PR #1064, opened 2018-11-10
Do not check if key is present before removing item in AuthenticationProperties

drieseng:
* Do not check if key is present before removing item, this eliminates double lookup.
* Use `Nullable<T>.GetValueOrDefault()` instead of `Nullable<T>.Value` when it is known to have a value. See #1063 for more info.
* Add some unit tests.

Eliminating the extra lookup yields the following result:

**Before:**

|                  Method |      Mean |     Error |    StdDev |         Op/s |  Gen 0 | Allocated |
|------------------------ |----------:|----------:|----------:|-------------:|-------:|----------:|
|    NullValue_KeyPresent | 123.84 ns | 2.1405 ns | 1.8975 ns |  8,074,763.9 | 0.0036 |     328 B |
| NullValue_KeyNotPresent |  41.63 ns | 0.5963 ns | 0.5577 ns | 24,022,181.2 | 0.0021 |     192 B |

**After:**

|                  Method |      Mean |     Error |    StdDev |         Op/s |  Gen 0 | Allocated |
|------------------------ |----------:|----------:|----------:|-------------:|-------:|----------:|
|    NullValue_KeyPresent | 107.04 ns | 0.9192 ns | 0.8598 ns |  9,342,589.5 | 0.0038 |     328 B |
| NullValue_KeyNotPresent |  39.53 ns | 0.5507 ns | 0.4882 ns | 25,300,109.1 | 0.0021 |     192 B |

Benchmark is available [here](https://gist.github.com/drieseng/96ad8c2680b3ac5b84b6fc2ac0bae100).

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationProperties.cs
@@ -122,7 +122,7 @@ namespace Microsoft.AspNetCore.Authentication
             {
                 Items[key] = value;
             }
-            else if (Items.ContainsKey(key))
+            else
             {
                 Items.Remove(key);
             }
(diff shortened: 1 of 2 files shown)


drieseng:
@poke That's knowledge you gain for free by aging :-)

drieseng:
@Tratcher I just happened to stumble upon it. I didn't come up in profiling session or anything.

poke:
@Tratcher I would assume that this was motivated by the `Value` to `ValueOrDefault()` change and then the key checks were seen in the code.. 😄

PR #1065, opened 2018-11-13
Adds PipeWriterAdapter

jkotalik:
Half of https://github.com/aspnet/AspNetCore/issues/3966.

### Few questions:

- What should we do with OnReaderCompleted? If we implement it, it means that we would probably need to create a PipeReaderAdapter (or somehow make a shared object).
- I think I need to synchronize FlushAsync and CancelPendingFlush

### Tests

- I took tests from https://github.com/dotnet/corefx/blob/master/src/System.IO.Pipelines/tests/PipeWriterTests.cs for all writer behavior.
- It will be difficult to directly port any tests that use both PipeReader and PipeWriter with the adapter implementations. This is because having a PipeReader/PipeWriter with a shared backing stream will not work because it would require seeking/setting positions to make it work. 
- I also added tests for cancellation

### Performance

Current performance numbers:

|                                  Method |          Mean |      Error |     StdDev |        Median |       Op/s | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|---------------------------------------- |--------------:|-----------:|-----------:|--------------:|-----------:|------------:|------------:|------------:|--------------------:|
|                         WriteHelloWorld |      6.280 us |  0.1119 us |   1.033 us |      5.965 us | 159,246.75 |           - |           - |           - |             4.72 KB |
|      WriteHelloWorldLargeNumberOfWrites |  2,515.595 us | 23.2819 us | 222.184 us |  2,512.710 us |     397.52 |           - |           - |           - |          4256.74 KB |
|               WriteHelloWorldLargeWrite |     22.644 us |  0.3660 us |   3.410 us |     22.165 us |  44,162.09 |           - |           - |           - |           117.45 KB |
| WriteHelloWorldLargeNumberOfLargeWrites | 16,120.442 us | 70.6716 us | 652.178 us | 15,988.595 us |      62.03 |   1000.0000 |   1000.0000 |   1000.0000 |         28904.27 KB |


benchmarks/Microsoft.AspNetCore.Http.Performance/PipeWriterAdapterBenchmark.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        [Benchmark]
+        public async Task WriteHelloWorldLargeWrite()
+        {
+            await _pipeWriter.WriteAsync(_largeWrite);
+        }
+
+
+        [Benchmark]
+        public async Task WriteHelloWorldLargeNumberOfLargeWrites()
```
pakrym: What does this benchmark show that `WriteHelloWorldLargeWrite` doesn't?

pakrym: Same for `WriteHelloWorldLargeNumberOfWrites`

build/dependencies.props
```diff
@@ -3,7 +3,7 @@
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
   </PropertyGroup>
   <PropertyGroup Label="Package Versions">
-    <BenchmarkDotNetPackageVersion>0.10.13</BenchmarkDotNetPackageVersion>
+    <BenchmarkDotNetPackageVersion>0.11.2</BenchmarkDotNetPackageVersion>
```
pakrym: Has to be done globally in AspNetCore repo.

src/Microsoft.AspNetCore.Http/PipeWriterAdapter.cs
```diff
@@ -0,0 +1,287 @@
...
+    /// Implements PipeWriter using a base stream implementation. 
+    /// </summary>
+    public class PipeWriterAdapter : PipeWriter, IDisposable
+    {
+        private readonly int _minimumSegmentSize;
+        private readonly Stream _writingStream;
+        private int _bytesWritten;
+
+        private List<CompletedBuffer> _completedSegments;
+        private byte[] _currentSegment;
```
pakrym: Memory<byte>

src/Microsoft.AspNetCore.Http/PipeWriterAdapter.cs
```diff
@@ -0,0 +1,287 @@
...
+        private async ValueTask<FlushResult> FlushAsyncInternal(CancellationToken cancellationToken = default)
+        {
+            if (_bytesWritten == 0)
+            {
+                return new FlushResult(isCanceled: false, IsCompletedOrThrow());
+            }
+
+            // Wrap the provided cancellationToken with an internal one
+            // to allow CancelPendingFlush to cancel writes and flushes
+            if (_currentFlushCanceled)
```
pakrym: Can you check if InternalTokenSource  was cancelled?

src/Microsoft.AspNetCore.Http/PipeWriterAdapter.cs
```diff
@@ -0,0 +1,287 @@
...
+                await _writingStream.FlushAsync(joinedToken);
+            }
+            catch (OperationCanceledException)
+            {
+                // Catch any cancellation and translate it into setting isCanceled = true
+                return new FlushResult(isCanceled: true, IsCompletedOrThrow());
+            }
+
+            // After writing to the stream, we can return all ArrayPool segments that are completed.
+            ReturnCompletedBuffers();
```
pakrym: Why not do it inline with writing?

src/Microsoft.AspNetCore.Http/PipeWriterAdapter.cs
```diff
@@ -0,0 +1,287 @@
...
+                {
+                    await _writingStream.WriteAsync(_currentSegment, 0, _position, joinedToken);
+                }
+
+                await _writingStream.FlushAsync(joinedToken);
+            }
+            catch (OperationCanceledException)
+            {
+                // Catch any cancellation and translate it into setting isCanceled = true
+                return new FlushResult(isCanceled: true, IsCompletedOrThrow());
```
pakrym: If `CancellationToken cancellationToken` was cancelled we need to throw the exception.

src/Microsoft.AspNetCore.Http/PipeWriterAdapter.cs
```diff
@@ -0,0 +1,287 @@
...
+                // Canceled. At this point, we create a fresh CTS that isn't canceled.
+                var result = new FlushResult(isCanceled: true, IsCompletedOrThrow());
+                InternalTokenSource = new CancellationTokenSource();
+                _currentFlushCanceled = false;
+                return result;
+            }
+
+            // As an optimization, we only create a new CTS if the previous one wasn't canceled.
+            // Otherwise, we would need to create a new CTS for every call to FlushAsync.
+            var joinedToken = CancellationTokenSource.CreateLinkedTokenSource(
```
pakrym: Do this only if `cancellationToken` is cancellable.

src/Microsoft.AspNetCore.Http/PipeWriterAdapter.cs
```diff
@@ -0,0 +1,287 @@
...
+                InternalTokenSource.Token).Token;
+
+            try
+            {
+                // Write all completed segments and whatever remains in the current segment
+                // and flush the result.
+                if (_completedSegments != null)
+                {
+                    var count = _completedSegments.Count;
+                    for (var i = 0; i < count; i++)
```
pakrym: foreach?

pakrym (changes requested):
I'd like to see more implementation specific tests that assert expectations about how data should be written to the stream, memory pool block management, etc.

src/Microsoft.AspNetCore.Http/PipeWriterAdapter.cs
davidfowl: That’s much less efficient. Let’s measure the before and after though

src/Microsoft.AspNetCore.Http/PipeWriterAdapter.cs
```diff
@@ -0,0 +1,287 @@
...
+                }
+
+                // Position might be less than the segment length if there wasn't enough space to satisfy the sizeHint when
+                // GetMemory was called. In that case we'll take the current segment and call it "completed", but need to
+                // ignore any empty space in it.
+                _completedSegments.Add(new CompletedBuffer(_currentSegment, _position));
+            }
+
+            // Get a new buffer using the minimum segment size, unless the size hint is larger than a single segment.
+            _currentSegment = ArrayPool<byte>.Shared.Rent(Math.Max(_minimumSegmentSize, sizeHint));
```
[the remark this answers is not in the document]
davidfowl: Hmm not sure we want that in the default path.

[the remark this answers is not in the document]
benaadams: `ArrayPool<byte>.Shared` devirtualizes now https://github.com/dotnet/coreclr/pull/20637 so if you were using `IMemoryPool` you'd want something like
```csharp
var size = Math.Max(_minimumSegmentSize, sizeHint);
_currentSegment = _pool?.AllocByte(size) ?? ArrayPool<byte>.Shared.Rent(size);
```

Though, isn't `IMemoryPool` in the wrong namespace? (i.e. downstream)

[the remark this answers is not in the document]
davidfowl: IMemoryPool allocates like a mofo, we can avoid it in this case with little downside.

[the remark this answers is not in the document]
pakrym: Why don't we do the same in pipe then?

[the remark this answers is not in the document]
pakrym: Or think about the way to fix that allocation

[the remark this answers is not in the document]
davidfowl: There are some concrete benefits you get from the IMemoryPool
1. Somebody can plug in an implementation (it's an abstraction)
2. We have an implementation (in ASP.NET Core) that uses pre-pinned memory (a variant of the first one)

The default implementation (ArrayMemoryPool) allocates a managed object per segment. It's not that big a deal when doing large allocations but can be avoided completely when using the array pool. We get no benefit in ASP.NET Core if these Streams use pre-pinned buffers since the Streams we'll be writing to won't be pinning buffer for async reads. That happens in the transport layer in Kestrel.

Before reverting to use the IMemoryPool, we should get the baseline so we can measure the differences.

src/Microsoft.AspNetCore.Http/PipeWriterAdapter.cs
```diff
@@ -0,0 +1,287 @@
...
+using System.Runtime.ExceptionServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Http
+{
+    /// <summary>
+    /// Implements PipeWriter using a base stream implementation. 
+    /// </summary>
+    public class PipeWriterAdapter : PipeWriter, IDisposable
```
davidfowl: StreamPipeWriter?

davidfowl: override WriteAsync and provide an optimized path that writes directly to the Stream if there's no buffered data.

src/Microsoft.AspNetCore.Http/PipeWriterAdapter.cs
davidfowl: +1 This logic needs to be a little smarter. We should have a single CTS per adapter in the common case. Usually, CancelPendingFlush is called once per PipeWriter.

src/Microsoft.AspNetCore.Http/PipeWriterAdapter.cs
jkotalik: I'll measure the perf with and without using Memory here, then compare it with using IMemoryPool vs ArrayPool.

src/Microsoft.AspNetCore.Http/PipeWriterAdapter.cs
jkotalik: A long while ago, I remember for was more preformant than foreach.

src/Microsoft.AspNetCore.Http/PipeWriterAdapter.cs
jkotalik: When you say "No buffered data", you mean no data in _completedSegments or CurrentSegment?

src/Microsoft.AspNetCore.Http/StreamPipeWriter.cs
```diff
@@ -0,0 +1,299 @@
...
+                    // We need to treat the first completed segment uniquely because we allow 
+                    // partial consumption of the first CompletedSegment.
+                    // Example:
+                    // A user calls GetMemory(11) to move the _currentSegment _position to 11 and then call FlushAsync().
+                    // Afterwards, they call GetMemory(1000), which would return Memory in _currentSegment.
+                    // Afterwards, they call GetMemory(8000), which would return a new segment, which moves _currentSegment to _completedSegments
+                    // Then the user calls FlushAsync().
+                    // At this point, the firstSegment needs to be treated differently because it was only partially written to.
+                    // To fix this, we use _previousSegmentPosition to store how far we were into the first _completedSegment.
+                    var firstSegment = _completedSegments[0];
```
jkotalik: @davidfowl @pakrym this logic is pretty complicated due to supporting partial writes in a CompletedSegment (rather than resetting every time). I feel a linked list may suit us better here.

jkotalik:
Better perf numbers.

``` ini

BenchmarkDotNet=v0.10.13, OS=Windows 10.0.17134
Intel Xeon CPU E5-1650 v4 3.60GHz, 1 CPU, 12 logical cores and 6 physical cores
.NET Core SDK=2.2.100-preview2-009404
  [Host]     : .NET Core 2.2.0-preview2-26905-02 (CoreCLR 4.6.26905.03, CoreFX 4.6.26905.02), 64bit RyuJIT
  Job-MVJLBU : .NET Core 2.2.0-preview2-26905-02 (CoreCLR 4.6.26905.03, CoreFX 4.6.26905.02), 64bit RyuJIT

Runtime=Core  Server=True  Toolchain=.NET Core 2.2  
RunStrategy=Throughput  

```
|                    Method |        Mean |      Error |     StdDev |        Op/s | Allocated |
|-------------------------- |------------:|-----------:|-----------:|------------:|----------:|
|           WriteHelloWorld |    149.7 ns |   2.952 ns |   4.327 ns | 6,679,947.5 |      32 B |
| WriteHelloWorldLargeWrite | 60,649.7 ns | 812.198 ns | 759.731 ns |    16,488.1 |  100002 B |

benchmarks/Microsoft.AspNetCore.Http.Performance/StreamPipeWriterBenchmark.cs
```diff
@@ -0,0 +1,39 @@
...
+        private StreamPipeWriter _pipeWriter;
+        private byte[] _helloWorldBytes;
+        private byte[] _largeWrite;
+
+        [IterationSetup]
+        public void Setup()
+        {
+            _memoryStream = new MemoryStream();
+            _pipeWriter = new StreamPipeWriter(_memoryStream);
+            _helloWorldBytes = Encoding.ASCII.GetBytes("Hello World");
```
pakrym: static fields

jkotalik:
Perf numbers:

Using Memory and MemoryPool:

|                    Method |        Mean |        Error |     StdDev |        Op/s | Allocated |
|-------------------------- |------------:|-------------:|-----------:|------------:|----------:|
|           WriteHelloWorld |    200.1 ns |     3.846 ns |   4.579 ns | 4,998,578.2 |      32 B |
| WriteHelloWorldLargeWrite | 61,825.2 ns | 1,081.980 ns | 959.147 ns |    16,174.6 |  100003 B |


Using byte[] and ArrayPool:

|                    Method |        Mean |      Error |     StdDev |        Op/s | Allocated |
|-------------------------- |------------:|-----------:|-----------:|------------:|----------:|
|           WriteHelloWorld |    149.7 ns |   2.952 ns |   4.327 ns | 6,679,947.5 |      32 B |
| WriteHelloWorldLargeWrite | 60,649.7 ns | 812.198 ns | 759.731 ns |    16,488.1 |  100002 B |

src/Microsoft.AspNetCore.Http/StreamPipeWriter.cs
```diff
@@ -0,0 +1,307 @@
...
+                }
+
+                // Position might be less than the segment length if there wasn't enough space to satisfy the sizeHint when
+                // GetMemory was called. In that case we'll take the current segment and call it "completed", but need to
+                // ignore any empty space in it.
+                _completedSegments.Add(new CompletedBuffer(_currentSegmentOwner, _position));
+            }
+
+            // Get a new buffer using the minimum segment size, unless the size hint is larger than a single segment.
+            _currentSegmentOwner = _pool?.Rent(Math.Max(_minimumSegmentSize, sizeHint));
```
benaadams: Is `_pool` ever null? `_pool?.` if it can be next line will fail.

As aside it will also devirtualize in .NET Core 3.0 https://github.com/dotnet/corefx/pull/33085 so again you could not set it to `MemoryPool<byte>.Shared` in the constructor when `null` and instead here have:
```csharp
var size = Math.Max(_minimumSegmentSize, sizeHint);
_currentSegmentOwner = _pool?.Rent(size) ?? MemoryPool<byte>.Shared.Rent(size);
```

jkotalik: I should update this. My intention was to compare MemoryPool<byte>.Shared vs ArrayPool with this change. _pool right now cannot be null. Let me remove the nullable check.

src/Microsoft.AspNetCore.Http/Microsoft.AspNetCore.Http.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>ASP.NET Core default HTTP feature implementations.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.2</TargetFrameworks>
```
rynowak: 2.2 is so retro tho

benaadams: 3.0 is the best yet; though hard to reference, at this time 😢

src/Microsoft.AspNetCore.Http/StreamPipeWriter.cs
```diff
@@ -0,0 +1,313 @@
...
+
+            return _currentSegment;
+        }
+
+        /// <inheritdoc />
+        public override Span<byte> GetSpan(int sizeHint = 0)
+        {
+            EnsureCapacity(sizeHint);
+
+            return _currentSegment.Slice(_position).Span;
```
davidfowl: I believe getting a span then slicing is faster then slicing memory

pakrym: Yep

src/Microsoft.AspNetCore.Http/StreamPipeWriter.cs
```diff
@@ -0,0 +1,313 @@
...
+                if (_completedSegments != null && _completedSegments.Count > 0)
+                {
+                    var count = _completedSegments.Count;
+                    for (var i = 0; i < count; i++)
+                    {
+                        var segment = _completedSegments[i];
+#if NETCOREAPP2_2
+                        await _writingStream.WriteAsync(segment.Buffer.Slice(0, segment.Length), InternalTokenSource.Token);
+#elif NETSTANDARD2_0
+                        await _writingStream.WriteAsync(segment.Buffer.ToArray(), 0, segment.Length, InternalTokenSource.Token);
```
pakrym: `MemoryMarshal.TryGetArray`?

src/Microsoft.AspNetCore.Http/StreamPipeWriter.cs
```diff
@@ -0,0 +1,313 @@
...
+                await _writingStream.FlushAsync(_internalTokenSource.Token);
+            }
+            catch (OperationCanceledException)
+            {
+                // Remove the cancellation token such that the next time Flush is called
+                // A new CTS is created.
+                Interlocked.Exchange(ref _internalTokenSource, null);
+
+                // Catch any cancellation and translate it into setting isCanceled = true
+                return new FlushResult(isCanceled: true, IsCompletedOrThrow());
```
pakrym: We need to throw if `cancellationToken` that was passed in FlushAsync() was cancelled.

test/Microsoft.AspNetCore.Http.Tests/TestAdaptedPipe.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            Reader = reader;
+            Writer = writer;
+        }
+
+        public PipeReader Reader { get; private set; }
+
+        public PipeWriter Writer { get; private set; }
+
+        public void Reset()
```
pakrym: Why do you even have it?

test/Microsoft.AspNetCore.Http.Tests/StreamPipeWriterTests.cs
```diff
@@ -0,0 +1,498 @@
...
+        }
+
+        [Fact]
+        public void FlushAsyncReturnsCanceledIfCanceledBeforeFlush()
+        {
+            CheckCanceledFlush();
+        }
+
+        [Fact]
+        public void FlushAsyncReturnsCanceledIfCanceledBeforeFlushMultipleTimes()
```
pakrym: Don't mix tests copied from Pipes and new ones. It would be very hard to merge them back into corefx.

src/Microsoft.AspNetCore.Http/StreamPipeWriter.cs
```diff
@@ -0,0 +1,313 @@
...
+                    {
+                        var segment = _completedSegments[i];
+#if NETCOREAPP2_2
+                        await _writingStream.WriteAsync(segment.Buffer.Slice(0, segment.Length), InternalTokenSource.Token);
+#elif NETSTANDARD2_0
+                        await _writingStream.WriteAsync(segment.Buffer.ToArray(), 0, segment.Length, InternalTokenSource.Token);
+#else
+#error Target frameworks need to be updated.
+#endif
+                        segment.Return();
```
pakrym: You need to test more cancellation scenarios. Current logic would do crazy things if you continue writing after CancelPendingFlush (you always reset _bytesWritten but clear segments only after writing them all, this might cause double writes/double returns/etc).

jkotalik: Still adding a few more tests.

davidfowl:
https://github.com/aspnet/HttpAbstractions/pull/1065#issuecomment-438898235

These numbers don't sit well with me. I don't see how the allocated bytes can be that similar.

src/Microsoft.AspNetCore.Http/StreamPipeWriter.cs
```diff
@@ -0,0 +1,314 @@
...
+        /// <inheritdoc />
+        public override void OnReaderCompleted(Action<Exception, object> callback, object state)
+        {
+            if (callback == null)
+            {
+                throw new ArgumentNullException(nameof(callback));
+            }
+
+            // Note this currently doesn't do anything.
+            // Implementing completions/callbacks would require creating a PipeReaderAdapter.
```
davidfowl: Throw a NotSupportedException

src/Microsoft.AspNetCore.Http/StreamPipeWriter.cs
```diff
@@ -0,0 +1,314 @@
...
+
+                    await _writingStream.FlushAsync(_internalTokenSource.Token);
+
+                    return new FlushResult(isCanceled: false, IsCompletedOrThrow());
+                }
+                catch (OperationCanceledException)
+                {
+                    // Remove the cancellation token such that the next time Flush is called
+                    // A new CTS is created.
+                    lock (lockObject)
```
pakrym: Reading is not using lock and writing does, suspicious.

jkotalik: It's used when accessing the property: https://github.com/aspnet/HttpAbstractions/pull/1065/files#diff-dcfa2ec5d0439ce1abf9d81d7f9fc302R41

pakrym:
What are the latest perf results?

src/Microsoft.AspNetCore.Http/StreamPipeWriter.cs
```diff
@@ -0,0 +1,308 @@
...
+        private List<CompletedBuffer> _completedSegments;
+        private Memory<byte> _currentSegment;
+        private IMemoryOwner<byte> _currentSegmentOwner;
+        private MemoryPool<byte> _pool;
+        private int _position;
+
+        private CancellationTokenSource _internalTokenSource;
+        private bool _isCompleted;
+        private ExceptionDispatchInfo _exceptionInfo;
+        private object lockObject = new object();
```
davidfowl: nit: _lockObject

jkotalik:
Removed the MemoryStream and made an underlying noop stream instead.

|                    Method |       Mean |     Error |    StdDev |        Op/s | Allocated |
|-------------------------- |-----------:|----------:|----------:|------------:|----------:|
|           WriteHelloWorld |   179.2 ns |  3.603 ns |  6.404 ns | 5,580,116.0 |       0 B |
| WriteHelloWorldLargeWrite | 2,029.8 ns | 40.592 ns | 94.883 ns |   492,654.3 |       0 B |

src/Microsoft.AspNetCore.Http/StreamPipeWriter.cs
```diff
@@ -0,0 +1,320 @@
...
+                            await _writingStream.WriteAsync(segment.Buffer.Slice(0, segment.Length), localToken);
+#elif NETSTANDARD2_0
+                            MemoryMarshal.TryGetArray<byte>(segment.Buffer, out var arraySegment);
+                            await _writingStream.WriteAsync(arraySegment.Array, 0, segment.Length, localToken);
+#else
+#error Target frameworks need to be updated.
+#endif
+                            _bytesWritten -= segment.Length;
+                            segment.Return();
+                            _completedSegments.RemoveAt(0);
```
pakrym: For next PR: this is slow and we don't have a benchmark that exercises this code path at all.

pakrym (approved):
Seems fine for the first iteration.

src/Microsoft.AspNetCore.Http/StreamPipeWriter.cs
```diff
@@ -0,0 +1,320 @@
...
+            public CompletedBuffer(IMemoryOwner<byte> buffer, int length)
+            {
+                Buffer = buffer.Memory;
+                Length = length;
+                _memoryOwner = buffer;
+            }
+
+            public void Return()
+            {
+                _memoryOwner.Dispose();
```
pakrym: For the next time: null `_memoryOwner` and add assert to avoid double returns and catch bugs.

PR #1066, opened 2018-11-16
Use object indirection in HttpContextAccessor

benaadams:
Use an extra level of indirection in `HttpContextAccessor` so what is being held onto is shared (rather than its own reference); so when the request is done that second indirection can be cleared for everything holding on to `HttpContextHolder` at which point it will remain inert and they won't see data for the next request (as a new `HttpContextHolder` will be created)


src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs
```diff
@@ -7,21 +7,39 @@ namespace Microsoft.AspNetCore.Http
...
+                return  _httpContextCurrent.Value?.Context;
             }
             set
             {
-                _httpContextCurrent.Value = (value?.TraceIdentifier, value);
+                var holder = _httpContextCurrent.Value;
+                if (holder != null)
+                {
+                    // Kill current HttpContext trapped in the AsyncLocals, as its done.
+                    holder.Context = null;
```
davidfowl: Clever!

davidfowl:
Change the HttpContextFactory to not set the TraceIdentifier to null. Also tweak some of these tests https://github.com/aspnet/HttpAbstractions/pull/1021/files to avoid the TraceIdentifier comparison

davidfowl:
It's an extra allocation per request but async locals aren't super light anyways so this is fine.

src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs
```diff
@@ -7,21 +7,39 @@ namespace Microsoft.AspNetCore.Http
...
-                return value.traceIdentifier == value.context?.TraceIdentifier ? value.context : null;
+                return  _httpContextCurrent.Value?.Context;
             }
             set
             {
-                _httpContextCurrent.Value = (value?.TraceIdentifier, value);
+                var holder = _httpContextCurrent.Value;
+                if (holder != null)
+                {
+                    // Kill current HttpContext trapped in the AsyncLocals, as its done.
```
Tratcher: Rephrase, this will get flagged.

Tratcher (approved):

src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs
davidfowl: Kill -> clear

benaadams: Done

src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs
```diff
@@ -7,21 +7,39 @@ namespace Microsoft.AspNetCore.Http
...
+                    holder.Context = null;
+                }
+
+                if (value != null)
+                {
+                    // Use an object indirection to hold the HttpContext in the AsyncLocal,
+                    // so it can be cleared in all ExecutionContexts when its cleared.
+                    _httpContextCurrent.Value = new HttpContextHolder { Context = value };
+                }
+                else if (holder != null)
```
davidfowl: You check holder != null  twice in this method.

benaadams: Only want to set it to null if its already set; as setting to null is a value and will move off default context

benaadams: (Though that might only be if you are also listening to change notifications also)

davidfowl (approved):

src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs
davidfowl: > Only want to set it to null if its already set; as setting to null is a value and will move off default context

Can you add a comment to the code?

benaadams: Done

src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs
benaadams: If the context starts throwing https://github.com/aspnet/AspNetCore/issues/3586; then it wouldn't necessarily have to clear it here?

The change of holder (below) would still prevent it being shared across requests.

davidfowl: People use null to detect if they are in a request aware place.

